PR #13, opened 2021-03-14
Add configurations for building layers for java apps using either age…

anuraaga:
…nt or wrapper.

Will follow up with a sample app or two.

/cc @wangzlei @willarmiros


java/awssdk-autoconfigure/src/main/resources/software/amazon/awssdk/global/handlers/execution.interceptors
```diff
@@ -0,0 +1 @@
+io.opentelemetry.instrumentation.awssdk.v2_2.autoconfigure.AutoconfiguredTracingExecutionInterceptor
```
wangzlei, line 1: How does it work? looks like SPI

wangzlei (approved):

java/awssdk-autoconfigure/src/main/resources/software/amazon/awssdk/global/handlers/execution.interceptors
anuraaga: Yeah it's an AWS SDK specific format for SPI

java/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Layers for running Java applications on AWS Lambda with OpenTelemetry.
+
+## Provided layers
+
+Two types of layers are provided
+
+### Java agent
+
+The [AWS OpenTelemetry Java Agent](https://github.com/aws-observability/aws-otel-java-instrumentation)
```
trask: what do we need to do for this to work with the standard OpenTelemetry Java agent? would be better not to depend on vendor-specific distro here if possible

trask (approved):

java/README.md
anuraaga: Ah that's true - just need to add the aws resource extension, which I guess we said we're OK with doing. Either way I'll change this to the normal agent

anuraaga: Fixed

java/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The [OpenTelemetry Java Agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation)
+is bundled into the base of the layer and can be loaded into a Lambda function by specifying the
+environment variable `JAVA_TOOL_OPTIONS=-javaagent:/opt/opentelemetry-javaagent.jar` in your Lambda
+configuration. The agent will automatically instrument your application for all supported libraries.
+Note, automatic instrumentation has a notable impact on startup time on AWS Lambda and you will
+generally need to use this along with provisioned concurrency and warmup requests to serve production
+requests without causing timeouts on initial requests while it initializes.
+
+### Wrapper
```
ghost, line 19: It may make sense to denote that using wrapper will yield only a single span per lambda invocation (plus additional spans if appropriate lib instrumentation is loaded). Out of what I have seen, this (single span) is kind of confusing for users that are used to autoinstrumentation.

java/awssdk-autoconfigure/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/autoconfigure/AutoconfiguredTracingExecutionInterceptor.java
```diff
@@ -0,0 +1,151 @@
...
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+import software.amazon.awssdk.core.sync.RequestBody;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.http.SdkHttpResponse;
+
+/**
+ * {@link ExecutionInterceptor} that delegates to {@link AwsSdkTracing}, using the autoconfigured
+ * global SDK.
+ */
+public class AutoconfiguredTracingExecutionInterceptor implements ExecutionInterceptor {
```
ghost, line 40: This is veeery nice!

ghost (approved):

java/README.md
anuraaga: I tried to capture that with my "For any other library" comment below. Is it too confusing still? Let me know if there's a way to make it clearer.

ghost: Out of my previous experience with wrapper users I'd state explicitly "wrapper will create a single span for lambda execution" but it may be just me being extra sensitive about that ;)

anuraaga: Let's see how it goes, it's a bit different thanks to the AWS SDK instrumentation here

PR #14, opened 2021-03-16
Add sample applications and deployment scripts for AWS SDK and OkHttp.

anuraaga:

PR #15, opened 2021-03-17
collector extension and python sdk initial commit

wangzlei:
Collector extension and Python sdk layer.
Be aware that:
* Collector in extension is AWS distro OpenTelemetry Collector
* Collector extension contains a default config, export traces to AWS XRay.
* OTel Python dependencies version are not pinged, will update in later PR.
* Python folder contains AWS Lambda function instrumentor, should be moved to OTel Python contrib in near future.
* README for Python layer is not ready.
* CI is not ready.


collector/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
anuraaga, line 1: Remove LICENSE is already in top of repo

collector/config.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+exporters:
+  awsxray:
+  awsemf:
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      exporters: [awsxray]
```
anuraaga, line 14: We need to come to a decision on whether it's ok for a repository in `open-telemetry` to host vendor-specific configs, distros, examples. I'm personally not ok with it and would recommend this be in an aws distro repo. Yes it's easier for a user to start if using the aws distros, but I would suggest that in that case this repo doesn't really belong here which has been my general stance. 

@open-telemetry/technical-committee can you advice on this, which originated from

https://github.com/open-telemetry/community/issues/591#issuecomment-800764602
https://github.com/open-telemetry/opentelemetry-lambda/pull/14#discussion_r595603941

There was some concern about TC oversight if in a different repo by @dyladan 

https://github.com/open-telemetry/community/issues/591#issuecomment-756534589

But this doesn't seem to be too big of a deal since instrumentation can still live in open-telemetry, and has already started with https://github.com/open-telemetry/opentelemetry-specification/pull/1442 (assuming it ever gets merged :P). I assume this is the focus of the TC, the instrumentation experience. The actual end user experience is more of a vendor concern IMO - in this case it would be AWS balancing its hat as a cloud vendor and an observability vendor, but possibly then optimizing for an AWS getting-started experience.

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+logger = logging.getLogger(__name__)
+
+
+class AwsLambdaInstrumentor(BaseInstrumentor):
+    def _instrument(self, **kwargs):
+        self._tracer = get_tracer(__name__, __version__, kwargs.get("tracer_provider"))
+
+        self._tracer_provider = get_tracer_provider()
+
+        lambda_handler = os.environ.get("ORIG_HANDLER", os.environ.get("_HANDLER"))
```
anuraaga, line 72: Are you planning on moving this to the python instrumentation repo similar to the Java version? For @open-telemetry/technical-committee reference, my above comment only holds true if this code is moved to instrumentation (which it should for sure)

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+            import_module(self._wrapped_module_name),
+            self._wrapped_function_name,
+        )
+
+    def _functionPatch(self, original_func, instance, args, kwargs):
+        lambda_context = args[1]
+        ctx_aws_request_id = lambda_context.aws_request_id
+        ctx_invoked_function_arn = lambda_context.invoked_function_arn
+        orig_handler = os.environ.get("ORIG_HANDLER", os.environ.get("_HANDLER"))
+        xray_trace_id = os.environ.get("_X_AMZN_TRACE_ID", "")
```
anuraaga, line 94: This is still set even when X-Ray is off or in passive mode, so we need to check whether it's sampled or not like described in https://github.com/open-telemetry/opentelemetry-specification/pull/1442/files. If off, then we need to propagate from the HTTP headers for proxy requests.

python/src/otel/otel_sdk/opentelemetry/resource/__init__.py
```diff
@@ -0,0 +1,19 @@
+from os import environ
+from opentelemetry.sdk.resources import (
+    Resource,
+    ResourceDetector,
+)
+
+
+class AwsLambdaResourceDetector(ResourceDetector):
```
anuraaga, line 8: Guessing we'd add resource detectors like this and more here

https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/sdk-extension/opentelemetry-sdk-extension-aws

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -0,0 +1,82 @@
...
+console_exporter = os.environ.get("OTEL_DEBUG", None)
+if not console_exporter is None:
+    from opentelemetry.sdk.trace.export import ConsoleSpanExporter
+
+    trace.get_tracer_provider().add_span_processor(
+        SimpleExportSpanProcessor(ConsoleSpanExporter())
+    )
+    logger.info("Console exporter initialized.")
+
+ci = os.environ.get("_OTEL_CI", None)
```
anuraaga, line 41: CI shouldn't leak into the SDK, this should just be using the `OTEL_TRACES_EXPORTER` variable from the spec I think. Though it makes me wonder what this file is for? Doesn't python support the variables natively? We use that support for the java lambda wrappers so no special config wrapper

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java#L23

anuraaga (changes requested):
Blocking this PR until further discussion on the vendor aspect

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
anuraaga: Ah sorry I see in the PR description now that you call this out as TODO, phew

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
wangzlei: yeah, we should add a switcher here to turn off aws trace id propagation.

python/src/otel/otel_sdk/opentelemetry/resource/__init__.py
wangzlei: yes, will move to python contrib with aws instrumentor, if Lambda resource detector is accepted by Python maintainer :)

python/src/otel/otel_sdk/otel_wrapper.py
wangzlei: good.

codeboten (commented):
It would be good to split this PR, one for the extension layer and another for the Python SDK

wangzlei:
Split to 2 PRs for collector and python

Issue #18, opened 2021-03-22
How to configure B3 Propagator?


alex-vladut:
First of all, great initiative for putting this AWS Lambda Extension together 👍 Is there any option to configure the propagator to be B3 Propagator?
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md
https://github.com/openzipkin/b3-propagation

anuraaga:
Hey - are you using Java? Currently the agent version supports it but indeed I forgot to add the dependency to the wrapper version, will fix.

For other languages we're still working on scripts but do indent to support extension propagators to make sure interop works well - they would be used when active tracing (i.e., AWS X-Ray) is disabled for the function.

alex-vladut:
Thanks a lot for your prompt answer, sounds great. We're actually JavaScript for our AWS Lambda functions.

codeboten:
hey @alex-vladut, here's an example from the opentelemetry-js repo: https://github.com/open-telemetry/opentelemetry-js/blob/main/examples/tracer-web/examples/xml-http-request/index.js

With additional details here:

https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-propagator-b3

hope this helps!

anuraaga:
Thanks @alex-vladut - we'll be adding layer build scripts here for JavaScript soon, and will make sure to have the propagators available.

alex-vladut:
Nice one, thanks for your help guys 👍

serkan-ozal:
Hi @alex-vladut, glad to hear, it seems that you got what you are looking for.  @tylerbenson should we close this issue?

PR #19, opened 2021-03-23
Include extension propagators in Java wrapper

anuraaga:
Already included in java agent

See #18


anuraaga:
cc @wangzlei (not sure why you don't show up in reviewers list yet maybe GitHub still needs to propagate)

wangzlei (approved):

PR #21, opened 2021-04-02
create lambdacomponent module, collector can be replaced in downstream

wangzlei:
AWS wants to build Lambda layer in downstream, replace collector with ADOT collector by 
```
replace github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents => github.com/aws-observability/aws-otel-collector/pkg/lambdacomponents v0.6.0
```


collector/lambdacomponents/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents
+
+go 1.16
+
+replace github.com/open-telemetry/opentelemetry-collector-contrib/internal/awsxray => github.com/open-telemetry/opentelemetry-collector-contrib/internal/awsxray v0.22.0
+
+require (
+	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/awsemfexporter v0.22.0
```
anuraaga: Shouldn't these be added by the downstream (cloudwatch lambda layer)? If we think `collector-contrib` entirely is too huge, than just `collector` seems like a better upstream default.

wangzlei: My understanding is OTel Lambda upstream repo corresponds to Collector/SDK contrib, it can contain components in contrib.  Here it picks up 4 exporters from Collector core and 2 from contrib. Please refer code in https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/defaultcomponents/default.go#L29-L34

wangzlei: Exclude contrib stuff from OTel Lambda Repo is not a good idea, it would drive contributors to aws-observability/aws-otel-lambda since any 3rd party exporter, for example datadog exporter, can only be added in downstream repo.

anuraaga: Let's decide what 3rd party exporters we'll add here, can you file an issue to discuss it? For now we should keep it core as a placeholder while we decide.

anuraaga (approved):

PR #24, opened 2021-04-07
add CI/CD convention, remove aws sample

wangzlei:


utils/sam/run.sh
```diff
@@ -0,0 +1,127 @@
...
+main() {
+	echo "running..."
+	saved_args="$@"
+	template='template.yml'
+	build=false
+	deploy=false
+	layer=false
+
+	region=${AWS_REGION-$(aws configure get region)}
+	stack=${KEY-"otel"}
```
anuraaga: `KEY` seems like too generic of a name. Maybe `STACK`?

anuraaga: Or OTEL_LAMBDA_STACK (OTEL_LAMBDA_KEY would be ok too)

wangzlei: yeah, update to OTEL_LAMBDA_STACK

anuraaga (approved):

PR #28, opened 2021-04-08
Use collector extension in nodejs sample app.

anuraaga:


nodejs/packages/layer/package.json
```diff
@@ -25,6 +27,7 @@
   },
   "dependencies": {
     "@opentelemetry/api": "^1.0.0-rc.0",
+    "@opentelemetry/exporter-collector-proto": "^0.18.2",
```
anuraaga, line 30: We can switch to grpc after https://github.com/open-telemetry/opentelemetry-js/issues/2087

PR #32, opened 2021-04-15
Reorganize terraform deployment scripts.

anuraaga:
- Extract common module for API gateway boilerplate (upstream terraform module contains support for API Gateway v2 which doesn't support X-Ray, and not v1 which does...)

- Separate sample app module and integration test module
  - sample app module for reference by end user of public layer and for use in canary tests
  - integration test module for testing during development and for use in integration tests

- Remove provider.tf to allow setting region


anuraaga:
Going to go ahead and merge to unblock followup, happy to address anything as followup

PR #33, opened 2021-04-15
Avoid using println

rakyll:
println is a built-in not guranteed to stay in the language.
Switching to log.Println to be able to log timestamps and introducing
log to consistently log with a prefix everywhere.

### DIFF
--- a/collector/main.go
+++ b/collector/main.go
@@ -16,11 +16,11 @@ import (
 var (
 	extensionName   = filepath.Base(os.Args[0]) // extension name has to match the filename
 	extensionClient = extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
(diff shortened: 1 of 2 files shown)


anuraaga (approved):

collector/main.go
```diff
@@ -79,3 +80,9 @@ func prettyPrint(v interface{}) string {
 	}
 	return string(data)
 }
+
+// log is similar to fmt.Println but it logs a
+// log prefix before the log message.
+func log(a ...interface{}) {
+	fmt.Println(logPrefix, a)
```
anuraaga, line 87: Is this supposed to be `fmt`? PR description mentions adding timestamps

rakyll:
Sorry, I meant log.Println but actually, this won't help because it'll stringily a as a slice. Let's switch to zap. I'll send another change.

PR #44, opened 2021-04-16
Use OTLP exporter for java apps.

anuraaga:


anuraaga (commented):
Going to go ahead and merge since small change that fixes java publishing, happy to follow up with anything.

PR #46, opened 2021-04-16
Add terraform deployment for python.

anuraaga:
This only adds files so doesn't interfere with current sam approach. But I think having a terraform config for python may be helpful when finishing out tests.

### DIFF
--- a/python/integration-tests/function/main.tf
+++ b/python/integration-tests/function/main.tf
@@ -1,0 +1,22 @@
+resource "aws_lambda_layer_version" "opentelemetry_python_wrapper" {
+  layer_name          = var.sdk_layer_name
+  filename            = "../../src/build/layer.zip"
+  compatible_runtimes = ["python3.8"]
+  license_info        = "Apache-2.0"
+  source_code_hash    = filebase64sha256("../../src/build/layer.zip")
+}
+
+resource "aws_lambda_layer_version" "opentelemetry_collector" {
+  layer_name          = var.collector_layer_name
+  filename            = "../../../collector/build/collector-extension.zip"
+  compatible_runtimes = ["nodejs10.x", "nodejs12.x", "nodejs14.x"]
+  license_info        = "Apache-2.0"
+  source_code_hash    = filebase64sha256("../../../collector/build/collector-extension.zip")
+}
+
+module "function" {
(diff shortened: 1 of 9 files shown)


wangzlei (approved):

python/integration-tests/function/main.tf
```diff
@@ -0,0 +1,22 @@
...
+  filename            = "../../src/build/layer.zip"
+  compatible_runtimes = ["python3.8"]
+  license_info        = "Apache-2.0"
+  source_code_hash    = filebase64sha256("../../src/build/layer.zip")
+}
+
+resource "aws_lambda_layer_version" "opentelemetry_collector" {
+  layer_name          = var.collector_layer_name
+  filename            = "../../../collector/build/collector-extension.zip"
+  compatible_runtimes = ["nodejs10.x", "nodejs12.x", "nodejs14.x"]
```
wangzlei, line 12: that is ok since Lambda limits at most 5 compatible runtimes, but does not impact adding it to any other runtimes.

python/src/otel/Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG runtime=python3.8
+
+FROM public.ecr.aws/sam/build-${runtime}
+
+ADD . /workspace
+
+WORKDIR /workspace
+
+RUN mkdir -p /build && \
```
wangzlei, line 9: So we separate SDK layer and Collector layer?

python/integration-tests/function/main.tf
```diff
@@ -0,0 +1,22 @@
+resource "aws_lambda_layer_version" "opentelemetry_python_wrapper" {
```
wangzlei, line 1: Don't we consider merging SDK and Collector into one Lambda layer?

anuraaga: I might have missed something but I've always been expecting to keep them separate. They're completely separate concepts, so we would release them separately, like lego blocks. E.g., version of collector layer can match collector version, version of sdk layer can match sdk version. 

If we have a combined layer, do we need an uncombined layer for each language for customers that don't need the collector (e.g., export to Zipkin). That's a lot of layers.

wangzlei: make sense

PR #47, opened 2021-04-18
allow sam layer published name can be configurable

wangzlei:
1. Python layer directory contains a basic sample application, can be reused in CI/CD directly
2. Layer name is configurable


python/src/template.yml
```diff
@@ -1,14 +1,52 @@
...
+      Handler: lambda_function.lambda_handler
+      Runtime: python3.8
+      CodeUri: ./function
+      Description: Build OTel Python Lambda layer and sample app from scratch
+      MemorySize: 512
+      Timeout: 15
+      Policies:
+        - AWSLambdaBasicExecutionRole
+        - AWSLambda_ReadOnlyAccess
+        - AWSXrayWriteOnlyAccess
```
anuraaga:
```suggestion
```

python/src/template.yml
```diff
@@ -1,14 +1,52 @@
...
+        - AmazonS3FullAccess
```
anuraaga: Only give S3ListBuckets

python/src/template.yml
anuraaga: Or otherwise put in a folder called `xray` if it's a sample for xray

python/src/template.yml
```diff
@@ -1,14 +1,49 @@
...
         - python3.8
     Metadata:
-      BuildMethod: makefile
\ No newline at end of file
+      BuildMethod: makefile
+  api:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: api
+      TracingEnabled: true
```
anuraaga: Delete this too

anuraaga (approved):

PR #72, opened 2021-04-24
Reformat imports

rakyll:
Per Go style, standard library imports should be seperated from other imports.


rakyll:
This change is now in https://github.com/open-telemetry/opentelemetry-lambda/pull/73.

PR #79, opened 2021-04-26
Allow enabling API Gateway tracing

anuraaga:


anuraaga:
@codeboten Please ack this PR as it includes the term xray - it's only to create the option for downstream tests to enable xray tracing (which is in some sense not aws-vendor-specific as other vendors integrate with that too FWIU) but not push for it, want to make sure a non-AWS maintainer agrees it's ok.

PR #92, opened 2021-05-18
Added terraform configs and sample app for OTel .NET Lambda

lupengamzn:
- Added terraform configs and sample app for OTel .NET Lambda CICD tests


dotnet/sample-apps/aws-sdk/wrapper/SampleApps/SampleApps/Function.cs
```diff
@@ -0,0 +1,48 @@
...
+using Amazon.S3;
+using OpenTelemetry;
+using OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation;
+using OpenTelemetry.Trace;
+using System;
+
+// Assembly attribute to enable the Lambda function's JSON input to be converted into a .NET class.
+[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.Json.JsonSerializer))]
+
+namespace SampleApps
```
anuraaga: Can you call this `AwsSdkSample` or similar? Currently it's `SampleApps/SampleApps`, but I think each app is supposed to have a name

anuraaga (commented):
While at a glance the formatting looks ok, can you run `terraform fmt --recursive`?

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/SampleApps/Function.cs
lupengamzn: Just updated, and it turns out there needs to be a space between each item in terraform config

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+mkdir -p build/dotnet
+git clone https://github.com/open-telemetry/opentelemetry-dotnet-contrib.git
```
anuraaga: This script looks like it's not deterministic since this doesn't reference a git hash and not rerunnable since I don't think git clone works if the folder is already there. How about we add this as a git submodule?

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/AwsSdkSample/aws-lambda-tools-defaults.json
```diff
@@ -0,0 +1,23 @@
...
+    "To learn more about the Lambda commands with the .NET Core CLI execute the following command at the command line in the project root directory.",
+    "dotnet lambda help",
+    "All the command line options for the Lambda command can be specified in this file."
+  ],
+  "profile": "default",
+  "region": "us-east-1",
+  "configuration": "Release",
+  "framework": "netcoreapp3.1",
+  "function-runtime": "dotnetcore3.1",
+  "function-memory-size": 256,
```
wangzlei: Most of Lambda user use 512

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/AwsSdkSample/aws-lambda-tools-defaults.json
```diff
@@ -0,0 +1,23 @@
+{
+  "Information": [
+    "This file provides default values for the deployment wizard inside Visual Studio and the AWS Lambda commands added to the .NET Core CLI.",
+    "To learn more about the Lambda commands with the .NET Core CLI execute the following command at the command line in the project root directory.",
+    "dotnet lambda help",
+    "All the command line options for the Lambda command can be specified in this file."
+  ],
+  "profile": "default",
+  "region": "us-east-1",
```
wangzlei: can we get region from user's config?

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/AwsSdkSample/Function.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        /// A simple function that takes a APIGatewayProxyRequest and returns a APIGatewayProxyResponse
+        /// </summary>
+        /// <param name="input"></param>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public APIGatewayProxyResponse FunctionHandler(APIGatewayProxyRequest request, ILambdaContext context)
+        {
+            var S3Client = new AmazonS3Client();
+            _ = S3Client.ListBucketsAsync().Result;
```
wangzlei, line 44: Is this an async call ? If .Result is a blocking operation it is OK, else change to sync call.

anuraaga: It's ok if Result blocks but yeah better to use idiomatic pattern for sync calls, or alternatively use `await` as idiomatic pattern of async call

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/build.sh
lupengamzn: Updated to git submodule. Also in the `aws-otel-lambda` repo, we'll have to update the [option](https://github.com/aws-observability/aws-otel-lambda/blob/main/.github/workflows/main-build.yml#L48) from `true` to `recursive` to recursively checkout submodules from each folder.

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/AwsSdkSample/aws-lambda-tools-defaults.json
lupengamzn: Removed this file as we're using terraform to deploy the Lambda, not aws lambda cli

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/AwsSdkSample/Function.cs
lupengamzn: @wangzlei @anuraaga AWS .NET SDK for `AmazonS3Client` only has `async` APIs for public use. See: https://github.com/aws/aws-sdk-net/blob/475822dec5e87954b7a47ac65995714ae1f1b115/sdk/src/Services/S3/Generated/_netstandard/AmazonS3Client.cs. We can add `.Result` for the `sync` use case.

anuraaga (approved):

wangzlei (approved):

PR #105, opened 2021-06-24
Add several processors since they are lightweight

### LINKED ISSUE #102: Missing the Attributors Processor
Tried to use attribute processor and it doesn't seem to be compatible with Lambda/Node JS layer.

I tested the two configuration files below, but they didn't work. Is it possible to add to Lambda layer?

**References:**
https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/README.md#general-information
https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/attributesprocessor

**Codes and Errors:**
Error: Error: cannot load configuration: unknown processors type "attributes" for attributes

```receivers:
  otlp:
    protocols:
      grpc:
      http:

exporters:
  logging:
    loglevel: debug
  awsxray:
    index_all_attributes: true
    
processors:
  attributes:
    exclude:
      match_type: strict
      attributes:
        - key: otel_resource_process_executable_name

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: 
      - attributes
      exporters: [logging, awsxray]
    metrics:
      receivers: [otlp]
      exporters: [logging]
```

Error: Error: cannot load configuration: unknown processors type "attributes" for attributes/delete

```receivers:
  otlp:
    protocols:
      grpc:
      http:

exporters:
  logging:
    loglevel: debug
  awsxray:
    index_all_attributes: true
    
processors:
  attributes/delete:
    actions:
      - key: otel_resource_process_executable_name
        action: delete

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [attributes/delete]
      exporters: [logging, awsxray]
    metrics:
      receivers: [otlp]
      exporters: [logging]```

### PR DESCRIPTION
**Pull request by NathanielRN**
# Description

Fixes #102 

Initially, we hesitated with adding the `processors` because every Lambda function is subject to a [250 MB unzipped limit](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html) for the function and all the layers it includes.

However, when we added many of the processors (all except `batchprocessor` and `processorhelper`) available in https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor, we found that the size of the binary in the build only increased by ~700 KB:

```
 $ git/opentelemetry-lambda/collector du -k no-processors-build/extensions/collector
20064   no-processors-build/extensions/collector
 $ git/opentelemetry-lambda/collector du -k yes-processors-build/extensions/collector
20744   yes-processors-build/extensions/collector
```

This PR adds the processors so that they are available for lambda layers created using this repository.

### DIFF
--- a/collector/config.yaml
+++ b/collector/config.yaml
@@ -4,6 +4,14 @@ receivers:
       grpc:
       http:
 
+processors:
+  attributesprocessor:
+  filterprocessor:
+  memorylimiter:
+  probabilisticsamplerprocessor:
+  resourceprocessor:
+  spanprocessor:
+
 exporters:
   logging:
     loglevel: debug
@@ -13,6 +21,7 @@ service:
     traces:
       receivers: [otlp]
       exporters: [logging]
+      processors: [attributesprocessor, filterprocessor, memorylimiter, probabilisticsamplerprocessor, resourceprocessor, spanprocessor]
     metrics:
       receivers: [otlp]
       exporters: [logging]
--- a/collector/config.yaml
+++ b/collector/config.yaml
@@ -4,14 +4,6 @@ receivers:
       grpc:
       http:
 
-processors:
-  attributesprocessor:
-  filterprocessor:
-  memorylimiter:
-  probabilisticsamplerprocessor:
-  resourceprocessor:
-  spanprocessor:
-
 exporters:
   logging:
     loglevel: debug
@@ -21,7 +13,6 @@ service:
     traces:
       receivers: [otlp]
       exporters: [logging]
-      processors: [attributesprocessor, filterprocessor, memorylimiter, probabilisticsamplerprocessor, resourceprocessor, spanprocessor]
     metrics:
       receivers: [otlp]
       exporters: [logging]
--- a/collector/lambdacomponents/default.go
+++ b/collector/lambdacomponents/default.go
@@ -20,6 +20,12 @@ import (
 	"go.opentelemetry.io/collector/exporter/loggingexporter"
 	"go.opentelemetry.io/collector/exporter/otlpexporter"
 	"go.opentelemetry.io/collector/exporter/otlphttpexporter"
+	"go.opentelemetry.io/collector/processor/attributesprocessor"
+	"go.opentelemetry.io/collector/processor/filterprocessor"
+	"go.opentelemetry.io/collector/processor/memorylimiter"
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: collector/config.yaml [hunk 1]
```diff
@@ -4,6 +4,14 @@ receivers:
       grpc:
       http:
 
+processors:
```
wangzlei: We don't have to enable these processors by default.

collector/config.yaml
```diff
@@ -13,6 +21,7 @@ service:
     traces:
       receivers: [otlp]
       exporters: [logging]
+      processors: [attributesprocessor, filterprocessor, memorylimiter, probabilisticsamplerprocessor, resourceprocessor, spanprocessor]
```
wangzlei: same with previous one, user can enable specific processors by cumtom config

collector/config.yaml
NathanielRN: Thank you, that make sense! In this case we make them available because we put them in `default.go`, but it is up to users to add them to their specific `config.yaml` files before they publish the layer as available 😄 .

collector/lambdacomponents/default.go
```diff
@@ -20,6 +20,12 @@ import (
 	"go.opentelemetry.io/collector/exporter/loggingexporter"
 	"go.opentelemetry.io/collector/exporter/otlpexporter"
 	"go.opentelemetry.io/collector/exporter/otlphttpexporter"
+	"go.opentelemetry.io/collector/processor/attributesprocessor"
+	"go.opentelemetry.io/collector/processor/filterprocessor"
+	"go.opentelemetry.io/collector/processor/memorylimiter"
+	"go.opentelemetry.io/collector/processor/probabilisticsamplerprocessor"
+	"go.opentelemetry.io/collector/processor/resourceprocessor"
+	"go.opentelemetry.io/collector/processor/spanprocessor"
```
NathanielRN, line 28: From what I understand, we don't need to add `processorhelper` because the many of these processors depend on it. The `go build` command from `make build` will pull in anything necessary so we don't need to write it out explicitly.

anuraaga (approved):

wangzlei (approved):

awssandra:
Don't merge this yet - need to clear with Min

wangzlei:
> Don't merge this yet - need to clear with Min

no impact on aws-observability

PR #109, opened 2021-07-02
bump collector to 0.29.0

wangzlei:


collector/go.mod
```diff
@@ -6,5 +6,5 @@ replace github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponent
 
 require (
 	github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents v0.0.0
-	go.opentelemetry.io/collector v0.27.0
```
mxiamxia, line 9: can we use directly use the `go.opentelemetry.io/collector` import from `lambdacomponents`?
github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents

NathanielRN: I was curious about this too so I asked @wangzlei and he helped me understand why it is needed.

It turns out that if you delete this line and run `go mod tidy` it will edit this file and add it back! For that reason we probably should leave it so that people don't accidentally add this line when committing 😃

PR #110, opened 2021-07-03
Update Python Sample Apps for Upstream 1.3

NathanielRN:
# Description

To make our sample apps work with the 1.3 upstream release of OTel Python, we need to make changes to our sample app for the minor breaking changes.

# Testing

I was able to get the correct trace both by executing from the Lambda console and by using `curl -sS` to the API gateway link where my lambda layer was being used.

# Upstream PRs that caused changes

* [Instrumentation packages must indicate which packages they instrument #475](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/475)


python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -30,7 +30,7 @@
     )
 
     # Enable instrumentation
-    AwsLambdaInstrumentor().instrument()
+    AwsLambdaInstrumentor().instrument(skip_dep_check=True)
```
wangzlei, line 33: Can we use flag `skip_dep_check` to remove [these code](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/otel_sdk/otel_wrapper.py#L21-L47) ?

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -85,35 +96,50 @@ def _uninstrument(self, **kwargs):
...
-            # Refer: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/faas.md#example
-            span.set_attribute("faas.execution", ctx_aws_request_id)
-            span.set_attribute("faas.id", ctx_invoked_function_arn)
-
-            # TODO: fix in Collector because they belong resource attrubutes
-            span.set_attribute("faas.name", lambda_name)
-            span.set_attribute("faas.version", function_version)
-
-            result = original_func(*args, **kwargs)
+            if span.is_recording():
```
wangzlei, line 114: I am not sure it's necessary. When start the span and sampled, is_recording() is always true?

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
NathanielRN: Hm I'm not sure what we would replace it with. The `_load_distros`, `_load_instrumentors` and `_load_configurators` all initialize the classes they find. Normally, its important to **not** `skip_dep_check` because the classes `_load_instrumentors` loads **requires** that the dependencies are installed. Attempting to instrument without the dependencies would cause an import error.

I think it might be possible to use the upstream `opentelemetry-instrumentation` package as it is today instead of doing this code ourselves, but we don't need that in this PR.

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
NathanielRN: That's a good question! Initially I didn't know why we did this and I just did it because we do it for every other instrumentation as seen in https://github.com/open-telemetry/opentelemetry-python/issues/1057, so I wanted `AwsLambdaInstrumentor` to be the same.

But now I understand that we need this because we don't know if the Span will be Sampled. It is started like you said but if the Sampler [doesn't sample this span](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L999) then we shouldn't record attributes.

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
wangzlei: Let's merge the code first. I feel if is_recording() is essential why it is not integrated inside of start_span. I can raise issue in python repo later.

wangzlei (approved):

PR #120, opened 2021-07-21
Add otel lambda public layer to Node.js example app

kbkk:


nodejs/sample-apps/aws-sdk/deploy/variables.tf
```diff
@@ -4,18 +4,6 @@ variable "name" {
...
-  type        = string
-  description = "ARN for the Lambda layer containing the OpenTelemetry collector extension"
-  // TODO(anuraaga): Add default when a public layer is published.
-}
-
-variable "sdk_layer_arn" {
-  type        = string
-  description = "ARN for the Lambda layer containing the OpenTelemetry NodeJS Wrapper"
-  // TODO(anuraaga): Add default when a public layer is published.
-}
```
kbkk, line 17: Layers are region-based so I figured it'd be ok to base it on the "current region" setting. Since var defaults cannot be computed off data sources, therefore I moved this to a local var

nodejs/sample-apps/aws-sdk/deploy/main.tf
```diff
@@ -1,3 +1,7 @@
+locals {
+  collector_layer_arn = "arn:aws:lambda:${data.aws_region.current.name}:901920570463:layer:aws-otel-nodejs-ver-0-23-0:1"
```
anuraaga, line 2: Hi @kbkk - this layer is the AWS vendor distro, so we don't add it to the examples in this repo. We have sample apps using it in aws-otel-lambda.

We currently haven't published a public layer for this repo so can't make this change unfortunately.

kbkk: Ah, I understand. Sorry for the confusion. Closing

PR #124, opened 2021-07-29
Support other propagators in Python Layer

### LINKED ISSUE #123: [Python] Support for `OTEL_PROPAGATORS` setting beyond AWS X-Ray
Moving the issue from the [original issue in aws-otel-lambda](https://github.com/aws-observability/aws-otel-lambda/issues/123).

I am using v1.3.0 of the Python layer and configured the lambda to use `tracecontext` propagation, but I can see the lambda reporting an additional extra span that seems to be related to X-Ray, which shouldn't be there.

Notable additional config:
```
    OTEL_PROPAGATORS: "tracecontext"
    OTEL_TRACES_SAMPLER: "Always_On"
```

Lambda tracing is turned off (PassThrough).

Code:
```python
def consumer(event, lambda_context):

    context = extract(event['headers'])

    # Create top level transaction representing the lambda work
    with tracer.start_as_current_span("consumer-function-top-level", context=context, kind=SpanKind.SERVER):

        # Some other internal work performed by the lambda
        with tracer.start_as_current_span("consumer-function-some-internal-work", kind=SpanKind.INTERNAL):

            time.sleep(1)

        time.sleep(0.5)

    return {'statusCode': 200}
```

This should produce 2 spans: top-level `consumer-function-top-level` and nested `consumer-function-some-internal-work`. Instead, I am seeing 3 spans created in CloudWatch log with Span 2 `065c07aa6cc4916d` that shouldn't be there:
```log


2021-07-24T04:36:51.316Z	INFO	loggingexporter/logging_exporter.go:41	TracesExporter	{     "#spans": 3 }
--
2021-07-24T04:36:51.316Z	DEBUG	loggingexporter/logging_exporter.go:51	ResourceSpans #0
Resource labels:
-> telemetry.sdk.language: STRING(python)
-> telemetry.sdk.name: STRING(opentelemetry)
-> telemetry.sdk.version: STRING(1.3.0)
-> cloud.region: STRING(ap-southeast-2)
-> cloud.provider: STRING(aws)
-> faas.name: STRING(aws-python-api-worker-project-dev-consumer)
-> faas.version: STRING($LATEST)
-> service.name: STRING(aws-python-api-worker-project-dev-consumer)
InstrumentationLibrarySpans #0
InstrumentationLibrary handler

Span #0
Trace ID       : ac4d7110c6d22e985b5db28e8ecca769
Parent ID      : 3c374013084a679c
ID             : 522890b38766e922
Name           : consumer-function-some-internal-work
Kind           : SPAN_KIND_INTERNAL
Start time     : 2021-07-24 04:36:49.813522476 +0000 UTC
End time       : 2021-07-24 04:36:50.814894967 +0000 UTC
Status code    : STATUS_CODE_UNSET
Status message :

Span #1
Trace ID       : ac4d7110c6d22e985b5db28e8ecca769
Parent ID      : e24504245b7cff0e
ID             : 3c374013084a679c
Name           : consumer-function-top-level
Kind           : SPAN_KIND_SERVER
Start time     : 2021-07-24 04:36:48.812135568 +0000 UTC
End time       : 2021-07-24 04:36:51.315562096 +0000 UTC
Status code    : STATUS_CODE_UNSET
Status message :

Span #2 <==== EXTRA SPAN!!!
Trace ID       : ac4d7110c6d22e985b5db28e8ecca769
Parent ID      : ad1bb604de1da312
ID             : 065c07aa6cc4916d
Name           : handler.consumer
Kind           : SPAN_KIND_SERVER
Start time     : 2021-07-24 04:36:48.811673087 +0000 UTC
End time       : 2021-07-24 04:36:51.315622948 +0000 UTC
Status code    : STATUS_CODE_UNSET
Status message :
Attributes:
-> faas.execution: STRING(e759abdb-6119-4f91-afe0-716a71276825)
-> faas.id: STRING(arn:aws:lambda:ap-southeast-2:401722391821:function:aws-python-api-worker-project-dev-consumer)
-> faas.name: STRING(aws-python-api-worker-project-dev-consumer)
-> faas.version: STRING($LATEST)


```

By the looks of it, the Python implementation doesn't support autodiscovery and trace context propagation assuming the propagation is done using *only* AWS X-Ray trace context propagation.

NathanielRN:
We want to support other propagators besides the X-Ray propagator in the opentelemetry python lambda layer.

These changes allow that, by making the `AwsLambdaInstrumentor` class
* Read the `_X_AMZN_TRACE_ID` environment variable FIRST and then, if that fails, from the Lambda event HTTP headers (either using the default `event_context_extractor` or the user provided one)
* The default `event_context_extractor` uses the global propagator to extract and parent spans with the context in the HTTP headers

Fixes #123

### DIFF
--- a/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
+++ b/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
@@ -130,7 +130,7 @@ def _default_event_context_extractor(lambda_event: Any) -> Context:
         lambda_event: user-defined, so it could be anything, but this
             method counts it being a map with a 'headers' key
     Returns:
-        A Context with configuration found in the carrier.
+        A Context with configuration found in the event.
     """
     try:
         headers = lambda_event["headers"]
--- a/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
+++ b/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
@@ -86,8 +86,8 @@ class AwsLambdaInstrumentor(BaseInstrumentor):
             **kwargs: Optional arguments
                 ``tracer_provider``: a TracerProvider, defaults to global
                 ``event_context_extractor``: a method which takes the Lambda
-                    Event as input and provides the object which contains the
-                    context as output (usually HTTP headers)
+                    Event as input and extracts an OTel Context from it. Usually
+                    the the context is extract from HTTP headers on the event.
         """
         tracer = get_tracer(
             __name__, __version__, kwargs.get("tracer_provider")
@@ -173,10 +173,6 @@ def _instrument(
             .trace_flags.sampled
         ):
             return parent_context
-
-        logger.debug(
-            "X-Ray propagation failed, using user-configured propagators to extract context from Lambda Event."
-        )
 
         if event_context_extractor:
             parent_context = event_context_extractor(lambda_event)
--- a/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
+++ b/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
@@ -86,8 +86,9 @@ class AwsLambdaInstrumentor(BaseInstrumentor):
             **kwargs: Optional arguments
                 ``tracer_provider``: a TracerProvider, defaults to global
                 ``event_context_extractor``: a method which takes the Lambda
-                    Event as input and extracts an OTel Context from it. Usually
-                    the the context is extract from HTTP headers on the event.
+                    Event as input and extracts an OTel Context from it. By default,
+                    the context is extracted from the HTTP headers of an API Gateway
+                    request.
         """
         tracer = get_tracer(
             __name__, __version__, kwargs.get("tracer_provider")
@@ -134,7 +135,9 @@ def _default_event_context_extractor(lambda_event: Any) -> Context:
     try:
         headers = lambda_event["headers"]
     except (TypeError, KeyError):
-        logger.warning("Failed to extract context from Lambda Event.")
+        logger.debug(
+            "Extracting context from Lambda Event failed: either enable X-Ray active tracing or configure API Gateway to trigger this Lambda function as a pure proxy. Otherwise, generated spans will have an invalid (empty) parent context."
+        )
         headers = {}
     return get_global_textmap().extract(headers)
 
--- a/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
+++ b/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
@@ -47,19 +47,27 @@ API
 import logging
 import os
 from importlib import import_module
-from typing import Collection
-from wrapt import wrap_function_wrapper
+from typing import Any, Collection
 
-# TODO: aws propagator
-from opentelemetry.sdk.extension.aws.trace.propagation.aws_xray_format import (
-    AwsXRayFormat,
-)
+from opentelemetry.context.context import Context
 from opentelemetry.instrumentation.aws_lambda.package import _instruments
 from opentelemetry.instrumentation.aws_lambda.version import __version__
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor
 from opentelemetry.instrumentation.utils import unwrap
+from opentelemetry.propagate import get_global_textmap
+from opentelemetry.sdk.extension.aws.trace.propagation.aws_xray_format import (
+    TRACE_HEADER_KEY,
+    AwsXRayFormat,
+)
 from opentelemetry.semconv.trace import SpanAttributes
-from opentelemetry.trace import SpanKind, get_tracer, get_tracer_provider
+from opentelemetry.trace import (
+    SpanKind,
+    Tracer,
+    get_tracer,
+    get_tracer_provider,
+)
+from opentelemetry.trace.propagation import get_current_span
+from wrapt import wrap_function_wrapper
 
 logger = logging.getLogger(__name__)
 
@@ -69,15 +77,22 @@ class AwsLambdaInstrumentor(BaseInstrumentor):
         return _instruments
 
     def _instrument(self, **kwargs):
-        """Instruments Lambda Handlers on AWS Lambda
+        """Instruments Lambda Handlers on AWS Lambda.
+
+        See more:
+        https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#instrumenting-aws-lambda
 
         Args:
             **kwargs: Optional arguments
                 ``tracer_provider``: a TracerProvider, defaults to global
+                ``event_context_extractor``: a method which takes the Lambda
+                    Event as input and provides the object which contains the
+                    context as output (usually HTTP headers)
         """
         tracer = get_tracer(
             __name__, __version__, kwargs.get("tracer_provider")
         )
+        event_context_extractor = kwargs.get("event_context_extractor")
 
         lambda_handler = os.environ.get(
             "ORIG_HANDLER", os.environ.get("_HANDLER")
@@ -87,7 +102,10 @@ class AwsLambdaInstrumentor(BaseInstrumentor):
         self._wrapped_function_name = wrapped_names[1]
 
         _instrument(
-            tracer, self._wrapped_module_name, self._wrapped_function_name
+            tracer,
+            self._wrapped_module_name,
+            self._wrapped_function_name,
+            event_context_extractor,
         )
 
     def _uninstrument(self, **kwargs):
@@ -97,16 +115,84 @@ class AwsLambdaInstrumentor(BaseInstrumentor):
         )
 
 
-def _instrument(tracer, wrapped_module_name, wrapped_function_name):
+def _default_event_context_extractor(lambda_event: Any) -> Context:
+    """Default way of extracting the context from the Lambda Event.
+
+    Assumes the Lambda Event is a map with the headers under the 'headers' key.
+    This is the mapping to use when the Lambda is invoked by an API Gateway
+    REST API where API Gateway is acting as a pure proxy for the request.
+
+    See more:
+    https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
+
+    Args:
+        lambda_event: user-defined, so it could be anything, but this
+            method counts it being a map with a 'headers' key
+    Returns:
+        A Context with configuration found in the carrier.
+    """
+    try:
+        headers = lambda_event["headers"]
+    except (TypeError, KeyError):
+        logger.warning("Failed to extract context from Lambda Event.")
+        headers = {}
+    return get_global_textmap().extract(headers)
+
+
+def _instrument(
+    tracer: Tracer,
+    wrapped_module_name,
+    wrapped_function_name,
+    event_context_extractor=None,
+):
+    def _determine_parent_context(lambda_event: Any) -> Context:
+        """Determine the parent context for the current Lambda invocation.
+
+        See more:
+        https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#determining-the-parent-of-a-span
+
+        Args:
+            lambda_event: user-defined, so it could be anything, but this
+                method counts it being a map with a 'headers' key
+        Returns:
+            A Context with configuration found in the carrier.
+        """
+        parent_context = None
+
+        xray_env_var = os.environ.get("_X_AMZN_TRACE_ID")
+
+        if xray_env_var:
+            parent_context = AwsXRayFormat().extract(
+                {TRACE_HEADER_KEY: xray_env_var}
+            )
+
+        if (
+            parent_context
+            and get_current_span(parent_context)
+            .get_span_context()
+            .trace_flags.sampled
+        ):
+            return parent_context
+
+        logger.debug(
+            "X-Ray propagation failed, using user-configured propagators to extract context from Lambda Event."
+        )
+
+        if event_context_extractor:
+            parent_context = event_context_extractor(lambda_event)
+        else:
+            parent_context = _default_event_context_extractor(lambda_event)
+
+        return parent_context
+
     def _instrumented_lambda_handler_call(call_wrapped, instance, args, kwargs):
         orig_handler_name = ".".join(
             [wrapped_module_name, wrapped_function_name]
         )
 
-        # TODO: enable propagate from AWS by env variable
-        xray_trace_id = os.environ.get("_X_AMZN_TRACE_ID", "")
-        propagator = AwsXRayFormat()
-        parent_context = propagator.extract({"X-Amzn-Trace-Id": xray_trace_id})
+        lambda_event = args[0]
+
+        parent_context = _determine_parent_context(lambda_event)
 
         with tracer.start_as_current_span(
             name=orig_handler_name, context=parent_context, kind=SpanKind.SERVER
--- a/python/src/otel/tests/test_otel.py
+++ b/python/src/otel/tests/test_otel.py
@@ -17,8 +17,10 @@ from importlib import import_module
 from unittest import mock
 
 from opentelemetry.instrumentation.aws_lambda import AwsLambdaInstrumentor
+from opentelemetry.propagate import get_global_textmap
 from opentelemetry.sdk.extension.aws.trace.propagation.aws_xray_format import (
     TRACE_ID_FIRST_PART_LENGTH,
+    TRACE_ID_VERSION,
 )
 from opentelemetry.semconv.resource import ResourceAttributes
 from opentelemetry.semconv.trace import SpanAttributes
@@ -26,6 +28,7 @@ from opentelemetry.test.test_base import TestBase
 from opentelemetry.trace import SpanKind
 
 _HANDLER = "_HANDLER"
+_X_AMZN_TRACE_ID = "_X_AMZN_TRACE_ID"
 AWS_LAMBDA_EXEC_WRAPPER = "AWS_LAMBDA_EXEC_WRAPPER"
 INSTRUMENTATION_SRC_DIR = os.path.join(
     *(os.path.dirname(__file__), "..", "otel_sdk")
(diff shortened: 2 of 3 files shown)


python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -99,36 +96,32 @@ def _uninstrument(self, **kwargs):
...
 def _instrument(tracer, wrapped_module_name, wrapped_function_name):
     def _instrumented_lambda_handler_call(call_wrapped, instance, args, kwargs):
-        orig_handler_name = ".".join(
-            [wrapped_module_name, wrapped_function_name]
+        lambda_event, lambda_context = args[0], args[1]
+
+        parent_context = get_global_textmap().extract(
+            lambda_event.get("headers", {}).update(
+                {"X-Amzn-Trace-Id": environ.get("_X_AMZN_TRACE_ID", "")}
+            )
```
NathanielRN: How will this work if you have _both_ `TraceContextPropagator` and `AwsPropagator`? Will the context be a fix of both? With 2 spans?

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -40,5 +43,68 @@ environ["OTEL_RESOURCE_ATTRIBUTES"] = "%s,%s" % (
...
+_load_instrumentors()
+
+# TODO: Move to opentelemetry-python-contrib so that adding
+# `opentelemetry-instrumentation-lambda` to the requirements.txt file will let
+# this automatically be picked up by `_load_instrumentors()` above
+AwsLambdaInstrumentor().instrument(skip_dep_check=True)
+
+# Start the runtime with the originally intended arguments
+# TODO: Ask, is this what we want? Can we just do `system(sys.argv[1:])`?
 system(" ".join(args))
```
NathanielRN: Was this intentional? I don't think any of the setup functions above modify the system arguments? And if they do, do we want to ignore that?

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -99,36 +96,32 @@ def _uninstrument(self, **kwargs):
 
 def _instrument(tracer, wrapped_module_name, wrapped_function_name):
     def _instrumented_lambda_handler_call(call_wrapped, instance, args, kwargs):
-        orig_handler_name = ".".join(
-            [wrapped_module_name, wrapped_function_name]
+        lambda_event, lambda_context = args[0], args[1]
+
+        parent_context = get_global_textmap().extract(
+            lambda_event.get("headers", {}).update(
```
garrettwegan: I don't think we know headers will always be located at lambda_event.headers. If a spec guaranteed all lambda entry points follow this convention we could do this, but I don't believe such a spec exists. See how eventContextExtractor in [JS Instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/68e54490b207980b080d6642be48e81d17659cc7/plugins/node/opentelemetry-instrumentation-aws-lambda#aws-lambda-instrumentation-options) handles this by requiring user to provide header extraction function. Note JS requires user to fully extract into context which is unnecessary for here since you are using global propagator to extract. Here could require user to just return headers from event - similar to WIP Golang instrumentation.

NathanielRN: Hm I see what you're saying. Yeah [this guide I found online](https://aws.amazon.com/premiumsupport/knowledge-center/custom-headers-api-gateway-lambda/) gives an example on how to populate `"headers"` in API Gateway, but it's not a general solution.

I think we could do something like JS for the manual instrumentation case, but not the auto instrumentation case. In auto instrumentation the `AwsLambdaInstrumentor` will run without hooks and later on it will be too late to add the hooks.

So either we force users to manual instrument `AwsLambdaInstrumentor` in their code before their `lambda_handler`, or we define a mapping "ourselves" here as `events["headers"]` and explain to users how they can add them...

I think we can add the hooks in a subsequent PR, the question here is just do we want to give users the expectation that we will looks for HTTP headers in the `event["headers"]`? I think it could for us even though there is no spec, as long as we document it somewhere here?

garrettwegan: So `event["headers"]` may or may not be pointing into the actual customer event. If you invoke lambda directly (cli/sdk) `event` is your custom event type so yes we could point to one place the header must be, however, if lambda is invoked via some other AWS service (API Gateway, SQS, etc) `event` is some AWS specified type and your custom event is within that type somewhere.

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
anuraaga: You should always use the xray propagator format when reading from the environment variable, and use the configured propagator if reading from HTTP headers when the environment variable wasn't sampled

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
michaelhyatt: First of all, many thanks for looking into this.

My 2c:

Lambda handlers always accept 2 arguments: event and context. Event is the message delivered by a `Carrier` (API GW, SQS, etc) and should have both, data and metadata about the message that can include the context. [Propagators](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md) define the rules about how the context should be propagated with a particular `Carrier`. 

So, it should be safe to assume that `event` should contain the required metadata with the context. As the default implementation, it should be safe to assume that `event["headers"]` will contain the context data and it may be 95% of all cases where people use API-GW and SQS triggered lambdas. In other cases, the users can provide their own extractor and injector implementations, as JS lambda implementation does.

In the case where there is no data in the message (i.e. `event["header"]`) that carries the context for the defined propagator, it should default to the span with no parent and start a new trace, as opposed to defaulting to AWS X-Ray propagator. 

Also, specifying AWS X-Ray propagator should be a choice, i.e. the user should explicitly define it in `OTEL_PROPAGATORS=aws_xray` env variable.

Hope it makes sense.

python/src/otel/otel_sdk/otel-instrument
wangzlei: args is sys.argv[1:], should be same.

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -40,5 +43,68 @@ environ["OTEL_RESOURCE_ATTRIBUTES"] = "%s,%s" % (
     environ.get("OTEL_RESOURCE_ATTRIBUTES"),
 )
 
-# start the runtime with the extra options
+import logging
+
+from opentelemetry.environment_variables import OTEL_PYTHON_DISABLED_INSTRUMENTATIONS
```
wangzlei: Not sure if bootstrap can find dependencies, may need to set dependency path.
Another concern is bootstrap is earlier than Lambda Runtime, can you get the log of bootstrap? If get any error here can we debug the problem by log?

NathanielRN: Ah thanks for your prompt, that's a good question!

I think I was too ambitious with all these changes here, this PR is really just about propagators.

I've decided to move them to a separate PR in #125 so we can discuss all the concerns there separately.

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
NathanielRN: Hm I think I've seen in other Python code that they're able to do just `get_global_textmap().extract(headers)`. But we can't do that and have the behavior you mentioned if the user sets something like `OTEL_PROPAGATORS=xray,tracecontext` for their environment variables.

Maybe the answer is to read the `OTEL_PROPAGATORS` variable directly and instantiate the propagators? But that doesn't feel right... Would probably need to investigate the behavior of when we have multiple propagators.

python/src/otel/otel_sdk/otel-instrument
NathanielRN: Resolving this with conversation continued in #125

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
NathanielRN: @garrettwegan

I guess we need to find out the different branching paths for what an "event could be". For now, I followed the spec where it says [how to determine the parent span in Lambda](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#determining-the-parent-of-a-span) to guide a "**guess**" for what API Gateway would look like, but I haven't tested it so I wouldn't be comfortable merging this _just_ yet. But if we test it and it works (In the X-Ray case, the not X-Ray case, and the not _any_ propagator case), and people agree with it, we can make this incremental improvement.

@michaelhyatt in trying to follow this spec on [how to determine the parent span in Lambda](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#determining-the-parent-of-a-span) I made several updates to this PR. It still requires serious testing though:
* test with API Gateway configured in its 2 possible forms
* test when multiple combinations of propagators are set via `OTEL_PROPAGATORS`

I can't test it this week but if we get a chance to test it in the near future and people are okay with checking `event["headers"]` as a default we can probably make this incremental improvement to unblock you.

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
anuraaga: The env variable only ever fills in with AWS format so we can't really rely on configured propagator to read from it. In other languages we directly reference the awsxraypropagator through code without using the global can't you do that here?

NathanielRN: Yes that's what we intend to do here! We first try the `XRay` propagator, but if that fails, we try the global propagator. That's what I interpret the specs on [how to parent a Lambda trace](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#determining-the-parent-of-a-span) saying we should do?

After updating the PR it looks like this:

```python
parent_context = None

xray_env_var = os.environ.get("_X_AMZN_TRACE_ID")

if xray_env_var:
    parent_context = AwsXRayFormat().extract(
        {"X-Amzn-Trace-Id": xray_env_var}
    )

event_span_attributes = {}

if not parent_context or not parent_context.trace_flags.sampled:
    lambda_event = args[0]
    # ... Try the global propagator
```

michaelhyatt:
It makes sense @NathanielRN 

The only thing I wanted to add is highlight the behaviour when `tracing=PassThrough`, which should switch off X-Ray propagation, as per the [document](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#determining-the-parent-of-a-span):
> The environment variable will be set and the Context will be sampled only if AWS X-Ray has been enabled for the Lambda function. A user can disable AWS X-Ray for the function if this propagation is not desired.

NathanielRN:
@michaelhyatt

> tracing=PassThrough, which should switch off X-Ray propagation, as per the document:

I didn't see the document mention `tracing=PassThrough`, did you see it somewhere else? In regards to the part you quoted, I think the document is saying that "switching off X-Ray" will mean that the environment variable is simply not set. Without the environment variable, we won't be able to use the X-Ray propagator anyways, so we can fallback to the user global configured propagator as they would expect 🙂 (Therefore, we don't need any additional configuration flags like `tracing=PassThrough` because it will just try possible actions after the other until a context is extracted).

michaelhyatt:
I was under the impression that `_X_AMZN_TRACE_ID` gets populated automatically from the HTTP header `X-Amzn-Trace-ID` rather than being exposed to the end user to control whether the tracing is switched on or off, but I may be wrong. I just want to make sure that I can use `OTEL_PTOPAGATORS=tracecontext` and not get any `parent.id` references when I am sending a request without the externally set `traceparent` header.

I interpreted the requirement for X-Ray tracing to be switched off as turning the `Tracing` control in the lambda UI off, i.e. setting `tracing=PassThrough` (as opposed to `Active`), but I may be wrong here too. BTW, I believe these two values are from CloudFormation.

NathanielRN:
> I was under the impression that _X_AMZN_TRACE_ID gets populated automatically from the HTTP header X-Amzn-Trace-ID rather than being exposed to the end user to control whether the tracing is switched on or off, but I may be wrong.

Thanks for that prompt! I did some investigation and found out that you're right about this, so I made a PR in the specifications to follow up on this: https://github.com/open-telemetry/opentelemetry-specification/pull/1867

> I interpreted the requirement for X-Ray tracing to be switched off as turning the Tracing control in the lambda UI off, i.e. setting tracing=PassThrough (as opposed to Active), but I may be wrong here too.

Yes I agree with you that sounds like the ideal customer experience!

michaelhyatt:
@NathanielRN I am curious if the fix made it into `v1.5.0`? Maybe not, I can see this one still being a draft.

NathanielRN:
@michaelhyatt Sorry about that, yes this change hasn't been merged in. Haven't had bandwidth to test this solution even though I think it should work, once it gets tested by publishing a built version of this layer we can merge this change.

NathanielRN:
Hey all, I've updated this PR to have the minimal changes we would need to support this feature 🙂 With these changes, we can parent the context of subsequent spans using the HTTP headers in the lambda event. Further, using `event_context_extractor` you can define a custom way of extracting the context.

Tests have also been added to validate this.

Please let me know what you think!

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -69,15 +77,22 @@ def instrumentation_dependencies(self) -> Collection[str]:
...
+
+        See more:
+        https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#instrumenting-aws-lambda
 
         Args:
             **kwargs: Optional arguments
                 ``tracer_provider``: a TracerProvider, defaults to global
+                ``event_context_extractor``: a method which takes the Lambda
+                    Event as input and provides the object which contains the
+                    context as output (usually HTTP headers)
```
anuraaga: IIUC, it returns the Context itself, not HTTP headers, right?

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -97,16 +115,84 @@ def _uninstrument(self, **kwargs):
...
+
+        if (
+            parent_context
+            and get_current_span(parent_context)
+            .get_span_context()
+            .trace_flags.sampled
+        ):
+            return parent_context
+
+        logger.debug(
```
anuraaga: X-Ray being disabled isn't a failure so we shouldn't log it as such. I guess we don't need this log message anyways

anuraaga (commented):
Thanks, just some small points

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
NathanielRN: I thought it could be useful as a "debug" log and so labeled it as such, but I agree it's not necessary so I'll remove it! 🙂

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
NathanielRN: Thanks for that catch! This was outdated, I've updated it now 🙂

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -69,15 +77,22 @@ def instrumentation_dependencies(self) -> Collection[str]:
...
+
+        See more:
+        https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#instrumenting-aws-lambda
 
         Args:
             **kwargs: Optional arguments
                 ``tracer_provider``: a TracerProvider, defaults to global
+                ``event_context_extractor``: a method which takes the Lambda
+                    Event as input and extracts an OTel Context from it. Usually
+                    the the context is extract from HTTP headers on the event.
```
anuraaga:
```suggestion
                    Event as input and extracts an OTel Context from it. By default,
                    the context is extracted from the HTTP headers of an API Gateway
                    request.
```

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -97,16 +115,80 @@ def _uninstrument(self, **kwargs):
...
+    REST API where API Gateway is acting as a pure proxy for the request.
+
+    See more:
+    https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
+
+    Args:
+        lambda_event: user-defined, so it could be anything, but this
+            method counts it being a map with a 'headers' key
+    Returns:
+        A Context with configuration found in the carrier.
```
anuraaga:
```suggestion
        A Context with configuration found in the event.
```

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -97,16 +115,80 @@ def _uninstrument(self, **kwargs):
...
+    """
+    try:
+        headers = lambda_event["headers"]
+    except (TypeError, KeyError):
+        logger.warning("Failed to extract context from Lambda Event.")
```
anuraaga: We can add more on what a user should do if they see this error. Either X-Ray needs to be enabled or the function must be serving API Gateway as a HTTP proxy. Otherwise, this instrumentation should not be included. Something like that

NathanielRN: That makes sense to me. What do you think about 

```
"Extracting context from Lambda Event failed: either enable X-Ray active tracing or configure API Gateway to trigger this Lambda function as a pure proxy. Otherwise, generated spans will have an invalid (empty) parent context."
```

? I'm wondering if there's a use case where someone would want instrumentation and they're okay with the empty parent context. That's why I though "instrumentation should not be included" to not fit as well.

As does the `warning` level makes sense?

anuraaga: Hmm - I guess there is a use case where the function does not use X-Ray, and is not API Gateway but meant to always be a root - still it would be a bit weird to have no ability to parent at all but it's conceivable. Your message seems fine then, and yeah it should be debug level if there is a conceivable use case for it rather than always being in error.

NathanielRN: Yeah like if they just trigger it from the console is the easiest case I can think of. Maybe that Lambda triggers another Lambda 😝

I've made the updates 👍

anuraaga (approved):

PR #125, opened 2021-07-30
Move OTel wrapper py code into instrument script

NathanielRN:
This PR hopes to move some of the logic from `otel-wrapper.py` into `otel-instrument` so that we have only 1 file used for Python Lambda.

This was just an initial investigation by me, and requires much more rigorous testing before we should implement these changes.

Creating this as a draft until I can come back to work on it.


python/src/otel/otel_sdk/otel-instrument
```diff
@@ -40,5 +43,67 @@ environ["OTEL_RESOURCE_ATTRIBUTES"] = "%s,%s" % (
...
+_load_configurators()
+_load_instrumentors()
+
+# TODO: Move to opentelemetry-python-contrib so that adding
+# `opentelemetry-instrumentation-lambda` to the requirements.txt file will let
+# this automatically be picked up by `_load_instrumentors()` above
+AwsLambdaInstrumentor().instrument(skip_dep_check=True)
+
+# Start the runtime with the originally intended arguments
 system(" ".join(args))
```
NathanielRN: When asking if this is intended, got a reply from @wangzlei :

> Yes, I copied system(" ".join(args)) from Lambda python example, you can modify it.

and

> args is sys.argv[1:], should be same. Follow up from [a previous PR](https://github.com/open-telemetry/opentelemetry-lambda/pull/124#discussion_r680089874)

I think we should just do `system(" ".join(sys.argv[1:]))` but I'm concerned this will ignore edits to the `argv` commands made in the code above? But I also don't think there will be any edits... have to confirm.

NathanielRN:
Looks like our suspicious were true 😕 We were able to fix the dependency path, but its true that logging doesn't work to show the errors.

The script mysterious ends at this line:

```python
entry_point.load()().configure()  # type: ignore
```

Specifically at the `configure()` call. However, it ends with no log information as to what the problem is. It also doesn't seem to raise an Exception, it just ends...

Even if I remove the configuration, this line also ends without any trace

```python
conflict = get_dist_dependency_conflicts(entry_point.dist)
```

Closing this PR because there doesn't seem to be an obvious solution for now.

anuraaga:
The current code follows my mental model - after `system()` it's another process, so python code executed in this script wouldn't affect the "real" code that is run after `system()`. So the most that the instrument script is set up environment variables, or modify the command line (which it doesn't seem to currently do)

An improvement that would probably make this more clear, and would make this more consistent with other languages is to convert `otel-instrument` into a shell script. 

https://github.com/open-telemetry/opentelemetry-lambda/blob/main/java/layer-wrapper/scripts/otel-handler#L6

Mucking with env variables has native syntax in shell which can make things simpler, but more importantly also makes clear that this is just a script and not something we would want to run arbitrary python code in.

PR #128, opened 2021-08-12
Log when a custom config is being used.

anuraaga:
When users are using a custom config but it isn't working, there is always the possibility that the custom config isn't being picked up. With this log, we can confirm that and then debug elsewhere.


NathanielRN (approved):

ghost (commented):
LGTM

ghost (approved):

codeboten (approved):

PR #130, opened 2021-08-13
Allow setting the args of the collector via environment variable.

### LINKED ISSUE #63: Ability to set log-level of the collector
OTel collector only accepts log-level as a command line flag. But since extensions are automatically started by lambda, there doesn't seem to be any way to change this. Being able to set log-level is important when debugging issues such as connectivity to backends, creds, etc. Perhaps we need to change the extension from the Go binary itself to a shell script which can have arguments passed via environment variable.

anuraaga:
For example `OPENTELEMETRY_COLLECTOR_ARGS=--loglevel=debug`.

Also removes the current `metrics-level` default setting which seems arbitrary, the collector extension should just run with collector defaults.

Fixes #63

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -73,3 +73,16 @@ You can configure environment variables via CloudFormation template as well:
         Variables:
           OPENTELEMETRY_COLLECTOR_CONFIG_FILE: /var/task/collector.yaml
 ```
+
+You can configure arguments passed to the collector command line with the
+`OPENTELEMETRY_COLLECTOR_ARGS` environment variable.
+
+```yaml
+  Function:
+    Type: AWS::Serverless::Function
+    Properties:
+      ...
+      Environment:
+        Variables:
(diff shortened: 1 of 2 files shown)


willarmiros (approved):
LGTM, but let's wait til @wangzlei reviews on Monday before merging.

anuraaga:
@wangzlei Can you review this?

wangzlei (approved):
Thanks! that is an enhancement we expect.
Better to add instruction in README.

maxdumas:
@willarmiros @wangzlei It looks like this change was blown away by commit https://github.com/open-telemetry/opentelemetry-lambda/commit/db6ff3a488e0d54f2f979f7540c5a5baccef5fe3. Is there a reason for this? I'd really love to be able to enable the debug logging using this. If there was a reason this change was reverted, is there another way that I can enable that debug logging?

PR #138, opened 2021-08-31
Python layer update

bhautikpip:
1.  Adds `aiohttp` instrumentation deps to fix downstream CD failures
2. Added instrumentation code from upstream


python/src/otel/otel_sdk/requirements-nodeps.txt
```diff
@@ -1,4 +1,6 @@
 opentelemetry-instrumentation-aiohttp-client==0.24b0
+opentelemetry-util-http==0.24b0
+asgiref~=3.0
```
NathanielRN, line 3: Nit, we should group these 3 dependencies together with a comment. Also we could use lexicographical where possible here (and opentelemetry-util-http==0.24b0 depends on asgiref anyways?).

```suggestion
# All these packages are required for aiohttp instrumentation

asgiref~=3.0
opentelemetry-util-http==0.24b0
opentelemetry-instrumentation-aiohttp-client==0.24b0
```

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -17,32 +18,53 @@
...
-
-def _load_distros():
+def _load_distros() -> BaseDistro:
     for entry_point in iter_entry_points("opentelemetry_distro"):
         try:
-            entry_point.load()().configure()  # type: ignore
-            logger.info("Distribution %s configured", entry_point.name)
+            distro = entry_point.load()()
+            if not isinstance(distro, BaseDistro):
+                logger.debug(
```
NathanielRN, line 26: Should we create a follow up issue to find out why these debug logs can’t be seen in Lambda? But print statements CAN be seen in Lambda?

NathanielRN (approved):
Great job with this fix 🥳

wangzlei (approved):

python/src/otel/otel_sdk/otel_wrapper.py
wangzlei: yes please. No idea why the log disappear now.

wangzlei (approved):

PR #140, opened 2021-09-02
feat(collector): remove compatibility runtimes

longility:


collector/Makefile
```diff
@@ -29,7 +29,7 @@ publish-layer: package
 	@echo Publishing collector extension layer...
 	aws s3 mb s3://$(BUCKET_NAME)
 	aws s3 cp $(BUILD_SPACE)/collector-extension.zip s3://$(BUCKET_NAME)
-	aws lambda publish-layer-version --layer-name $(LAYER_NAME) --content S3Bucket=$(BUCKET_NAME),S3Key=collector-extension.zip --compatible-runtimes nodejs12.x nodejs10.x java11 python3.8 --query 'LayerVersionArn' --output text
+	aws lambda publish-layer-version --layer-name $(LAYER_NAME) --content S3Bucket=$(BUCKET_NAME),S3Key=collector-extension.zip --compatible-runtimes nodejs14.x nodejs12.x nodejs10.x java11 python3.8 --query 'LayerVersionArn' --output text
```
anuraaga: @wangzlei Should we just remove the flag? We don't use it when publishing SDK layers I think

willarmiros: Yeah I'm in favor of removing - the [Lambda docs](https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-create) make it pretty clear it's optional

longility: I updated the PR to where I removed the compatibility runtimes arg.

anuraaga (approved):
Thanks @longility!

collector/Makefile
wangzlei: If removing these flags, the layer would not be showed in Lambda console drop-down list, user has to add layer by layer ARN. I guess that is the reason @longility add nodejs14.x
But Lambda also has a limit that a layer has 5 runtime flag at most. So I would prefer add 5 most popular runtimes.

anuraaga: Is the option only about what shows in the console, not what can be applied by specifying arn manually? If so then that makes sense too (though the name of the flag would be quite confusing...)

wangzlei: Seems recently Lambda has increased compatible runtime to 15
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-layerversion.html#aws-resource-lambda-layerversion-properties

wangzlei: User can add layer either by full ARN or selecting a layer from a drop-list in Lambda console, the drop-list only shows the runtime compatible layers. 
Last year when I worked on python support, a layer had 5 compatible-runtimes at most, I was plagued by which 5 rumtimes I grant to Collector layer :)

longility: Is our primary audience going to be going through a drop list to adding or setting the ARN value in a config?

My gut tells me the latter.

Another option that adds complexity is to have one for each language in order to fully support all runtimes.

Yeah, I actually forked the repo for now so I can tweak and use the collector extension layer. Having an ARN would be nice so I don’t have to maintain it. I wouldn’t be picking from a drop down but setting the ARN value as long as I can find that value. Where do I go to find that? (I’m still fairly new to the AWS world)

wangzlei: For advanced users getting used to manage infrastructure by code, adding Lambda layer by ARN is ok without compatible-runtime flags.
At the same time, we hope to provide the public user guide to get users started in console, the console user experience is considered as well. 
Adding layer by ARN and by console drop-down are not incompatible. Having only compatible-runtime node14, the layer still can be added to python runtime by ARN. So, I would suggest keep your first commit, add `nodejs14.x` in cli.

`Another option that adds complexity is to have one for each language in order to fully support all runtimes.` -- that is a good point!

wangzlei: @anuraaga @longility no feedback for a long time :) Let's align with the conclusion.
My opinion is using the first commit, adding `nodejs14.x` in compatible runtime to help on console user experience.
If remove that flag, console user cannot pick up this layer from drop-down list `custom layer`
Screen Shot 2021-09-30 at 12 01 49 PM

longility: What is the common use case? Are we expecting a user to create their own lambda layer? If so, then I would leave the runtimes to have that shown in dropdown. If the lambda layer is mostly created and others are using it by specifying the ARN, then I would probably remove compatibility runtimes to keep it simple??

wangzlei: We have a long term goal that moving otel layers to `AWS layers`, the iterm on `Custom layers` left, users no need build layer on their own accounts.

PR #143, opened 2021-09-22
chore: updated collector to v0.36.0

willarmiros:
Updates to upstream collector v0.36.0. I was pretty confident that I should be adding the newly separated components to `collector/lambdacomponens/go.mod` instead of `collector/go.mod`, but would appreciate if someone could double check.


collector/lambdacomponents/go.mod
```diff
@@ -2,4 +2,13 @@ module github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents
 
 go 1.16
```
anuraaga, line 3: Think this file's go.sum isn't updated yet

collector/lambdacomponents/go.mod
```diff
@@ -2,4 +2,13 @@ module github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents
...
 
-require go.opentelemetry.io/collector v0.33.0
+require (
+  go.opentelemetry.io/collector v0.36.0
+  github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter v0.36.0
```
anuraaga, line 7: Just wondering where did this list come from?

collector/lambdacomponents/go.mod
willarmiros: Should be good now.

willarmiros:
Fixed the breaking change with the new `Command` interface by referencing this PR: https://github.com/open-telemetry/opentelemetry-collector/pull/4074

collector/lambdacomponents/go.mod
willarmiros: It's based on all the components here that were no longer in the core collector: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/lambdacomponents/default.go#L19-L30

collector/go.mod
```diff
@@ -5,6 +5,21 @@ go 1.16
 replace github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents => ./lambdacomponents
 
 require (
+	github.com/cespare/xxhash/v2 v2.1.2 // indirect
```
willarmiros, line 8: Is there any reason not to have these indirect imports? I noticed we were deleting them before.

anuraaga: We should keep whatever go adds

anuraaga (approved):

PR #145, opened 2021-09-22
chore: updated python deps

willarmiros:
Updated AWS extension to 1.0.1 stable version. No further upstream releases.

### DIFF
--- a/python/src/otel/otel_sdk/requirements.txt
+++ b/python/src/otel/otel_sdk/requirements.txt
@@ -3,4 +3,4 @@ opentelemetry-sdk==1.5.0
 opentelemetry-exporter-otlp==1.5.0
 opentelemetry-distro==0.24b0
 opentelemetry-instrumentation==0.24b0
-opentelemetry-sdk-extension-aws==0.24b0
+opentelemetry-sdk-extension-aws==1.0.1
--- a/python/src/otel/tests/test-requirements.txt
+++ b/python/src/otel/tests/test-requirements.txt
@@ -2,5 +2,5 @@ opentelemetry-api==1.5.0
 opentelemetry-sdk==1.5.0
 opentelemetry-distro==0.24b0
 opentelemetry-instrumentation==0.24b0
-opentelemetry-sdk-extension-aws==0.24b0
+opentelemetry-sdk-extension-aws==1.0.1
 pytest


anuraaga (approved):

python/src/otel/otel_sdk/requirements.txt
```diff
@@ -3,4 +3,4 @@ opentelemetry-sdk==1.5.0
 opentelemetry-exporter-otlp==1.5.0
 opentelemetry-distro==0.24b0
 opentelemetry-instrumentation==0.24b0
-opentelemetry-sdk-extension-aws==0.24b0
\ No newline at end of file
+opentelemetry-sdk-extension-aws==1.0.1
```
NathanielRN, line 6: Now that this is `1.0.1`, it might be a good idea to _only_ install `opentelemetry-sdk-extension-aws` like we do in the [auto instrumentation example app](https://github.com/aws-observability/aws-otel-python/blob/653f11725c03366afda26004b7323dba1ac7deae/integration-test-apps/auto-instrumentation/flask/requirements.txt#L5-L6).

Because, downloading `opentelemetry-sdk-extension-aws` will automatically install `opentelemetry-api`, `opentelemetry-sdk`, and `opentelemetry-instrumentation`. The `opentelemetry-sdk-extension-aws` package pins a version of each of these, and if that doesn't match with the pins from above, there would be a Version Conflict error.

python/src/otel/tests/test-requirements.txt
```diff
@@ -2,5 +2,5 @@ opentelemetry-api==1.5.0
 opentelemetry-sdk==1.5.0
 opentelemetry-distro==0.24b0
 opentelemetry-instrumentation==0.24b0
-opentelemetry-sdk-extension-aws==0.24b0
+opentelemetry-sdk-extension-aws==1.0.1
```
NathanielRN, line 5: Same comment as above.

python/src/otel/otel_sdk/requirements.txt
willarmiros: I didn't realize this! We'll proceed with it as is for the current release, but feel free to open a PR to update for future releases if it accomplishes the same with less.

NathanielRN: We went ahead with this change in #148

PR #148, opened 2021-09-22
Remove explicit deps on OTel Python Core pkgs

NathanielRN:
# Description

Follow up to the conversation on https://github.com/open-telemetry/opentelemetry-lambda/pull/145#discussion_r714246315 with @willarmiros, where we mentioned that because we already install `opentelemetry-sdk-extension==1.0.1`, we get the following packages installed as dependencies.
* `opentelemetry-api`
* `opentelemetry-sdk`
* `opentelemetry-instrumentation`

This is the preferable pattern because we can get a version conflict otherwise.

### DIFF
--- a/python/src/otel/otel_sdk/requirements.txt
+++ b/python/src/otel/otel_sdk/requirements.txt
@@ -1,6 +1,3 @@
-opentelemetry-api==1.5.0
-opentelemetry-sdk==1.5.0
 opentelemetry-exporter-otlp==1.5.0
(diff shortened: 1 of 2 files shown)


codeboten (approved):
👍

willarmiros (approved):
Thanks!

willarmiros:
@NathanielRN I updated the wording of the Python layer contents description, let me know if you think it makes sense: https://github.com/aws-observability/aws-otel-lambda/pull/154

PR #150, opened 2021-10-05
Refactor tests to match OTel Python format

NathanielRN:
# Description

To make adding future unit tests more easily, and to anticipate a move of this Lambda instrumentation to the upstream OTel Python repo, we format the tests to follow the format of upstream.

To run these tests locally you can run

`CORE_REPO_SHA=10208c1be1e720925a80a66f711b8afbe67537f4 tox -e test-instrumentation-aws_lambda`


anuraaga (approved):

wangzlei (approved):

python/src/tox.ini
```diff
@@ -14,11 +15,21 @@ passenv = TOXENV
...
+
+    aws_lambda: pip install {toxinidir}/otel[test]
 
 deps = 
-    -r{toxinidir}/otel/tests/test-requirements.txt
+    test: pytest
 
 commands =
-    pytest {toxinidir}/otel/tests/ -s
+    test: pytest {posargs} -s
```
NathanielRN: @wangzlei Do you know if we need this `-s` (i.e. `--capture=no`) configuration flag for `pytest`?

```
$ pytest --help
usage: pytest [options] [file_or_dir] [file_or_dir] [...]

positional arguments:
  file_or_dir

general:
  --capture=method      per-test capturing method: one of fd|sys|no|tee-sys.
  -s                    shortcut for --capture=no.
```

I don't think we need it so I'd like to remove it because we get an error message even though the tests work fine:

```
test_otel.py sh: -c: option requires an argument
```

wangzlei: I added `-s` for printing log in console. If it does not make sense please remove it.

wangzlei (approved):

PR #154, opened 2021-10-13
Init TracerProvider ourselves to set immutable Resource attribute

NathanielRN:
# Description

We are getting ready to upstream this `AwsLambdaInstrumentor` to [the OpenTelemetry Python Contrib repo](https://github.com/open-telemetry/opentelemetry-python-contrib/). In doing so, there is a caveat because one of the `Resource` attributes cannot be set until the Lambda function is actually invoked and the `lambda_context` is available.

This PR tries to solve that. Instead of letting the `opentelemetry_configurator` entry point configure the `TracerProvider`. We take the relevant methods and try to initialize the `TraceProvider` ourselves. We must, because we can't initialize again after it's done. It is immutable (and so is its `Resource`).

There are some other miscellaneous formatting changes and doc changes. If it's too busy I can remove these and focus on this solution first!

### DIFF
--- a/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
+++ b/python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
@@ -62,7 +62,7 @@ from opentelemetry.sdk.extension.aws.trace.propagation.aws_xray_format import (
 from opentelemetry.semconv.trace import SpanAttributes
 from opentelemetry.trace import (
     SpanKind,
-    Tracer,
+    TracerProvider,
     get_tracer,
     get_tracer_provider,
 )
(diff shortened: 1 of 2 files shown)


python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
```diff
@@ -189,37 +198,35 @@ def _instrumented_lambda_handler_call(call_wrapped, instance, args, kwargs):
...
             if span.is_recording():
                 lambda_context = args[1]
-                # Refer: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/faas.md#example
+                # Refer: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/faas.md#example
                 span.set_attribute(
                     SpanAttributes.FAAS_EXECUTION, lambda_context.aws_request_id
                 )
-                span.set_attribute(
-                    "faas.id", lambda_context.invoked_function_arn
-                )
```
NathanielRN, line 207: This attribute is what motivated the large changes in this PR. Semantic conventions tells us this should be a `ResourceAttribute` and not a `SpanAttribute`. `Resource` can only be set once, and the `lambda_context` isn't available until later. That's why we have to do all this.

I wonder if this will make the Lambda take much longer if we can't initialize the `TracerProvider` before the Lambda gets invoked. My gut reaction says it won't take too much longer.

NathanielRN:
@wangzlei @anuraaga Still thinking this through and will have to test it. But would appreciate your initial thoughts? Especially because there are a lot of changes needed just to get 1 attribute on the spans.

python/src/otel/otel_sdk/opentelemetry/instrumentation/aws_lambda/__init__.py
anuraaga: The spec explicitly makes an exception for `faas.id`

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/faas.md#function-as-a-service

`On some providers, it may not be possible to determine the full ID at startup, which is why this field cannot be made required. For example, on AWS the account ID part of the ARN is not available without calling another AWS API which may be deemed too slow for a short-running lambda function. As an alternative, consider setting faas.id as a span attribute instead.`

Losing out on the standard configuration mechanism along with all the code overhead isn't worth it I think vs just using a span attribute instead, which we do in Java / JS. It's precisely what the spec is trying to avoid

Ideally Lambda team implemented the ticket I filed about a year ago to make this available in an environment variable :)

NathanielRN: Ah, this is an incredibly useful comment, thank you! I'll put is as a span attribute then 🙂

NathanielRN:
Closing because as mentioned above, the specs make an exception for an AWS Lambda instrumentation.

PR #158, opened 2021-10-21
Using upstream OTel auto instrumentation and minimize otel_wrapper

### LINKED ISSUE #152: Use opentelemetry-instrument script to invoke python app
Currently, we have a wrapper script for initializing the python function

https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/otel_sdk/otel_wrapper.py

I'm wondering if there is any known issue with instead invoking the standard `opentelemetry-instrument` script?

https://github.com/open-telemetry/opentelemetry-python/tree/main/opentelemetry-instrumentation#opentelemetry-instrument

Perhaps it's related to the two TODOs?

https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/otel_sdk/otel_wrapper.py#L15

If we could leverage `opentelemetry-instrument`, it ensures there is only one consistent and officially maintained method for enabling Python instrumentation.

Note I am thinking of this in the context of adding auto instrumentation to k8s - I think the concept of updating the runtime for auto instrumentation in a k8s pod and a lambda function are almost, if not always, equivalent.

https://github.com/open-telemetry/opentelemetry-operator/issues/455

@wangzlei @NathanielRN Any ideas?

NathanielRN:
# Description

@wangzlei and I worked together to figure out how to _mostly_ auto-instrument Lambda functions using the upstream `opentelemetry-instrument` [auto-instrumentation package](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/3049b4bfc5e2b64679f27543c1bd9d220047626e/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation).

We made two major updates

## Update the PYTHONPATH so we can call `opentelemtry-instrument`

The auto-instrumentation package counts on 2 things:
* the `opentelemetry` package
* locating all the python packages the user wants to instrument

Both these 2 things require us modifying the environment variable `PYTHONPATH` right away in `otel-instrument`. AWS Lambda will add the correct paths but it does it too late. (It only does it once it calls it's originally intended entry point of `python3 /var/runtime/bootstrap.py`).

## Update the `otel_wrapper.py` to only call the `AwsLambdaInstrumentor`

All of the instrumentation is done in `sitecustomize.py` by `opentelemetry-instrument`. However, the way the Lambda Handler is imported in the AWS Lambda `bootstrap.py` file which is run after `sitecustomize.py` CLEARS any instrumentation done on `lambda_function.lambda_handler`. That's why we keep `otel_wrapper.py` around. So that `bootstrap.py` can call it, do any destructive imports it needs to do, and then call `AwsLambdaInstrumentor.instrument()` explicitly. This way we _know_ the Lambda function is instrumented, we can import, and give `bootstrap.py` the Lambda Handler that we know is instrumented.

## Future Work

We would ideally like to modify `boostrap.py` or investigate how we can get rid of `otel_wrapper.py` and have all the instrumentation be finished in `sitecustomize.py` such that `bootstrap.py` can just import the normal lambda handler at the `_HANDLER` environment variable without destroying the import at all.

Fixes #152

### DIFF
--- a/python/src/otel/otel_sdk/otel-instrument
+++ b/python/src/otel/otel_sdk/otel-instrument
@@ -23,14 +23,17 @@ variables OpenTelemetry uses, and then initializing OpenTelemetry using the
 `opentelemetry-instrument` auto instrumentation script from the
 `opentelemetry-instrumentation` package.
 
-Additionally, this configuration the user is using packages derived from
-`opentelemetry-instrumentation` and `opentelemetry-sdk` in their implementation.
+Additionally, this configuration assumes the user is using packages conforming
+to the `opentelemetry-instrumentation` and `opentelemetry-sdk` specifications.
+
+DO NOT use this script for anything else besides SETTING ENVIRONMENT VARIABLES.
+The final `execl` command starts a new python process which replaces the current
+one. One environment variables are preserved.
 
 Usage
 -----
-In the configuration of an AWS Lambda function with
-`otel-instrument` at the root level of a Lambda
-Layer:
+In the configuration of an AWS Lambda function with this file at the
+root level of a Lambda Layer:
 
 .. code::
 
@@ -48,13 +51,13 @@ LAMBDA_RUNTIME_DIR = "LAMBDA_RUNTIME_DIR"
 
 # Update the python paths for packages with `sys.path` and `PYTHONPATH`
 
-# - Expect this `otel-instrument` file to be at the Lambda Layer root. Then, we
-#   know where the OpenTelemetry Python packages are and can add them to the
-#   PYTHONPATH.
+# - Expect this file to be at the Lambda Layer root. Then, we know where the
+#   OpenTelemetry Python packages are and can add them to the PYTHONPATH.
+#
+#   See more:
+#   https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-path
 
-LAMBDA_LAYER_PKGS_DIR = os.path.abspath(
-    os.path.join(os.path.dirname(__file__), "python")
-)
+LAMBDA_LAYER_PKGS_DIR = os.path.abspath(os.path.join(os.sep, "opt", "python"))
 
 # - Set Lambda Layer python packages in PYTHONPATH so `opentelemetry-instrument`
 #   script can find them (it needs to find `opentelemetry` find the auto
@@ -85,9 +88,11 @@ if LAMBDA_LAYER_PKGS_DIR not in sys.path:
 from opentelemetry.environment_variables import (
     OTEL_PROPAGATORS,
     OTEL_TRACES_EXPORTER,
-    OTEL_PYTHON_DISABLED_INSTRUMENTATIONS,
 )
-from opentelemetry.sdk.environment_variables import OTEL_RESOURCE_ATTRIBUTES
+from opentelemetry.sdk.environment_variables import (
+    OTEL_RESOURCE_ATTRIBUTES,
+    OTEL_SERVICE_NAME,
+)
 
 # - Set the default Trace Exporter
 
@@ -95,20 +100,7 @@ environ.setdefault(OTEL_TRACES_EXPORTER, "otlp_proto_grpc_span")
 
 # - Set the service name
 
-service_name_resource_attribute = "service.name"
-if environ.get(OTEL_RESOURCE_ATTRIBUTES) is None:
-    environ[OTEL_RESOURCE_ATTRIBUTES] = "%s=%s" % (
-        service_name_resource_attribute,
-        environ.get(AWS_LAMBDA_FUNCTION_NAME),
-    )
-elif service_name_resource_attribute not in environ.get(
-    OTEL_RESOURCE_ATTRIBUTES
-):
-    environ[OTEL_RESOURCE_ATTRIBUTES] = "%s=%s,%s" % (
-        service_name_resource_attribute,
-        environ.get(AWS_LAMBDA_FUNCTION_NAME),
-        environ.get(OTEL_RESOURCE_ATTRIBUTES),
-    )
+environ.setdefault(OTEL_SERVICE_NAME, environ.get(AWS_LAMBDA_FUNCTION_NAME))
 
 # - Set the Resource Detectors (Resource Attributes)
 #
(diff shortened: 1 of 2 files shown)


python/src/otel/otel_sdk/otel-instrument
```diff
@@ -1,44 +1,160 @@
...
-# enable OTel wrapper
-environ["ORIG_HANDLER"] = environ.get("_HANDLER")
-environ["_HANDLER"] = "otel_wrapper.lambda_handler"
+# Update the python paths for packages with `sys.path` and `PYTHONPATH`
+
+# - Expect this `otel-instrument` file to be at the Lambda Layer root. Then, we
+#   know where the OpenTelemetry Python packages are and can add them to the
+#   PYTHONPATH.
+
+LAMBDA_LAYER_PKGS_DIR = os.path.abspath(
```
anuraaga: I think this file would be located at `/var/task` - shouldn't this explicitly reference `/opt/python` which while not an environment variable is a well defined path for lambda?

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -1,44 +1,160 @@
...
+
+"""
+
+import os
 import sys
+from os import environ, execl
 
-# the path to the interpreter and all of the originally intended arguments
-args = sys.argv[1:]
+AWS_LAMBDA_FUNCTION_NAME = "AWS_LAMBDA_FUNCTION_NAME"
```
anuraaga, line 48: We'd expect these to actually refer to a value, not just the name itself. I'd just inline

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -1,44 +1,160 @@
...
+)
+from opentelemetry.sdk.environment_variables import OTEL_RESOURCE_ATTRIBUTES
+
+# - Set the default Trace Exporter
+
+environ.setdefault(OTEL_TRACES_EXPORTER, "otlp_proto_grpc_span")
+
+# - Set the service name
+
+service_name_resource_attribute = "service.name"
```
anuraaga: This seems to not respect the user's service.name

anuraaga: Actually just use OTEL_SERVICE_NAME instead, respecting the default if already set

https://github.com/open-telemetry/opentelemetry-python/blob/18b5cb046bebab0c70841c662268bb7980c2fd5e/opentelemetry-sdk/src/opentelemetry/sdk/environment_variables.py#L356

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -1,44 +1,160 @@
...
+    environ.get(OTEL_RESOURCE_ATTRIBUTES),
 )
 
-# start the runtime with the extra options
-system(" ".join(args))
+# - Set the default propagators
+
+environ.setdefault(OTEL_PROPAGATORS, "tracecontext,b3,xray")
+
+# - Disable Lambda instrumentation because we will instrument when we know it
```
anuraaga: We should just remove the auto instrumentation entry point from setup.cfg instead as it can't really be used

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -1,44 +1,160 @@
...
+#   patch in a way that is visible to the calling AWS Lambda `boostrap.py` file.
+
+environ[OTEL_PYTHON_DISABLED_INSTRUMENTATIONS] = "aws_lambda"
+
+# - Use a wrapper because instrumentations don't last across parent -> child
```
anuraaga: I don't think we need this comment, we already found enough confusion on the topic of parent / child already

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -1,44 +1,160 @@
...
+#   processes (i.e. using `excel`) if the child process imports packages
+#   using `getattr(module_name, function_name)`. The child will just import an
+#   uninstrumented package if we don't have `otel_wrapper.py`.
+
+environ["ORIG_HANDLER"] = environ.get("_HANDLER")
+environ["_HANDLER"] = "otel_wrapper.lambda_handler"
+
+# - Call the upstream auto instrumentation script
+
```
anuraaga, line 147: Strongly recommend rewriting this file into a shell script. It solidifies two facts

1) It's supposed to only do very simple things like setting environment variables, no business logic can happen in here
2) There is no parent python process in our lambda handler, the fact that this file is python I feel can confuse a reader into wondering why we don't load instrumentation here (I think you tried this in a PR once :P). A shell script makes obvious this isn't the app, it's just configuring the environment

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -1,101 +1,19 @@
...
-                entry_point.name,
-                configured,
-            )
-            continue
-        try:
-            entry_point.load()().configure()  # type: ignore
-            configured = entry_point.name
-        except Exception as exc:  # pylint: disable=broad-except
-            logger.debug("Configuration of %s failed", entry_point.name)
+AwsLambdaInstrumentor().instrument()
```
anuraaga, line 6: When upstreaming, though initially probably a PR for this repo since it's easier to test I guess, strongly recommend providing the original handler as a parameter to this function. That's all it should take to make the instrumentation work both manually, if a user wants to, or automatically from the lambda layer. We just want to decouple ORIG_HANDLER as it's a mysterious dependency between the two components.

I'd recommend sending a PR that only adds AwsLambdaInstrumenter file and has unit tests that exercise the instrumentation manually first to keep the scope small and easy to review.

anuraaga: I realize currently the instrumentation seems to handle by falling back to `_HANDLER` when `ORIG_HANDLER` isn't set. It works ok but less coupling will make the parts easier to follow. Instrumentation is about instrumenting the handler, wrapper is about setting up the instrumentation automatically in the runtime. The separation will improve the code.

python/src/otel/otel_sdk/otel-instrument
NathanielRN: We could do a different name for the constant if you think that's better? I think this pattern is a good idea though because we reference this variable in 3 different places and that's 3 different chances to have a typo (as has happened to me before).

It's also pretty normal in the upstream to do this? https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/environment_variables.py

python/src/otel/otel_sdk/otel-instrument
NathanielRN: > I think this file would be located at /var/task

Didn't you also say this in [#157](https://github.com/open-telemetry/opentelemetry-lambda/pull/157/files#diff-3bdb05ad1b8dfe19a9868aa352532bc42af15de7035c12c7b909a6475fd35206R15-R16)

> Note, the file structure is important. Lambda does not seem to allow loading the handler if it is placed at a subdirectory such as opentelemetry.instrumentation.lambda.handler, when inside a layer it must be at the top level.

so this file has to be moved to the top level as we do in the build:

https://github.com/open-telemetry/opentelemetry-lambda/blob/71221eea34b616df64e5a98736cc574146fd88a2/python/src/otel/Dockerfile#L15

> shouldn't this explicitly reference /opt/python which while not an environment variable is a well defined path for lambda?

Yeah we can do that, it was just annoying to test locally because `../python` always existed but `/opt/python` did not. I'll move it back though because that explains the intent well.

python/src/otel/otel_sdk/otel-instrument
anuraaga: Renaming to something else seems like it reduces readability of the code that uses the variable. Using variables that refer to names, instead of values, also reduces readability though. I disagree with the upstream pattern here.

anuraaga: For reference, readability always needs to trump writability - typos during writing is still sort of OK for readability sakes. In many cases it results in duplicate docs, etc that get out of sync but it is what it is. In this case, readability suffers from the constants since then you have to jump to the constant definition to know the value every time - using the string has only positive impacts on readibility though

python/src/otel/otel_sdk/otel-instrument
NathanielRN: I think this is the right idea, will do that!

python/src/otel/otel_sdk/otel-instrument
NathanielRN: Thanks for this, yes it is confusing, removing it and explaining it better.

python/src/otel/otel_sdk/otel-instrument
NathanielRN: There's been a lot of talk at the python SIG about moving _all_ bash scripts to python. Python does scripts really well! 

> 1. It's supposed to only do very simple things like setting environment variables, no business logic can happen in here

One huge benefit is being able to import other python packages to GET environment variable constants which like you said is the point of this file. Then we can be sure updates get picked up. (Although they are spec defined things in the instrumentation section are still changing).

All in all a python script should be easier to maintain by a python script.

> 2. There is no parent python process in our lambda handler, the fact that this file is python I feel can confuse a reader into wondering why we don't load instrumentation here (I think you tried this in a PR once :P). A shell script makes obvious this isn't the app, it's just configuring the environment

Lol yes I did but I learned a lot about the Lambda instrumentation in #152  🥲 I added an explicit comment at the top description. I feel like that was what the file was missing and with this it shouldn't confuse future readers.

python/src/otel/otel_sdk/otel_wrapper.py
NathanielRN: > I'd recommend sending a PR that only adds AwsLambdaInstrumenter file and has unit tests that exercise the instrumentation manually first to keep the scope small and easy to review.

Will definitely keep this in mind as I ask for feedback from the Python SIG. The actual code change in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/739 is small though and when I brought it up last week they said they would be able to help me review it! It's inflated by the LICENSE.

The tests already give an example of BOTH auto instrumentation AND manual instrumentation. Auto Instrumentation is a super important part to this product so I feel reluctant to commit it later when we _already_ have tests that can give us confidence the scripts is working as intended.

(I put in a lot of effort into #150 to make sure the tests use the script because it helps give us confidence that instrumentation works as intended in Lambda).

PR #160, opened 2021-10-22
feat(opentelemetry-lambda): env config + exporter extension hook

ghost:
Added env config for:
- logging (otel_log_level)

Plus additional hook for configuring TracerProvider (eg downstream exporters)

### DIFF
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -1,12 +1,28 @@
 import { NodeTracerConfig, NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
 import {
   BatchSpanProcessor,
+  ConsoleSpanExporter,
   SDKRegistrationConfig,
+  SimpleSpanProcessor,
 } from '@opentelemetry/sdk-trace-base';
+import { registerInstrumentations } from '@opentelemetry/instrumentation';
+import { awsLambdaDetector } from '@opentelemetry/resource-detector-aws';
+import {
+  detectResources,
+  envDetector,
+  processDetector,
+} from '@opentelemetry/resources';
+import { AwsInstrumentation } from '@opentelemetry/instrumentation-aws-sdk';
+import {
+  diag,
+  DiagConsoleLogger,
+  DiagLogLevel,
+} from "@opentelemetry/api";
+import { getEnv } from '@opentelemetry/core';
 
 // Use require statements for instrumentation to avoid having to have transitive dependencies on all the typescript
 // definitions.
-const{ AwsLambdaInstrumentation } = require('@opentelemetry/instrumentation-aws-lambda');
+const { AwsLambdaInstrumentation } = require('@opentelemetry/instrumentation-aws-lambda');
 const { DnsInstrumentation } = require('@opentelemetry/instrumentation-dns');
 const { ExpressInstrumentation } = require('@opentelemetry/instrumentation-express');
 const { GraphQLInstrumentation } = require('@opentelemetry/instrumentation-graphql');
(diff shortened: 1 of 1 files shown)


nodejs/packages/layer/src/wrapper.ts
```diff
@@ -59,27 +68,59 @@ const instrumentations = [
...
+const logLevel = getEnv().OTEL_LOG_LEVEL
+diag.setLogger(new DiagConsoleLogger(), logLevel)
+
 // Register instrumentations synchronously to ensure code is patched even before provider is ready.
 registerInstrumentations({
   instrumentations,
 });
 
+function configureSampler() : Sampler {
+  const sampler = getEnv().OTEL_TRACES_SAMPLER as TracesSamplerValues
```
anuraaga: Is it possible to implement this in the JS SDK itself, similar to how java and python use autoconfigure for env vars like sampler?

ghost: Yes, I think it should be a separate module - like "our own" Java SDK autoconf :) I'm however quite new in the otel/js world and this would definitely be a bigger undertaking. Let's do this in a following way:
- add minimal env var conf here for lambda purposes
- open an OTel JS issue to discuss / introduce autoconf for Node 
How does that sound?

anuraaga: Sure - before step 1 though I would confirm that JS doesn't already have such a module, and perhaps we are just missing something trivial to be able to use it. For example, a search brings up some code

https://github.com/open-telemetry/opentelemetry-js/search?q=OTEL_TRACES_SAMPLER

Can you confirm this variable isn't currently respected, or what it would take for that code to kick in if not?

ghost: Well, seems you're right. I did search for sth alike SDK autocofigure but using different env keys (EXPORTER, LOG) but could not find anything relevant. Yet, the SAMPLER has built in support. Hence I'm removing this and leaving only extension hook and log/diag support.

ghost:
@anuraaga PTAL, removed SAMPLER config (as already implemented in the node.js distro).

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -70,16 +76,25 @@ async function initializeProvider() {
...
   const tracerProvider = new NodeTracerProvider(config);
-  tracerProvider.addSpanProcessor(
-    new BatchSpanProcessor(new CollectorTraceExporter())
-  );
+  if (typeof configureTracerProvider === 'function') {
+    configureTracerProvider(tracerProvider)
+  } else {
+    // defaults
+    tracerProvider.addSpanProcessor(
+      new BatchSpanProcessor(new CollectorTraceExporter())
```
anuraaga: This was needed at the time but perhaps it can be removed too now? BSP + OTLP exporter is supposed to be the default for OTel SDKs I think

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -59,6 +61,10 @@ const instrumentations = [
   new RedisInstrumentation(),
 ];
 
+// configure lambda logging
+const logLevel = getEnv().OTEL_LOG_LEVEL
+diag.setLogger(new DiagConsoleLogger(), logLevel)
```
anuraaga, line 74: It seems very likely this is already handled by the SDK

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -1,12 +1,21 @@
 import { NodeTracerConfig, NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
 import {
   BatchSpanProcessor,
+  ConsoleSpanExporter,
   SDKRegistrationConfig,
+  SimpleSpanProcessor
```
anuraaga: We don't currently run linter on the build, though probably should. But if possible would be good to stick to keeping the trailing comma for all elements of a collection
```suggestion
  SimpleSpanProcessor,
```

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -70,16 +76,25 @@ async function initializeProvider() {
   });
 
   let config: NodeTracerConfig = {
-    resource,
+    resource
```
anuraaga:
```suggestion
    resource,
```

anuraaga (approved):

nodejs/packages/layer/src/wrapper.ts
anuraaga: Dug into the JS codebase a bit and find that this still isn't handled using env vars it seems, hopefully someday.

anuraaga (commented):
LGTM

nodejs/packages/layer/src/wrapper.ts
ghost: I tried to find it (this time ;) ) but nothing popped up. That is I'm fairly certain that the SDK does not set diag ConsoleLogger based on the env value anywhere.

ghost:
Rebased to main && resolved conflict.

ghost:
@codeboten @wangzlei please have a look

ghost:
Can we get this reviewed && (merged || rejected)?  @wangzlei @codeboten ?

wangzlei (approved):
Thanks!

PR #164, opened 2021-10-25
Lambda exec wrapper calls upstream OTel Python auto instr script

### LINKED ISSUE #152: Use opentelemetry-instrument script to invoke python app

NathanielRN:
# Description

@wangzlei and I worked together to figure out how to _mostly_ auto-instrument Lambda functions using the upstream `opentelemetry-instrument` [auto-instrumentation package](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/3049b4bfc5e2b64679f27543c1bd9d220047626e/opentelemetry-instrumentation/src/opentelemetry/instrumentation/auto_instrumentation).

In doing so, we found it best to re-write `otel-instrument` as a bash script.

We made two major updates

## Update the PYTHONPATH in `otel-instrument` so we can call `opentelemtry-instrument`

The auto-instrumentation package counts on 2 things:
* the `opentelemetry` package
* locating all the python packages the user wants to instrument

Both these 2 things require us modifying the environment variable `PYTHONPATH` right away in the `otel-instrument` script. AWS Lambda will add the correct paths but it does it too late. (It only does it once it calls it's originally intended entry point of `python3 /var/runtime/bootstrap.py`).

## Update the `otel_wrapper.py` to only call the `AwsLambdaInstrumentor`

All of the instrumentation is done in `sitecustomize.py` by `opentelemetry-instrument`. However, the way the Lambda Handler is imported in the AWS Lambda `bootstrap.py` file which is run after `sitecustomize.py` CLEARS any instrumentation done on `lambda_function.lambda_handler`. That's why we keep `otel_wrapper.py` around. So that `bootstrap.py` can call it, do any destructive imports it needs to do, and then call `AwsLambdaInstrumentor.instrument()` explicitly. This way we _know_ the Lambda function is instrumented, we can import, and give `bootstrap.py` the Lambda Handler that we know is instrumented.

## Future Work

We would ideally like to modify AWS Lambda's `boostrap.py` or investigate how we can get rid of `otel_wrapper.py` and have all the instrumentation be finished in `sitecustomize.py` such that `bootstrap.py` can just import the normal lambda handler at the `_HANDLER` environment variable without destroying the import at all.

Fixes #152

### DIFF
--- a/python/src/otel/otel_sdk/otel-instrument
+++ b/python/src/otel/otel_sdk/otel-instrument
@@ -100,13 +100,8 @@ fi
 #
 #   export OTEL_RESOURCE_DETECTORS="aws_lambda";
 #
-export LAMBDA_RESOURCE_ATTRIBUTES="cloud.region=$AWS_REGION,cloud.provider=aws,faas.name=$AWS_LAMBDA_FUNCTION_NAME,faas.version=$AWS_LAMBDA_FUNCTION_VERSION";
 
-if [ -z ${OTEL_RESOURCE_ATTRIBUTES} ]; then
-    export OTEL_RESOURCE_ATTRIBUTES=$LAMBDA_RESOURCE_ATTRIBUTES;
-else
-    export OTEL_RESOURCE_ATTRIBUTES="$LAMBDA_RESOURCE_ATTRIBUTES,$OTEL_RESOURCE_ATTRIBUTES";
-fi
+export OTEL_RESOURCE_ATTRIBUTES="cloud.region=$AWS_REGION,cloud.provider=aws,faas.name=$AWS_LAMBDA_FUNCTION_NAME,faas.version=$AWS_LAMBDA_FUNCTION_VERSION,$OTEL_RESOURCE_ATTRIBUTES";
 
 # - Uses the default `OTEL_PROPAGATORS` which is set to `tracecontext,baggage`
 
--- a/python/src/otel/otel_sdk/otel-instrument
+++ b/python/src/otel/otel_sdk/otel-instrument
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
 # Copyright The OpenTelemetry Authors
 #
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-: '
+: <<'END_DOCUMENTATION'
 `otel-instrument`
 
 This script configures and sets up OpenTelemetry Python with the values we
@@ -43,7 +43,7 @@ root level of a Lambda Layer:
 
     AWS_LAMBDA_EXEC_WRAPPER = /opt/otel-instrument
 
-'
+END_DOCUMENTATION
 
 # Use constants to access the environment variables we want to use in this
 # script.
@@ -68,17 +68,13 @@ root level of a Lambda Layer:
 #   See more:
 #   https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-path
 
-export LAMBDA_LAYER_PKGS_DIR="/opt/python"
+export LAMBDA_LAYER_PKGS_DIR="/opt/python";
 
 # - Set Lambda Layer python packages in PYTHONPATH so `opentelemetry-instrument`
 #   script can find them (it needs to find `opentelemetry` to find the auto
 #   instrumentation `run()` method later)
 
-if [ -z ${PYTHONPATH} ]; then
-    export PYTHONPATH=$LAMBDA_LAYER_PKGS_DIR;
-else
-    export PYTHONPATH="$LAMBDA_LAYER_PKGS_DIR:$PYTHONPATH";
-fi
+export PYTHONPATH="$LAMBDA_LAYER_PKGS_DIR:$PYTHONPATH";
 
 # - Set Lambda runtime python packages in PYTHONPATH so
 #   `opentelemetry-instrument` script can find them during auto instrumentation
@@ -88,11 +84,7 @@ export PYTHONPATH="$LAMBDA_RUNTIME_DIR:$PYTHONPATH";
 
 # Configure OpenTelemetry Python with environment variables
 
-# - Set the default Trace Exporter
-
-if [ -z ${OTEL_TRACES_EXPORTER} ]; then
-    export OTEL_TRACES_EXPORTER="otlp_proto_grpc_span";
-fi
+# - Uses the default `OTEL_TRACES_EXPORTER` which is set to `otlp_proto_grpc`
 
 # - Set the service name
 
@@ -108,7 +100,7 @@ fi
 #
 #   export OTEL_RESOURCE_DETECTORS="aws_lambda";
 #
-export LAMBDA_RESOURCE_ATTRIBUTES="cloud.region=$AWS_REGION,cloud.provider=aws,faas.name=$AWS_LAMBDA_FUNCTION_NAME,faas.version=$AWS_LAMBDA_FUNCTION_VERSION"
+export LAMBDA_RESOURCE_ATTRIBUTES="cloud.region=$AWS_REGION,cloud.provider=aws,faas.name=$AWS_LAMBDA_FUNCTION_NAME,faas.version=$AWS_LAMBDA_FUNCTION_VERSION";
 
 if [ -z ${OTEL_RESOURCE_ATTRIBUTES} ]; then
     export OTEL_RESOURCE_ATTRIBUTES=$LAMBDA_RESOURCE_ATTRIBUTES;
@@ -116,11 +108,7 @@ else
     export OTEL_RESOURCE_ATTRIBUTES="$LAMBDA_RESOURCE_ATTRIBUTES,$OTEL_RESOURCE_ATTRIBUTES";
 fi
 
-# - Set the default propagators
-
-if [ -z ${OTEL_PROPAGATORS} ]; then
-    export OTEL_PROPAGATORS="tracecontext,b3,xray";
-fi
+# - Uses the default `OTEL_PROPAGATORS` which is set to `tracecontext,baggage`
 
 # - Use a wrapper because AWS Lambda's `python3 /var/runtime/bootstrap.py` will
 #   use `imp.load_module` to load the function from the `_HANDLER` environment
@@ -137,5 +125,3 @@ export _HANDLER="otel_wrapper.lambda_handler";
 # - Call the upstream auto instrumentation script
 
 python3 $LAMBDA_LAYER_PKGS_DIR/bin/opentelemetry-instrument "$@"
-
-
--- a/python/src/otel/otel_sdk/otel-instrument
+++ b/python/src/otel/otel_sdk/otel-instrument
@@ -1,44 +1,141 @@
-#!/usr/bin/env python3
-# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-# SPDX-License-Identifier: MIT-0
+#!/usr/bin/env bash
 
-from os import environ, system
-import sys
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


python/src/otel/otel_sdk/otel-instrument
```diff
@@ -1,44 +1,141 @@
...
+#   See more:
+#   https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-path
+
+export LAMBDA_LAYER_PKGS_DIR="/opt/python"
+
+# - Set Lambda Layer python packages in PYTHONPATH so `opentelemetry-instrument`
+#   script can find them (it needs to find `opentelemetry` to find the auto
+#   instrumentation `run()` method later)
+
+if [ -z ${PYTHONPATH} ]; then
```
anuraaga: For any path type variabe it's conventional to not check for presence, empty strings are fine (they should be made fine for the resource attributes variable in the python SDK separately too at some point)

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -1,44 +1,141 @@
...
+#   and instrument them.
+
+export PYTHONPATH="$LAMBDA_RUNTIME_DIR:$PYTHONPATH";
+
+# Configure OpenTelemetry Python with environment variables
+
+# - Set the default Trace Exporter
+
+if [ -z ${OTEL_TRACES_EXPORTER} ]; then
+    export OTEL_TRACES_EXPORTER="otlp_proto_grpc_span";
```
anuraaga: Per the otel spec, this should be the default for the SDK - perhaps it wasn't a while ago but is this logic still needed?

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -1,44 +1,141 @@
...
+if [ -z ${OTEL_RESOURCE_ATTRIBUTES} ]; then
+    export OTEL_RESOURCE_ATTRIBUTES=$LAMBDA_RESOURCE_ATTRIBUTES;
+else
+    export OTEL_RESOURCE_ATTRIBUTES="$LAMBDA_RESOURCE_ATTRIBUTES,$OTEL_RESOURCE_ATTRIBUTES";
+fi
+
+# - Set the default propagators
+
+if [ -z ${OTEL_PROPAGATORS} ]; then
+    export OTEL_PROPAGATORS="tracecontext,b3,xray";
```
anuraaga: xray is directly used by the instrumentation, the global propagators should not have xray enabled by default in an upstream distribution. Per spec, the otel default is OTEL_PROPAGATORS=tracecontext,baggage, assuming the python SDK uses that default we shouldn't need to set anything, otherwise we should use tracecontext,baggage

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -1,101 +1,55 @@
...
-
-    for entry_point in iter_entry_points("opentelemetry_instrumentor"):
-        if entry_point.name in package_to_exclude:
-            logger.debug(
-                "Instrumentation skipped for library %s", entry_point.name
-            )
-            continue
-
-        try:
-            conflict = get_dist_dependency_conflicts(entry_point.dist)
```
anuraaga, line 54: 🎉

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -1,101 +1,55 @@
...
-distro.configure()
-_load_configurators()
-_load_instrumentors(distro)
-# TODO: move to python-contrib
-AwsLambdaInstrumentor().instrument(skip_dep_check=True)
+
+AwsLambdaInstrumentor().instrument()
 
 path = os.environ.get("ORIG_HANDLER", None)
 if path is None:
```
anuraaga, line 56: It might be good to reproduce the error messages bootstrap.py would produce when validating `_HANDLER`

python/src/otel/otel_sdk/otel-instrument
NathanielRN: Okay sounds good! Yeah I didn't know if user would dislike us modifying a variable they already set, but then again they're probably using this script BECAUSE they want us to handle all these little settings 😛

Changed it to this!

```bash
export PYTHONPATH="$LAMBDA_LAYER_PKGS_DIR:$PYTHONPATH";
```

And I guess you mean in the future someone can do this:

```bash
export OTEL_RESOURCE_ATTRIBUTES="$LAMBDA_RESOURCE_ATTRIBUTES,$OTEL_RESOURCE_ATTRIBUTES";
```

python/src/otel/otel_sdk/otel-instrument
NathanielRN: Hm yes that makes sense, I tried to find it in the spec but could not. Yet I know you're correct because [in the default "OTel Distro for Python" we _do_ set `otlp_protoc_grpc_span`](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/286f26bbed53e87ac332fc25b802eb5e06b06f91/opentelemetry-distro/src/opentelemetry/distro/__init__.py#L34) for `OTEL_TRACES_EXPORTER`.

This is the package we install for OTel Python Lambda:

https://github.com/open-telemetry/opentelemetry-lambda/blob/e6e7905b5c905e238745c68fdb9a51308424564c/python/src/otel/otel_sdk/requirements.txt#L2

In this PR we use `opentelemetry-instrument` to "ask this distro" what the exporter should be. However this is almost the same as what we are doing right now where we just do this by explicitly copying the code:

https://github.com/open-telemetry/opentelemetry-lambda/blob/e6e7905b5c905e238745c68fdb9a51308424564c/python/src/otel/otel_sdk/otel_wrapper.py#L21-L37

I will remove it and leave the comment? Let me know if I should delete the comment too 🙂

python/src/otel/otel_sdk/otel-instrument
NathanielRN: Ah thanks for this I was probably editing the wrong version of the `otel-instrument` script 🙂 

Yes looks like [OTEl Python does set the correct default](https://github.com/open-telemetry/opentelemetry-python/blob/10231d127f5bef6d1e9bc88a7cbd455c2e560c0d/opentelemetry-api/src/opentelemetry/propagate/__init__.py#L129-L132).

I will delete this line and leave a comment! Let me know what you think.

python/src/otel/otel_sdk/otel_wrapper.py
NathanielRN: THANK YOU THANK YOU 😄

python/src/otel/otel_sdk/otel_wrapper.py
NathanielRN: Makes sense to me too 🙂

I changed it look like this:

```python
try:
    (mod_name, handler_name) = path.rsplit(".", 1)
except ValueError as e:
    raise HandlerError(
        "Bad path '{}' for ORIG_HANDLER: {}".format(path, str(e))
    )
```

Let me know if this is what you had in mind?

python/src/otel/tests/test_otel.py
```diff
@@ -77,8 +90,26 @@ def mock_aws_lambda_exec_wrapper():
...
+            "pass",
+        ]
+    )
+
+    # NOTE: This should have been done by `otel-instrument`... but because there
+    # is no way to run this bash script so that it can affect this python
+    # environment, we have to COPY and PASTE its duplicate here ourselves.
+
+    os.environ[ORIG_HANDLER] = os.environ[_HANDLER]
+    os.environ[_HANDLER] = "otel_wrapper.lambda_handler"
```
NathanielRN: This is one downside of converting the script to a `bash` script. When it was a python script we could use `exec` and call it in the same process. But it fails with a `SyntaxError` when we try to `exec` this bash script.

anuraaga: You could make this a bit more robust by having the `subprocess` e.g. print out its `os.environ` instead of just doing `pass`. So here you could parse that stdout and copy in the actual values that were set by that script instead of reproducing the logic here.

anuraaga (approved):

python/src/otel/tests/test_otel.py
NathanielRN: That's a great idea! 😄 I made the update and I think it's much better, thanks again 🙂

python/src/otel/tests/test_otel.py
```diff
@@ -77,8 +89,40 @@ def mock_aws_lambda_exec_wrapper():
...
     """
-    # NOTE: AwsLambdaInstrumentor().instrument() is done at this point
-    exec(open(os.path.join(INSTRUMENTATION_SRC_DIR, "otel-instrument")).read())
+
+    # NOTE: Because we run as a subprocess, the python packages are NOT patched
+    # with instrumentation. In this test we just make sure we can complete auto
+    # instrumentation without error and the correct environment variabels are
+    # set. A future improvement might have us run `opentelemetry-instrument` in
+    # this process to imitate `otel-instrument`, but our lambda handler does not
+    # call other instrumented libraries so we have no use for it for now.
```
NathanielRN, line 98: Just want to flag this comment, I tested it and the subprocess **does not** patch the libraries in the parent process as I initially assumed 😞 I think we should merge as is, and the only suggestion I can think of is to call `opentelemetry-instrument` ourselves in this process if we really want the packages to be instrumented.

Either way the `check=True` below will help us make sure we can run both `otel-instrument` and `opentelemetry-instrument` through to completion without any errors.

wangzlei (approved):
LGTM!

python/src/otel/tests/test_otel.py
NathanielRN: What's weird about this is that the `TracerProvider` must be getting initialized by the child `opentelemetry-instrument` call, but when I make a call to `botocore` for example I should be seeing more spans and yet I do not...

NathanielRN: Actually I understand why now. We use the `from opentelemetry.test.test_base import TestBase` which initializes the `TracerProvider` at the `setupClass` stage.

We can fix this in another PR. See Issue #168.

python/src/otel/otel_sdk/otel-instrument
NathanielRN: Thanks again for the prompt!

I looked into this and am fairly certain OTel Python SDK currently [_does_ handle empty Resource Attributes](https://github.com/open-telemetry/opentelemetry-python/blob/05e9fe171fdfd6a4d7d2336f6ac41bab69c8a71f/opentelemetry-api/src/opentelemetry/attributes/__init__.py#L50-L52).

```python
if not (key and isinstance(key, str)):
    _logger.warning("invalid key `%s`. must be non-empty string.", key)
    return None
```

They even [have a test to protect against this](https://github.com/open-telemetry/opentelemetry-python/blob/05e9fe171fdfd6a4d7d2336f6ac41bab69c8a71f/opentelemetry-sdk/tests/resources/test_resources.py#L212-L229):

```python
def test_invalid_resource_attribute_values(self):
    resource = resources.Resource(
        {
            resources.SERVICE_NAME: "test",
            "non-primitive-data-type": {},
            "invalid-byte-type-attribute": b"\xd8\xe1\xb7\xeb\xa8\xe5 \xd2\xb7\xe1",
            "": "empty-key-value",
            None: "null-key-value",
            "another-non-primitive": uuid.uuid4(),
        }
    )
    self.assertEqual(
        resource.attributes,
        {
            resources.SERVICE_NAME: "test",
        },
    )
    self.assertEqual(len(resource.attributes), 1)
```

So I'll simplify this even further!

NathanielRN: I actually was wrong about this... I've made a PR to fix this upstream: https://github.com/open-telemetry/opentelemetry-python/pull/2256

And will follow up with a fix https://github.com/open-telemetry/opentelemetry-lambda/pull/173

PR #165, opened 2021-10-25
Only test Auto Instrumentation related Python Tests

### LINKED ISSUE #168: [Python] Tests that init OTel with otel-instrument script should not use TestBase setup method.
# Description

As is common with all instrumentations upstream, AWS Lambda instrumentation testing involves using `TestBase`. `TestBase` has a default implementation of `super().setUpClass()` which **initializes OTel Python with a TracerProvider**. 

This is a problem because we want the tests to test the `otel-instrument` command which ALSO sets the `TracerProvider`. But setting the `TracerProvider` a 2nd time is **not allowed** so the effects of `otel-instrument` are ignored. What's more, `otel-instrument` runs as a subshell as of #164 so we'll have to be more clever as to how we propagate its changes to the parent process while also making sure we do NOT prematurely init OTel's `TracerProvider` with the `TestBase` default methods.

One solution is to call the upstream `opentelemetry-instrumentation` auto initialization script ourselves in the test.

NathanielRN:
# Description

Follow up to #164 and https://github.com/open-telemetry/opentelemetry-python-contrib/pull/777

Now that upstream has the manual tests, the OpenTelemetry Lambda repo can focus on just testing the auto instrumentation part of AWS Lambda Instrumentation.

Fixes #168

### DIFF
--- a/python/src/otel/tests/mock_user_lambda.py
+++ b/python/src/otel/tests/mock_user_lambda.py
@@ -1,2 +1,0 @@
-def handler(event, context):
-    return "200 ok"
--- a/python/src/otel/tests/mocks/lambda_function.py
+++ b/python/src/otel/tests/mocks/lambda_function.py
@@ -1,0 +1,17 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


NathanielRN:
Replaced by #177

PR #167, opened 2021-10-27
Lambda Layer Update - OTel Python 1.6.2 and 0.25b2

NathanielRN:
# Description

Updates the Python Lambda Layer to use the latest versions of OTel Python: 1.6.2 for Core and 0.25b2 for Contrib.


NathanielRN:
One confusing thing is that `from opentelemetry.test.test_base import TestBase` will initialize a `TracerProvider` and `otel-instrument` will ALSO try to initialize a `TracerProvider`. Only the 1st one will be used. We can fix this in another PR. See Issue #168.

wangzlei (approved):
LGTM

PR #169, opened 2021-10-28
Lambda Layer Update - OTel JS 1.0.3 and 0.26.0

NathanielRN:
# Description

Updates the NodeJS Lambda Layer to use the latest versions of OTel JS
* 1.0.3 for `@opentelemetry/api`
* 1.0.0 for `@opentelemetry/sdk-trace-node`, `@opentelemetry/resources`, `@opentelemetry/sdk-trace-base`
* 1.0.0 for `@opentelemetry/propagator-aws-xray` and 
* 0.27.0 for `@opentelemetry/instrumentation-aws-lambda`
* 0.26.0 for most instrumentation packages
* 0.26.0 for the `@opentelemetry/exporter-otlp-proto` package
* 0.1.0 for @opentelemetry/instrumentation-aws-sdk


nodejs/packages/layer/package.json
```diff
@@ -26,28 +26,28 @@
...
+    "@opentelemetry/instrumentation-mongodb": "^0.26.0",
+    "@opentelemetry/instrumentation-mysql": "^0.26.0",
+    "@opentelemetry/instrumentation-net": "^0.26.0",
+    "@opentelemetry/instrumentation-pg": "^0.26.0",
+    "@opentelemetry/instrumentation-redis": "^0.26.0",
+    "@opentelemetry/propagator-aws-xray": "~1.0.0",
+    "@opentelemetry/resource-detector-aws": "~1.0.0",
+    "@opentelemetry/resources": "~1.0.0",
+    "@opentelemetry/sdk-trace-base": "~1.0.0",
+    "@opentelemetry/sdk-trace-node": "~1.0.0"
```
willarmiros: You can just use the `^` instead of the tilde, no one really uses it for Node that I've seen since it offers no more protection over breaking changes.

willarmiros (approved):

nodejs/packages/layer/package.json
NathanielRN: Thanks for that! Yeah I had never seen the `~` before either until I [read this article online](https://michaelsoolee.com/npm-package-tilde-caret/).

I learned that:
* `^` will download the latest **minor version** and latest patch version, **unless the package is pre-1.0.0**
* `~` will only download the latest **patch version**

When we were in `^0.25.0` it would only take _patch updates_, but now that packages have reached `1.0.0`, we are okay with taking minor version updates, for example, `1.1.0`.

Like you said, this "offers no more protection over breaking changes" so I agree that `^` makes sense!

I'll change it to use that 🙂

wangzlei (approved):

PR #172, opened 2021-10-30
feat: updated collector to v0.38.0

willarmiros:
As title, and also fixed breaking changes from this new version.


collector/collector.go
```diff
@@ -74,15 +75,15 @@ func (c *Collector) Start() error {
...
+		ConfigMapProvider: c.mapProvider,
 		Factories:      c.factories,
 	}
 	var err error
 	c.svc, err = service.New(params)
 	if err != nil {
 		return err
 	}
-	cmd := service.NewCommand(c.svc)
+	cmd := service.NewCommand(params)
```
willarmiros: @Aneurysm9 Could you please review this? Specifically I'd like to understand what to do with:

```
c.svc, err = service.New(params)
```

After this latest release. I think we can remove it, but it's used a couple times in this file.

jefchien: `service.NewCommand` no longer takes the service instance and instead uses the settings provided to create the  instance as part of `RunE`. https://github.com/open-telemetry/opentelemetry-collector/blob/release/v0.38.x/service/command.go#L34.

If you're using `service.NewCommand` as is, you can no longer use `service.New` to create the instance. It looks like you're using the service instance quite a bit, so you may need to create your own command instead of using `service.NewCommand`.

jefchien: Had to do something similar in https://github.com/aws-observability/aws-otel-collector/blob/main/cmd/awscollector/main.go#L108 to limit the flags.

willarmiros: I see, so we can't access the service instance created by the latest `NewCommand` function?

willarmiros: And assuming that's the case, do you think there's any reason we couldn't just copy the aws-otel-collector's `newCommand` implementation and use it here?

Aneurysm9: I'd go the other direction.  Remove the use of `service.NewCommand()` since it doesn't do anything not already done here except set up some deprecated CLI flags.

collector/collector.go
```diff
@@ -74,27 +74,14 @@ func (c *Collector) Start() error {
...
 		},
-		ParserProvider: c.parserProvider,
+		ConfigMapProvider: c.mapProvider,
 		Factories:      c.factories,
 	}
 	var err error
 	c.svc, err = service.New(params)
 	if err != nil {
 		return err
 	}
```
Aneurysm9, line 77: `Start()` should now take a context and invoke `go c.svc.Run(ctx)`.

willarmiros: Done.

collector/collector.go
```diff
@@ -62,35 +63,29 @@ func NewCollector(factories component.Factories) *Collector {
...
-	cmd := service.NewCommand(c.svc)
-	if args, ok := os.LookupEnv("OPENTELEMETRY_COLLECTOR_ARGS"); ok {
-		cmd.SetArgs(strings.Split(args, " "))
-	}
```
willarmiros, line 88: @Aneurysm9 do we need any sort of replacement for reading this env var?

collector/collector.go
```diff
@@ -62,35 +63,29 @@ func NewCollector(factories component.Factories) *Collector {
 	var r io.Reader = f
 	col := &Collector{
 		factories:      factories,
-		parserProvider: parserprovider.NewInMemory(r),
+		mapProvider: parserprovider.NewInMemoryMapProvider(r),
```
willarmiros: So should this be `ExpandedMapProvider` instead?

Aneurysm9:
```suggestion
		mapProvider: parserprovider.NewExpandMapProvider(parserprovider.NewFileMapProvider(getConfig())),
```

This also allows removing the first part of this function that reads the config into memory since it uses a `FileMapProvider`.

collector/collector.go
Aneurysm9: I don't think so.  In this configuration the collector doesn't have any CLI flags to process.  It takes the location of the config file from the environment via `OPENTELEMETRY_COLLECTOR_CONFIG_FILE` and all other configuration happens there.

collector/collector.go
willarmiros: Got it, done.

collector/collector.go
willarmiros: Ahh I see. I confirmed in our docs for ADOT Lambda we only recommend using the `OPENTELEMETRY_COLLECTOR_CONFIG_FILE` env var for config.

collector/collector.go
```diff
@@ -54,43 +53,31 @@ func getConfig() string {
...
 		return err
 	}
-	cmd := service.NewCommand(c.svc)
-	if args, ok := os.LookupEnv("OPENTELEMETRY_COLLECTOR_ARGS"); ok {
-		cmd.SetArgs(strings.Split(args, " "))
-	}
 
-	c.appDone = make(chan struct{})
 	go func() {
-		defer close(c.appDone)
```
Aneurysm9, line 92: I think this might still be needed.  It is read from in `Stop()`.

willarmiros: Good catch

Aneurysm9 (approved):

wangzlei (approved):

PR #177, opened 2021-11-15
Lambda Layer Update - OTel Python 1.7.1 & 0.26b1

NathanielRN:
# Description

Because OTel Python has [released version 1.7.1](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.7.1) at https://github.com/open-telemetry/opentelemetry-python/commit/281c97bf8f9e31392859e006e13c9c8eac8967c3, we are able to use the upstreamed code for the AWS Lambda Instrumentation instead of having a copy of it here in this repo.

We also included some small nit updates in passing to the build files + the CI files.

Built the layer locally and tested on Lambda, confirmed that OTel Python is still initialized correctly and traces appear.


python/src/otel/tests/test_otel.py
```diff
@@ -241,77 +261,16 @@ def test_parent_context_from_lambda_event(self):
...
-        self.assertEqual(len(parent_context.trace_state), 3)
-        self.assertEqual(
-            parent_context.trace_state.get(MOCK_W3C_TRACE_STATE_KEY),
-            MOCK_W3C_TRACE_STATE_VALUE,
-        )
-        self.assertTrue(parent_context.is_remote)
-
-        test_env_patch.stop()
-
-    def test_using_custom_extractor(self):
```
NathanielRN, line 280: This test [is already upstream](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/a9507b3d1daf79b412b67fd97ddb4e2b7117ebcd/instrumentation/opentelemetry-instrumentation-aws-lambda/tests/test_aws_lambda_instrumentation_manual.py#L197) so we don't need to duplicate it here.

wangzlei (approved):

PR #182, opened 2021-11-24
Use Go 1.17 because Collector cannot be built without it

NathanielRN:
## Description

Build collector before we use SAM Docker Container.
* The recent v0.39.0 collector introduced a change that required go 1.17
* We can't download go 1.17 in the Docker Container used by SAM
* We build the collector outside the Docker Container and just copy over later


wangzlei (approved):
LGTM!

PR #186, opened 2021-11-27
Combine sample app integration test terraform into common directory

NathanielRN:
## Description

This PR attempts to combine the terraform from all the support languages into one common terraform directory with more variables. This should make it easier to modify the terraform, at the cost of supplying more variables when calling the terraform.


NathanielRN:
Spoke with @anuraaga, we decided against this change. We want to make it easy for users to find this repo and build their own layers, and a common directory does not help with that.

Instead, since the files are so similar, we will use a codegen tool (like `jinja2` in python) to generate the files from a template.

PR #188, opened 2021-11-29
Support architectures across all sample apps

NathanielRN:
## Description

This PR adds `architectures` as an input variable for every sample app created using terraform.


NathanielRN:
Because [the main-build workflow on my repo in the aws-otel-lambda downstream](https://github.com/NathanielRN/aws-otel-lambda/actions/runs/1528365651) _mostly_ works, this PR is ready to merge!

The reason we say _mostly_ is because of an issue we are tracking internally where X-Ray spans appear out of order on the console **at inconsistent times** for the .NET, NodeJS, Python, and Go Lambda Functions. (Java has never failed in this regard).

wangzlei (approved):

PR #194, opened 2021-12-02
 Fix scoping bug for errors on collector service Run

phenixrizen:
I had a tab error in my config yaml and the extension layer collector was hanging with no error output which led me to this bug.

The current logic will never receive the error returned from `appErr := c.svc.Run(ctx)` because error will be out of scope and the collector will hang. This creates a channel and assures we check the error. This allows us to get the config error.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -78,11 +78,11 @@ func (c *Collector) Start(ctx context.Context) error {
 
 	c.appDone = make(chan struct{})
 	runErr := make(chan error, 1)
-	go func(runErr chan error) {
+	go func() {
 		defer close(c.appDone)
-		err := c.svc.Run(context.Background())
+		err := c.svc.Run(ctx)
 		runErr <- err
-	}(runErr)
+	}()
 
 	rErr := <-runErr
 	if rErr != nil {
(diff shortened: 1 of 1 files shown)


collector/collector.go
```diff
@@ -77,13 +77,18 @@ func (c *Collector) Start(ctx context.Context) error {
...
-	go func() {
+	runErr := make(chan error, 1)
+	go func(runErr chan error) {
 		defer close(c.appDone)
-		appErr := c.svc.Run(ctx)
-		if appErr != nil {
-			err = appErr
-		}
-	}()
+		err := c.svc.Run(context.Background())
```
NathanielRN: Just curious, why is `context.Background()` better than `ctx` here?

NathanielRN (approved):
Thanks for this change! It looks useful 🙂

collector/collector.go
phenixrizen: Good catch that should be the passed context. I was working off a older version that had no context passed to Run.

phenixrizen:
Any update on getting this merged would like to use this instead of a fork.

NathanielRN:
It's hard with people on holidays I guess, but I also would like to see this merged because I have ran into it too 😞

@codeboten think you could merge this when you have a chance? 🙂

codeboten (approved):
Thanks for fixing!

PR #195, opened 2021-12-07
Fix spacing of ambiguous list

AjohnsonLS:
in Section 3.3, a list is presented on a single line with no distinguishing punctuation. New lines were not being correctly rendered. This fix makes the meaning clear.

### DIFF
--- a/docs/design_proposal.md
+++ b/docs/design_proposal.md
@@ -59,8 +59,8 @@ In previous technical challenges we mentioned two things: Memory consumption and
 As a prototype, Python is implemented in [current Repo](https://github.com/open-telemetry/opentelemetry-lambda/tree/main/python).
 
 We don’t provide SDK public layers for every language. This is required in the following two cases:
-Languages with auto-instrumentation support, such as Python and Java
-Scripting languages such as Python, Node.js, and Ruby
+Languages with auto-instrumentation support, such as Python and Java, and 
+Scripting languages such as Python, Node.js, and Ruby.
 At the moment only Python and Java support auto-instrumentation. Other languages such as Go, .Net support library instrumentation only. For these languages, users must explicitly add OpenTelemetry SDK dependencies and instrumentation code, then re-compile the Lambda application to enable library instrumentation. In such cases we would not provide OpenTelemetry SDK layer because it does not help in compilation and deployment.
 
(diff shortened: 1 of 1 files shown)


docs/design_proposal.md
```diff
@@ -59,8 +59,8 @@ In previous technical challenges we mentioned two things: Memory consumption and
 As a prototype, Python is implemented in [current Repo](https://github.com/open-telemetry/opentelemetry-lambda/tree/main/python).
 
 We don’t provide SDK public layers for every language. This is required in the following two cases:
-Languages with auto-instrumentation support, such as Python and Java
-Scripting languages such as Python, Node.js, and Ruby
+Languages with auto-instrumentation support, such as Python and Java, and 
+Scripting languages such as Python, Node.js, and Ruby.
```
NathanielRN, line 63: Maybe we could do this as a bulleted list?

```suggestion
- Languages with auto-instrumentation support, such as Python and Java
- Scripting languages such as Python, Node.js, and Ruby
```

docs/design_proposal.md
```diff
@@ -69,4 +69,4 @@ Even if we don't provide an SDK layer for a language, we still need to provide a
 
 ## 4. CI/CD convention
 Each sample application and layer source folder must have a one-click script `run.sh`, to build and deploy OpenTelemetry Lambda application in a uniform manner. It is not only providing a good user experience but also for building a generic CI/CD workflow, github action can simply run tests and publish Lambda layer no matter what the stack tool(Terraform, CloudFormation), or language(Java, Python) the sample application is. 
-See an example for publishing [Python layer](../python/src/).
\ No newline at end of file
+See an example for publishing [Python layer](../python/src/).
```
NathanielRN, line 72: What is the change on this line? I can't tell from the GitHub diff 😓

AjohnsonLS: I didn't intend to make a change here. I'm uncertain what it is picking up

docs/design_proposal.md
AjohnsonLS: No objection to that.

docs/design_proposal.md
NathanielRN: Hm you probably want to do something like `git checkout origin/main -- docs/design_proposal.md` to rest the changes, and then apply your fix above again in a new commit? That might be the only way to stop git from saying there is a change here.

NathanielRN (changes requested):
This looks helpful, thanks! Think you can change the wording format and remove that change? Appreciate it!

PR #204, opened 2022-01-26
Lambda Layer Update - OTel JS 1.0.1

### LINKED ISSUE #183: Node.js Layer using Depreated Exporter
If I'm reading things correctly, the Node.js exporter that's used in the Node.js layer

https://github.com/open-telemetry/opentelemetry-lambda/blob/f6aadfd5932fa7e1e590936be153294645143694/nodejs/packages/layer/src/wrapper.ts#L39

is deprecated.  See https://cloud-native.slack.com/archives/C01NL1GRPQR/p1637779230143700

In addition to being deprecated, it does not appear that the deprecated `@opentelemetry/exporter-otlp-proto` allows users to export using OTLP over GRPC.  It only supports exporting over HTTP using `@opentelemetry/exporter-otlp-http`.

It would be great if this could be updated to use one of currently supported exporters _and_ if data could be exported via either GRPC and HTTP.  

(Not so great that I have time to take on the work, but great regardless ;))

vasireddy99:
This PR updates the OTel JS Core and OTel JS Contrib dependencies.

Fixes #183


NathanielRN (approved):
Nice!

wangzlei (approved):

PR #215, opened 2022-02-07
feat: add extension point for customising lambda instrumentation

chrisrichardsevergreen:
## Which problem is this PR solving?
There is no easy way to customise the ```AwsLambdaInstrumentation```, it's currently created with default arguments when there are in reality a range of things that can be customised on it, such as ```disableAwsContextPropagation``` and the custom event extractor.

This PR adds a global ```configureLambdaInstrumentation``` function which allows downstream code to customise the config before the instrumentation is created

### Short description of the changes
- Add ```configureLambdaInstrumentation``` alongside the other provided ```configure...``` extension points

### DIFF
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -19,6 +19,7 @@ import {
   DiagLogLevel,
 } from "@opentelemetry/api";
 import { getEnv } from '@opentelemetry/core';
+import { AwsLambdaInstrumentationConfig} from '@opentelemetry/instrumentation-aws-lambda';
 
 // Use require statements for instrumentation to avoid having to have transitive dependencies on all the typescript
 // definitions.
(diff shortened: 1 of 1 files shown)


nodejs/packages/layer/src/wrapper.ts
```diff
@@ -19,6 +19,7 @@ import {
   DiagLogLevel,
 } from "@opentelemetry/api";
 import { getEnv } from '@opentelemetry/core';
+import { AwsLambdaInstrumentationConfig} from '@opentelemetry/instrumentation-aws-lambda';
```
NathanielRN:
```suggestion
import { AwsLambdaInstrumentationConfig } from '@opentelemetry/instrumentation-aws-lambda';
```

NathanielRN (approved):
This looks cool! Thanks!

ryansonshine:
How does the scoping on this work?

eternalmatt:
Hey @chrisrichardsevergreen - looks great, but how do I use this?

I have a layer that is working well with the Layer configured, but how can I add a global `configureLambdaInstrumentation` function?

PR #217, opened 2022-02-10
Lambda Layer Update - JS Contrib Feb 2022

chrisrichardsevergreen:
- Update @opentelemetry/instrumentation-aws-lambda to pull in newly added functionality
- Update @opentelemetry/instrumentation-express and @opentelemetry/instrumentation-graphql

PR #222, opened 2022-02-19
Lambda Layer Update - JS Contrib Feb 2022

vasireddy99:
This PR updates the following OTel JS Contrib dependencies.
 
 *`instrumentation-aws-lambda`
 *`instrumentation-express`
 *`instrumentation-graphql`


NathanielRN (approved):
Thanks! Can you update the description and title to uniquely identify these changes a bit more? Maybe like "Lambda Layer Update - JS Contrib Feb 2022" for the title and something similar for the description?

vasireddy99:
> Thanks! Can you update the description and title to uniquely identify these changes a bit more? Maybe like "Lambda Layer Update - JS Contrib Feb 2022" for the title and something similar for the description?

thanks for the suggestion :-), named it since it was just js contrib dependencies. sure, yes renamed it as it is useful.

NathanielRN:
Hey! Sorry I didn't notice this before, but this PR is exactly like #217. Since that one was first, can we close this one in favor of that one?

PR #242, opened 2022-05-19
Lambda Layer Update: OTel Python 1.12.0 & 0.33b0

vasireddy99:
This PR updates the OTel Python Core dependencies to 1.12.0 and OTel Python Contrib dependencies to 0.33b0

Adding detail for PR build;
Executed the command `tox` in local and can verify both `python3.8` ,  `python3.9` are succeeded

image

Resolves :

 * https://github.com/open-telemetry/opentelemetry-lambda/issues/269
`


NathanielRN:
We spoke offline and decided we do not want to jump the gun and release Python Lambda layers while the OTel Python SIG has them in **pre-release**. That is the case for `1.12.0rc1`:

image

@vasireddy99 was able to pass the tests (and avoid the problem where conflicting versions like `1.11.0` are installed), by updating the `CORE_REPO_SHA` environment variable to point to the repo at the pre-release itself. Otherwise, it installs `opentelemetry-test-utils` at the specified commit and that package in turns pulls the latest _stabe_ release of `opentelemetry-api`. Which is `1.11.0` right now.

We shouldn't release pre-releases because they are still being tested for bugs. For example consider this open issue with the latest `1.12.0rc1` release candidate 🙂

https://github.com/open-telemetry/opentelemetry-python/issues/2716

suqi:
Hello @NathanielRN  @vasireddy99 , please help check again, OTel Python has released v1.12.0 officially several days ago.
https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.12.0

And if it cannot get upgrade, now my OTel lambda layer is reporting error
> TracerProvider was missing force_flush method. This is necessary in case of a Lambda freeze and would exist in the OTel SDK implementation.

More detailed error info:

image

Thank you Team, have a good day!

vasireddy99:
This PR will be updated to use `1.12.0` soon.

.github/workflows/pr-python.yml
```diff
@@ -9,14 +9,16 @@ on:
 
 env:
   AWS_REGION: us-east-1
+  #copied the core repo sha from here:https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/.github/workflows/test.yml#L9
+  CORE_REPO_SHA: ece57231cde5bb5656775a62f027cab0b1c76fce
```
Aneurysm9, line 14: Is this still necessary?

.github/workflows/pr-python.yml
```diff
@@ -9,14 +9,16 @@ on:
...
 
 jobs:
   PR-Python:
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
-        python: [3.8, 3.9]
+        python: ['3.8', '3.9']
```
NathanielRN, line 22: Why do we need to do this?

vasireddy99: The usage of quotes is recommended by [setup-python](https://github.com/actions/setup-python#basic-usage) in gh actions. It would cause error when suppose a version `3.10` has to be used.

.github/workflows/pr-python.yml
vasireddy99: this env variable would be necessary to test against the specific commit other than `main` and is in a format the tests to follow the format of upstream

NathanielRN (approved):

PR #248, opened 2022-06-14
Working version for Lambda - AWS Managed Prometheus integration

pcolazurdo:
This PR shows a fully working Lambda function based on Python sending metrics to AWS Managed Prometheus using an OTel Layer.

A few comments:
* As force_flush is not defined yet to wait for the actual collector to flush the metrics we add a fake wait of 1 second at the end of the function call for it to succeed. This should be removed once the SDK/API team solve the discussions around force_flush functionality for FaaS
* The current version of the Python opentelemetry-api module (1.11.1) has issues with the protobuf version which makes the library unusable ([more info](https://github.com/open-telemetry/opentelemetry-python/issues/2717)). To solve this we are asking the Function to use the python implementation of the protobuf compiler on the fly by adding an Environment variable ``` PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION: python``` in the ```template.yaml``` for the function.
* Also, the current version of the opentelemetry-api module (1.11.1) marks the ```metrics``` api as internal, that's why ```_metrics``` is used. Once the next stable version is released, I'll update the code with a new PR.


collector/config.yaml
```diff
@@ -7,6 +7,11 @@ receivers:
 exporters:
   logging:
     loglevel: debug
+  awsprometheusremotewrite:
+    endpoint: "https://aps-workspaces.eu-west-1.amazonaws.com/workspaces/ws-12345678-1234-1234-1234-123456789012/api/v1/remote_write"
+    aws_auth:
+      service: "aps"
+      region: "eu-west-1"
```
NathanielRN: Is this a real endpoint that is used? If it's not, would this cause a problem if downstream repositories are using this file?

python/sample-apps/template.yml
```diff
@@ -12,19 +12,26 @@ Resources:
     Type: AWS::Serverless::Function
     Properties:
       Handler: lambda_function.lambda_handler
-      Runtime: python3.9
+      Runtime: python3.7
```
NathanielRN: Sorry, why did we change this to 3.7?

python/sample-apps/function/lambda_function.py
```diff
@@ -3,6 +3,30 @@
...
+
+from opentelemetry import _metrics
+from opentelemetry.exporter.otlp.proto.grpc._metric_exporter import (
+    OTLPMetricExporter,
+)
+import opentelemetry.exporter.otlp.proto.grpc.version
+
+print( "opentelemetry.exporter.otlp.proto.grpc version", opentelemetry.exporter.otlp.proto.grpc.version.__version__)
+
+from opentelemetry._metrics import (
```
NathanielRN, line 17: Should we use the metrics library? I know they announced GA for it, but from what I've seen OTel Python is still making updates that we could afford to wait to include?

collector/config.yaml
jeromeinsf: I suspect this should be replaced with actual workspace from customer.
I think one addition to the PR would be a readme/documentation about what is added and how to use it

pcolazurdo: This was a sample number to show how to fill the endpoint URL - I will add a comment

python/sample-apps/template.yml
pcolazurdo: Good catch - I was running with an old version of SAM and 3.9 was failing. I'll change that.

python/sample-apps/function/lambda_function.py
pcolazurdo: The whole idea of this sample is to show how to use the metrics library. I know this is in active development, but this allows users to see how the functionality would work. Not many changes are in the library in terms of API & SDK. There is a mention about some things that are broken at the minute in the first comment to give some ideas about things that may change.

collector/config.yaml
Aneurysm9: Rather than changing the default configuration here, perhaps include a configuration with the sample app that includes its use.  Also, note that the `awsprometheusremotewrite` exporter is deprecated and should instead use the `prometheusremotewrite` exporter and `sigv4auth` extension for authentication.

python/sample-apps/function/lambda_function.py
Aneurysm9: I think the question is whether this should use `metrics` or `_metrics`.  The former being the stable release version of the metrics API that should be used going forward.

python/sample-apps/function/lambda_function.py
```diff
@@ -21,8 +45,18 @@ def lambda_handler(event, context):
 
     loop = asyncio.get_event_loop()
     loop.run_until_complete(callAioHttp())
+    
+    counter = meter.create_counter(name="first_counter", description="TODO", unit="1",)
```
Aneurysm9, line 49: This should be defined outside of the handler, right?  I'd expect it to be created once and used with each invocation.

python/sample-apps/template.yml
```diff
@@ -12,19 +12,26 @@ Resources:
...
       CodeUri: ./function
       Description: Build OTel Python Lambda Sample from scratch
       MemorySize: 512
-      Timeout: 15
+      Timeout: 310
```
Aneurysm9: This is a significant change.  Why?

python/sample-apps/function/lambda_function.py
```diff
@@ -21,8 +45,18 @@ def lambda_handler(event, context):
...
     for bucket in s3.buckets.all():
-        print(bucket.name)
-
+        i = i + 1
+        # print(bucket.name)
+        
+    counter.add(i, attributes={"invocation": context.aws_request_id})
+
+    provider.force_flush() # needed to be sure that metrics are sent to OTel before shutting down the worker
+    time.sleep(1)  # This is needed because at the moment there is no API to force the collector to push the message before the Worker is suspended
```
Aneurysm9, line 59: Is `force_flush()` not synchronous?  The collector used in the extension is synchronous (there is no `batch` processor in its pipeline and no WAL in the PRW exporter config) so as long as the exporter here waits for a response from the collector then the data will have been sent out via the collector's exporter.

collector/config.yaml
pcolazurdo: Thank you @Aneurysm9 - I'll update those

python/sample-apps/function/lambda_function.py
pcolazurdo: Will test with ```metrics``` and get back

python/sample-apps/function/lambda_function.py
pcolazurdo: Good catch

python/sample-apps/template.yml
pcolazurdo: This was from an old test - Will fix

python/sample-apps/function/lambda_function.py
pcolazurdo: In the current implementation ```foce_flush() ``` force_flush for metrics doesn't seem to be synchronous (at least not in my tests) - I expect this to change once the stable version is released.

python/sample-apps/function/lambda_function.py
pcolazurdo: I've reviewed this, and it isn't a good idea for an example. As this metric is a counter, if ```create_counter``` is moved outside, the value will be increased across invocations for as long as the execution environment is the same.
I think, for this example, and for most use cases, creating the counter inside the function handler is the right approach

python/sample-apps/function/lambda_function.py
pcolazurdo: Lates Release available [1.11.1](https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.11.1) is still using ```_metrics``` - will update on a new PR once the next stable version is released

NathanielRN:
@pcolazurdo Are you able to fix the unit tests?

I'll defer to @Aneurysm9 for the approval here 🙂

pcolazurdo:
Updated to match latest commit on main. Some comments:
* go.mod and go.sum are different because I had to include ```awsprometheuswriter```. This is because the current collector doesn't seem to include  the ```sigv4auth extension```
* ```config.yaml``` is modified on the ```collector/``` path because there isn't a mechanism defined to override configuration. I would have used ```AOT_CONFIG_CONTENT``` environment variable as discussed [here](https://aws-otel.github.io/docs/setup/ecs/config-through-ssm) but this option doesn't seem to be available in the stripped-down version version of the collector in the project.

erichsueh3:
> Updated to match latest commit on main. Some comments:
> 
> * go.mod and go.sum are different because I had to include `awsprometheuswriter`. This is because the current collector doesn't seem to include  the `sigv4auth extension`
> * `config.yaml` is modified on the `collector/` path because there isn't a mechanism defined to override configuration. I would have used `AOT_CONFIG_CONTENT` environment variable as discussed [here](https://aws-otel.github.io/docs/setup/ecs/config-through-ssm) but this option doesn't seem to be available in the stripped-down version version of the collector in the project.

Hi @pcolazurdo, can you clarify on your first point? The `sigv4auth` extension has been included since 0.48, and the current collector version is latest, at 0.56.

pcolazurdo:
@erichsueh3 I've tried initialising the extension and I just get an extension crash on Lambda. As soon as I add:

```yaml
extensions:
  sigv4auth:
    region: "us-west-2"
...
service:
  extensions: [sigv4auth]
```

I get:
```
2022/08/02 11:41:36 Failed to start the extension: unable to start, otelcol state is 3 
```

Also, looking in the source code in the repo for ```sigv4authextension```, I couldn't find any reference to it, not even in the go.mod.

collector/lambdacomponents/default.go
```diff
@@ -23,6 +23,7 @@ import (
 	"go.opentelemetry.io/collector/receiver/otlpreceiver"
 	"go.uber.org/multierr"
 
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/awsprometheusremotewriteexporter"
```
Aneurysm9, line 26: The `sigv4auth` extension should be added here rather than the `awsprometheusremotewrite` exporter.

pcolazurdo: I didn't find the right way to add it. Looking at the code in the collector directory and the code in https://github.com/aws-observability/aws-otel-collector regarding the extensions - specifically [this](https://github.com/aws-observability/aws-otel-collector/blob/2d4dd74947dc95826001a82e23a51c47a2d7f43b/pkg/defaultcomponents/defaults.go#L73) , they look different (no ExtensionFactoryMap is present) so I didn't know what the strategy to add extensions in this stripped-down version was. Adding ```awsprometheuswriter``` was straightforward and didn't need major modifications in the existing collector. It looks like we I could to sync the code in https://github.com/aws-observability/aws-otel-collector/tree/2d4dd74947dc95826001a82e23a51c47a2d7f43b/pkg/lambdacomponents with the code here? I think this should be a totally different PR, am I right?

erichsueh3: You can add the ExtensionFactoryMap in this PR, and add the `sigv4` extension instead of the `awsprometheusremotewrite` exporter, it doesn't have to be in a separate PR

python/sample-apps/template.yml
```diff
@@ -16,15 +16,22 @@ Resources:
...
+        - Statement:
+          - Sid: AMPRemoteWrite
+            Effect: Allow
+            Action:
+            - aps:RemoteWrite
+            Resource: '*'
       Environment:
         Variables:
           AWS_LAMBDA_EXEC_WRAPPER: /opt/python/otel-instrument
+          PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION: python  # needed because https://github.com/open-telemetry/opentelemetry-python/issues/2717 / when new release of library is released this should be solved and can be removed
```
vasireddy99, line 34: This issue is now resolved as the `opentelemetry-python` has pinned the protobuf version.

vasireddy99:
This example in the PR is really helpful. 
* `prometheusremotewriteexporter` with `sigv4` auth extension can now be used as [PR](https://github.com/open-telemetry/opentelemetry-lambda/pull/278) is merged. 
* I second this [comment](https://github.com/open-telemetry/opentelemetry-lambda/pull/248#discussion_r904355529) as it would be the be better case [example](https://github.com/aws-observability/aws-otel-lambda/blob/main/java/sample-apps/aws-sdk/deploy/agent/main.tf) as in the downstream repo used for integration tests.
* force_flush() is now supported for [metrics](https://github.com/open-telemetry/opentelemetry-python/pull/2852) in `v1.12.0`, although this [issue](https://github.com/open-telemetry/opentelemetry-lambda/issues/272) still exists in [opentelemetry-instrumentation-aws-lambda](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-aws-lambda).

pcolazurdo:
Lot has changed in the project since I built this demo and now. It served its purpose but it is now mostly obsolete. I'll close this for now to avoid people getting confused, and try to create a new one in the coming weeks (no promise)

PR #207, opened 2022-01-29
Wait on state change to start + Catch config err

NathanielRN:
## Description

Follow up to #205

This PR fixes 2 issues that our integration tests discovered:
1. We must wait for the Collector to change to `Running` State before finishing returning from the Start.
2. The code to catch bad config files works for bad configs, but blocks the code for good configs and doesn't allow the collector to start

PR #255, opened 2022-07-02
Fix race on Collector.Start() loosing error info

iximiuz:
When the inner collector (go.opentelemetry.io/collector/service) Run() call fails
(likely due to a bad config file) the lambda collector wrapper would
likely loose the actual error information and report just the unexpected
collector state (CLOSED) back to the caller. This PR makes the collector wrapper's
Start() method race-free, so that the actual error information is not
lost.

Another improvement is related to the main() procedure - before this PR the main loop
wouldn't handle situations when the inner collector exits before the
SHUTDOWN event.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"time"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config"
@@ -43,7 +44,8 @@ type Collector struct {
 	factories      component.Factories
 	configProvider service.ConfigProvider
 	svc            *service.Collector
-	appDone        chan struct{}
+	svcDone        chan struct{}
+	svcErr         error
 	stopped        bool
 }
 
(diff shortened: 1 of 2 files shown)


pomidoroshev (approved):

NathanielRN:
@Aneurysm9 could you please help take a look at this? I remember last time we edited these files in #194 we were burned pretty bad before a release until we fixed in #207 😓 Please let me know what you think of this change!

iximiuz:
Kind reminder that this PR is still awaiting its review 😇

Aneurysm9:
I think this PR will not cause the same issues that #194 did.  It does not hang the initialization until the end of `svc.Run()` like that one did.  I will try to run this through some tests today, but the approach and code seem sound to me.

NathanielRN:
Thank you @Aneurysm9! I defer to you to merge if you think it looks good 🙂

tylerbenson:
Is this still a problem? This is a pretty old PR.  I'm inclined to close it unless you want to revisit/rebase on main.

iximiuz:
@tylerbenson feel free to close it. I won't have time to check if the issue remains, and we don't use this code in prod now.

PR #262, opened 2022-07-23
docs: update example to match Go naming conventions

a-h:
* Package names in Go should be lowercase (https://go.dev/blog/package-names)
* Variable and function names should use mixed caps instead of snake case (https://go.dev/doc/effective_go#mixed-caps)


codeboten (commented):
Thanks for the contribution @a-h! Please sign the CLA and we can get this reviewed/merged

Aneurysm9 (approved):

Issue #230, opened 2022-03-30
Logging from the lambda extension should have a config to disable it


lephuongbg:
Currently logging in the collector lambda extension is not configurable (using go's `log.Println` directly), which is drowning out most of our lambda logs. Easiest way to do IMO is simply read a env variable (e.g `OTEL_LAMBDA_EXTENSION_LOG_QUIET`) then skip all log lines that is not for printing out error. Or alternatively, be quiet by default, then read `OTEL_LAMBDA_EXTENSION_LOG_VERBOSE` to enable these logs.

https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/main.go

For context, example of the lambda extension logs:

```
2022/03/30 07:06:21 [collector] Launching OpenTelemetry Lambda extension, version:  v0.1.0
--
2022/03/30 07:06:21 Using config file at path /var/task/collector.yaml
2022/03/30 07:06:21 Registered extension ID: "0bc52143-9d3f-493f-a186-ab4d5c3242ff"
2022/03/30 07:06:21 [collector] Register response: {
"functionName": "xxxxxxxxxxxxxxxxxxxxxxxxx",
"functionVersion": "$LATEST",
"handler": "api_handler.handler"
}
2022/03/30 07:06:21 [collector] Waiting for event...
[WARNING]	2022-03-30T07:06:23.078Z		Could not convert OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT value  to int
START RequestId: 204a0995-9fca-4ce6-9486-a81ef62281d7 Version: $LATEST
EXTENSION	Name: collector	State: Ready	Events: [INVOKE,SHUTDOWN]
2022/03/30 07:06:28 [collector] Received event: {
"eventType": "INVOKE",
"deadlineMs": 1648624017646,
"requestId": "204a0995-9fca-4ce6-9486-a81ef62281d7",
"invokedFunctionArn": "arn:aws:lambda:ap-northeast-1:520076009725:function:xxxxxxxxxxxxxxxxx",
"tracing": {
"type": "X-Amzn-Trace-Id",
"value": "Root=1-6244016c-32b2a684443120b1408e9200;Parent=405e5cb7181a127b;Sampled=1"
}
}
2022/03/30 07:06:28 [collector] Waiting for event...
```

NathanielRN:
Hey @lephuongbg, thanks for posting this issue!

[...]

PR #273, opened 2022-08-10
make log level configurable, cleanup logging

nevins-b:
This tries to address #230 by exposing a new setting `OPENTELEMETRY_EXTENSION_LOG_LEVEL` which allows the log level to be configured.


nevins-b:
CLA should now be signed, not sure how to kick off a retest

collector/main.go
```diff
@@ -26,45 +24,64 @@ import (
...
-	logger.Debug("Register ", zap.String("response :", prettyPrint(res)))
+	logger.Debugw("Register", "response", res)
 	// Will block until shutdown event is received or cancelled via the context.
 	processEvents(ctx, collector)
 }
 
+func configureLogger() {
+	atom := zap.NewAtomicLevel()
+
+	level, err := zapcore.ParseLevel(os.Getenv("OPENTELEMETRY_COLLECTOR_LOG_LEVEL"))
```
klagupl: Maybe OPENTELEMETRY_EXTENSION_LOG_LEVEL? these are not logs from collector, but from extension. Collector log level is set in yml config.

klagupl:
any reason why its not merge yet?

NathanielRN:
Could @Aneurysm9 or @vasireddy99 take a look and approve?

vasireddy99:
It will be useful to have the usage of env variable documented as part of the PR.

collector/main.go
```diff
@@ -26,45 +24,64 @@ import (
...
 	"github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 var (
 	extensionName   = filepath.Base(os.Args[0]) // extension name has to match the filename
 	extensionClient = extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
-	logger          = zap.NewExample()
+	logger          *zap.SugaredLogger
```
vasireddy99, line 33: this looks like a better idea too. here : https://pkg.go.dev/go.uber.org/zap#hdr-Choosing_a_Logger
LGTM !

vasireddy99 (approved):

collector/main.go
```diff
@@ -26,45 +24,64 @@ import (
...
-	logger.Debug("Register ", zap.String("response :", prettyPrint(res)))
+	logger.Debugw("Register", "response", res)
 	// Will block until shutdown event is received or cancelled via the context.
 	processEvents(ctx, collector)
 }
 
+func configureLogger() {
+	atom := zap.NewAtomicLevel()
+
+	level, err := zapcore.ParseLevel(os.Getenv("OPENTELEMETRY_EXTENSION_LOG_LEVEL"))
```
vasireddy99, line 72: Thank you so much. Can you add in PR description to indicate what keyword should be assigned to 
`OPENTELEMETRY_EXTENSION_LOG_LEVEL` in order to set log level to debug. So that it could be improved in a next PR's to only accept certain keyword for eg. Debug. Right now it sets to debug level whenever it get the env variable OPENTELEMETRY_EXTENSION_LOG_LEVEL.

vasireddy99:
Adding PR desc: Test results in this comment shows it doesn't provide extension logs when the variable is not set.
Further chore work could be done in next PR's. 
Screen Shot 2022-09-14 at 10 29 00 AM

collector/main.go
klagupl: From what I see, it accepts only a specific keywoard for loglevel and if there is none or wrong it sets it as debug.
Look at what zapcore.ParseLevel does.

klagupl: Some clarification to my previous comment: if os.getEnv returns an empty string zapcore.ParseLevel will parse it as info.
`func (l *Level) unmarshalText(text []byte) bool {
	switch string(text) {
	case "debug", "DEBUG":
		*l = DebugLevel
	case "info", "INFO", "": // make the zero value useful
		*l = InfoLevel
`
Thats why you dont see any logs when there's no env var provided.

collector/main.go
```diff
@@ -26,45 +24,64 @@ import (
...
+	if err != nil {
+		level = zap.DebugLevel
```
klagupl: Maybe info would be more suitable as default? 
If you dont provide the env variable its setting it as info, so I think it should be the same here.

collector/main.go
nevins-b: ah, I didn't realize that `""` check existed, good to know.

collector/main.go
nevins-b: no preference here, I wanted to try to keep it consistent with how it was logging before this change

klagupl (approved):

collector/main.go
vasireddy99: Can you please add PR description about the usage of `OPENTELEMETRY_EXTENSION_LOG_LEVEL` like if env variable is not set then the log level will be set to `info`.  Also, the CLA authorization can be done from [here](https://contributor.easycla.lfx.linuxfoundation.org/#/cla/project/d8cead54-92b7-48c5-a2c8-b1e295e8f7f1/user/474001e0-2aaa-4ed7-92a4-416c8322a693?redirect=https:%2F%2Fgithub.com%2Fopen-telemetry%2Fopentelemetry-lambda%2Fpull%2F273)
@nevins-b

vasireddy99 (approved):

goyatparmod:
You guys still working  on this one.It will help if this get merge.

nerochiaro:
For what is worth, I have tested this branch and it seem to do what it says. Hopefully it will be merged soon, as it's really useful not to be swamped in log messages.

One note would be to document somewhere how to pass the log levels. I had to dig into the zap library source code to know it's "warn" or "WARN" and not "Warning", or "warning"...

See https://github.com/uber-go/zap/blob/ed5598a9e26d75aae01f4846ce31b3fe785c57c7/zapcore/level.go#L180

monken:
How can we get more eyes on this so that this can get merged?

Aneurysm9:
This has been superseded via #316.

PR #276, opened 2022-08-29
Lambda Layer Update- Update OTel JS to 1.6

Kausik-A:
This PR updates OTel JS to `1.6`, OTel API to `1.2.0` and other sdk dependencies to the latest available.


nodejs/packages/layer/package.json
```diff
@@ -26,29 +26,29 @@
     "node": ">=10.0.0"
```
Aneurysm9: Should this be `>=12.0.0`?  v10 is not supported by lambda.

vasireddy99: +1

Kausik-A: updated

nodejs/packages/layer/package.json
```diff
@@ -23,32 +23,32 @@
   "author": "OpenTelemetry Authors",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=12.0.0"
   },
   "dependencies": {
-    "@opentelemetry/api": "^1.1.0",
+    "@opentelemetry/api": "^1.2.0",
```
vasireddy99, line 29: The latest tag is still pointed to 1.1.0. we should double check to use the `next` tag, maybe better to use the version that has `latest` tag.
image

vasireddy99:
I think we can update all the dependencies to the latest available. 
eg. 

```
    "@opentelemetry/exporter-trace-otlp-proto": "^0.32.0",
    "@opentelemetry/exporter-metrics-otlp-proto": "^0.32.0",
```

nodejs/packages/layer/package.json
Kausik-A: The [opentelemetry-js-api](https://github.com/open-telemetry/opentelemetry-js-api/releases) shows `1.2.0` to be the `latest` and doesn't call out anywhere to be RC. Is there a reason to still be using `1.1.0` ?

nodejs/packages/layer/package.json
```diff
@@ -23,32 +23,32 @@
...
     "@opentelemetry/exporter-trace-otlp-proto": "^0.31.0",
     "@opentelemetry/exporter-metrics-otlp-proto": "^0.31.0",
     "@opentelemetry/instrumentation": "^0.31.0",
     "@opentelemetry/instrumentation-aws-lambda": "^0.32.0",
-    "@opentelemetry/instrumentation-aws-sdk": "^0.8.0",
+    "@opentelemetry/instrumentation-aws-sdk": "^0.8.1",
     "@opentelemetry/instrumentation-dns": "^0.29.0",
     "@opentelemetry/instrumentation-express": "^0.30.0",
     "@opentelemetry/instrumentation-graphql": "^0.29.0",
     "@opentelemetry/instrumentation-grpc": "^0.31.0",
```
vasireddy99: "^0.32.0" is available for us to update 🙂

nodejs/packages/layer/package.json
vasireddy99: Yeah, not sure why it is not targeted to latest in `npm` though. 
image

vasireddy99: Looks like it is safe to use the version `1.2.0`. it is compatible and can confirm from here : https://github.com/open-telemetry/opentelemetry-js/pull/3177

nodejs/packages/layer/package.json
Kausik-A: updated :)

vasireddy99 (approved):

wangzlei (approved):

PR #292, opened 2022-09-27
Add X-Ray support

### LINKED ISSUE #291: [Feature Request] X-Ray exporter support in ADOT collector
I'd like to use AWS X-Ray with AWS Lambda Extension layer. It seems that the collector doesn't support AWS X-Ray exporter.
I found that there is configuration for X-Ray exporter (https://github.com/open-telemetry/opentelemetry-lambda/blob/898c38c69cf540979a820fc6b85bf9a3fe35aed2/collector/Makefile#L13). But the exporter itself isn't imported.
I think X-Ray is one of popular options for OTEL exporter in AWS, so it is worth supporting it by default.

garasubo:
Fix #291 

Add X-Ray support in ADOT collector

### DIFF
--- a/collector/go.mod
+++ b/collector/go.mod
@@ -13,6 +13,7 @@ require (
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2 // indirect
 	github.com/antonmedv/expr v1.9.0 // indirect
+	github.com/aws/aws-sdk-go v1.44.96 // indirect
 	github.com/aws/aws-sdk-go-v2 v1.16.14 // indirect
 	github.com/aws/aws-sdk-go-v2/config v1.17.5 // indirect
 	github.com/aws/aws-sdk-go-v2/credentials v1.12.18 // indirect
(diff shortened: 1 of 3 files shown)


Aneurysm9:
@open-telemetry/lambda-extension-maintainers thoughts on this?  I'm not sure whether we should include non-OSS exporters in this distribution.

NathanielRN:
This question always comes up from time to time.

We have concluded that we should not include the X-Ray Exporter. This repo serves as a clean foundation for vendors to build their vendor-specific Lambda Layers and distribute with their offerings. The AWS users should use the https://github.com/aws-observability/aws-otel-lambda repository to advance any development on a Lambda Layer which includes the `aws-xray-exporter`. The `aws-otel-lambda` repo will always have this `opentelemetry-lambda` repository as a submodule so it will get the latest features from OTel Lambda without needing to edit the foundation code with Vendor specific X-Ray packages.

NathanielRN (changes requested):
Please use the https://github.com/aws-observability/aws-otel-lambda repo to have a Lambda Layer with the X-Ray Exporter!

garasubo:
@NathanielRN Oh, I got it. I didn't know aws-oobservability repository since the document in AWS introduced this repository for the manual setup.
Then, it seems better to close it. Thanks for following up.

Issue #269, opened 2022-07-29
Remove protobuf pin when the fix is released in OpenTelemetry-python


vasireddy99:
This Issue is created to follow up for removing the pin of protobuf once the Opentelemetry-python-contrib has a stable release, since the workaround is already merged in main branch. 
reference : https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-grpc/setup.cfg#L53

* https://github.com/open-telemetry/opentelemetry-lambda/pull/268

PR #296, opened 2022-09-28
Lambda Layer Update: OTel Python 1.13.0 & 0.34b0

vasireddy99:
This PR updates the OTel Python Core dependencies to 1.13.0 and OTel Python Contrib dependencies to 0.34b0


python/src/otel/otel_sdk/requirements.txt
```diff
@@ -1,4 +1,3 @@
-opentelemetry-exporter-otlp-proto-http==1.12.0
-opentelemetry-distro==0.33b0
-opentelemetry-instrumentation-aws-lambda==0.33b0
-protobuf==3.20.1
```
bryan-aguilar, line 4: Why was the protobuf dependency removed?

vasireddy99: The protobuf is now pinned in upstream

https://github.com/open-telemetry/opentelemetry-lambda/issues/269

wangzlei (approved):

PR #305, opened 2022-11-01
Ensure collector extension waits until telemetry exported before requesting next event

### LINKED ISSUE #224: ADOT collector sends response to Runtime API instead of waiting for sending traces
Hi!

After a [discussion on Twitter](https://twitter.com/nikovirtala/status/1501530677031976962), I realized that the ADOT collector extension doesn't wait for data to be sent before making the next call to the Lambda Runtime API.

Here's the section of the code that loops on next events: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/main.go#L71

This can cause issues when the destination for the collector is slow and the Lambda function itself finishes in a short period of time. In this situation, the collector could still be sending data while Lambda freezes the execution environment. I've drawn a timing diagram if it helps:

extension_timing drawio

Aneurysm9:
This PR attempts to fix #224 by ensuring the use of a new extension that registers a `SpanProcessor` with the `TracerProvider` used by the Collector's `obsreport` package that can allow the Lambda extension event handling loop to  determine whether there is any activity in any of the collector's pipelines.  It does this by keeping a map of active spans and exposing an ability to determine the number of items in the map.  

The event handling loop will block until a telemetry processing operation begins, or for one second if no spans are observed by the `SpanProcessor`.  This appears to function well as long as each invocation produces some telemetry to the local collector, but may need refinement to avoid negative impact to functions that do not emit telemetry.


Aneurysm9:
This has been superseded by #310.

PR #310, opened 2022-11-16
Use Telemetry API to guide Collector extension lifecycle

### LINKED ISSUE #224: ADOT collector sends response to Runtime API instead of waiting for sending traces

Aneurysm9:


collector/main.go
```diff
@@ -17,20 +17,21 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/open-telemetry/opentelemetry-lambda/collector/internal/extensionAPI"
+	"github.com/open-telemetry/opentelemetry-lambda/collector/internal/telemetryAPI"
```
codeboten: these probably need to be moved down around line 27

collector/collector.go
```diff
@@ -87,7 +87,7 @@ func NewCollector(factories component.Factories) *Collector {
 func (c *Collector) Start(ctx context.Context) error {
 	params := service.CollectorSettings{
 		BuildInfo: component.BuildInfo{
-			Command:     "otelcol",
+			Command:     "otelcol-lambda",
```
codeboten: probably good to do anyway if you wanted to move this out of this PR

Aneurysm9: Yeah, I'll make a separate PR for this.  Didn't even intend to leave it here.

Aneurysm9: #314 fixes this.

collector/main.go
```diff
@@ -97,10 +109,16 @@ func (lm *lifecycleManager) processEvents(ctx context.Context) {
 				return
 			}
 
+			err = lm.listener.Wait(ctx, res.RequestID)
+			if err != nil {
+				lm.logger.Error("problem waiting for platform.runtimeDone event", zap.Error(err), zap.String("requestID", res.RequestID))
```
codeboten: is this necessarily a platform.runtimeDone event here?

collector/main.go
```diff
@@ -17,13 +17,13 @@ package main
...
 	"context"
 	"fmt"
-	"log"
 	"os"
 	"os/signal"
 	"path/filepath"
 	"syscall"
 
-	"github.com/open-telemetry/opentelemetry-lambda/collector/extension"
+	"github.com/open-telemetry/opentelemetry-lambda/collector/internal/extensionAPI"
```
codeboten: nit: package names in go should be lower case

> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps.

collector/main.go
codeboten: i see the listener does the filtering here.

collector/internal/telemetryAPI/listener.go
```diff
@@ -0,0 +1,146 @@
...
+func (s *Listener) Wait(ctx context.Context, reqID string) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			s.logger.Debug("looking for platform.runtimeDone event")
+			items, err := s.queue.Get(10)
+			if err != nil {
+				return fmt.Errorf("unable to get telemtry events from queue: %w", err)
```
bolwie: Nit: typo in telemtry

collector/main.go
```diff
@@ -97,10 +109,16 @@ func (lm *lifecycleManager) processEvents(ctx context.Context) {
 				return
 			}
 
+			err = lm.listener.Wait(ctx, res.RequestID)
```
bolwie: This will block indefinitely for a extensionAPI.Shutdown event, as that event does not carry a RequestID, so this call will wait for requestID "" to arrive, which never happens. (and the process will be killed by the environment) - 
need to move  the check for `res.EventType == extensionAPI.Shutdown` before the call to `lm.listener.Wait`

codeboten (approved):

PR #321, opened 2022-12-05
Revert "Support for node18.x runtime and remove nodejs12.x"

Aneurysm9:
Reverts open-telemetry/opentelemetry-lambda#318.  Support for Node 1.18 does not yet exist in the Terraform Lambda module used here.  See https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/382.


codeboten (approved):
Thanks for catching this @Aneurysm9

PR #346, opened 2022-12-08
[processor/coldstart] add coldstart processor

### LINKED ISSUE #273: faas.coldstart attribute is not being captured by the Lambda layers
Hello team,

I'm doing some testing with AWS Lambda and the OTel Lambda layer. Here is a log generated by the logging exporter.

```
EXTENSION	Name: collector	State: Ready	Events: [INVOKE,SHUTDOWN]
2022/07/28 17:16:42 [collector] Received event: {
	"eventType": "INVOKE",
	"deadlineMs": 1659029202511,
	"requestId": "a4453547-abc6-4f97-9fc8-4bd874ce9d6b",
	"invokedFunctionArn": "arn:aws:lambda:us-east-1:[redacted]:function:otel-sample",
	"tracing": {
		"type": "X-Amzn-Trace-Id",
		"value": "Root=1-62e2c477-06f824e21a7e58fe5ef16c43;Parent=2925122b30b17859;Sampled=1"
	}
}
2022/07/28 17:16:42 [collector] Waiting for event...
Hello, world
2022-07-28T17:16:42.583Z	INFO	loggingexporter/logging_exporter.go:42	TracesExporter	{"#spans": 1}
2022-07-28T17:16:42.583Z	DEBUG	loggingexporter/logging_exporter.go:51	ResourceSpans #0
Resource SchemaURL: 
Resource labels:
     -> telemetry.sdk.language: STRING(python)
     -> telemetry.sdk.name: STRING(opentelemetry)
     -> telemetry.sdk.version: STRING(1.11.1)
     -> cloud.region: STRING(us-east-1)
     -> cloud.provider: STRING(aws)
     -> faas.name: STRING(otel-sample)
     -> faas.version: STRING($LATEST)
     -> service.name: STRING(otel-sample)
     -> telemetry.auto.version: STRING(0.30b1)
ScopeSpans #0
ScopeSpans SchemaURL: 
InstrumentationScope opentelemetry.instrumentation.aws_lambda 0.30b1
Span #0
    Trace ID       : 62e2c47706f824e21a7e58fe5ef16c43
    Parent ID      : c9e8c429e812c43d
    ID             : cb3315ce306fdd4b
    Name           : app.lambda_handler
    Kind           : SPAN_KIND_SERVER
    Start time     : 2022-07-28 17:16:42.512959436 +0000 UTC
    End time       : 2022-07-28 17:16:42.513069571 +0000 UTC
    Status code    : STATUS_CODE_UNSET
    Status message : 
Attributes:
     -> faas.id: STRING(arn:aws:lambda:us-east-1:[redacted]:function:otel-sample)
     -> faas.execution: STRING(a4453547-abc6-4f97-9fc8-4bd874ce9d6b)
```

As you can see, attributes like `faas.name`, `faas.version`, `faas.id`, and `faas.execution` are captured in my traces. However, I don't see an attribute for cold starts. According to this [doc](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/faas.md#incoming-faas-span-attributes), there is an attribute called faas.coldstart and it is a boolean that is true if the serverless function is executed for the first time. However, this attribute is not shown in my traces when using the Otel Lambda layer. I've tested with the Java layer and the Python layer and neither of them were showing that attribute.

codeboten:
Follow up to #345 

This processor associates cold start timing information generated by the `telemetryapireceiver` with the first span processed by the Collector extension containing a `faas.execution` attribute.


collector/processor/coldstartprocessor/processor.go
```diff
@@ -0,0 +1,109 @@
...
+type coldstartProcessor struct {
+	coldstartSpan *ptrace.Span
+	faasExecution *faasExecution
+	logger        *zap.Logger
+	nextConsumer  consumer.Traces
+}
+
+func (p *coldstartProcessor) processTraces(ctx context.Context, td ptrace.Traces) (ptrace.Traces, error) {
+	var errors error
+	td.ResourceSpans().RemoveIf(func(rs ptrace.ResourceSpans) bool {
```
Aneurysm9, line 47: Since this should only ever associate cold start data once per collector instance, would it make sense to add a latch that can be thrown once that work is done and used in a check to short-circuit around this scan of all spans received?

collector/processor/coldstartprocessor/processor_test.go
```diff
@@ -0,0 +1,156 @@
...
+		expectedError error
+	}{
+		{
+			desc:          "no traces",
+			input:         ptrace.NewTraces(),
+			expected:      ptrace.NewTraces(),
+			expectedError: processorhelper.ErrSkipProcessingData,
+		},
+		{
+			desc: "coldstart without executioin",
```
Aneurysm9:
```suggestion
			desc: "coldstart without execution",
```

collector/processor/coldstartprocessor/processor_test.go
```diff
@@ -0,0 +1,156 @@
...
+				td := ptrace.NewTraces()
+				addExecutionSpan(td.ResourceSpans().AppendEmpty().ScopeSpans().AppendEmpty().Spans().AppendEmpty())
+				return td
+			}(),
+		},
+		{
+			desc: "faas.execution and faas.coldstart with coldstart is first",
+			input: func() ptrace.Traces {
+				td := ptrace.NewTraces()
+				span := td.ResourceSpans().AppendEmpty().ScopeSpans().AppendEmpty().Spans().AppendEmpty()
```
Aneurysm9, line 84: What about non-empty resource or scope attributes?  Or if they're different between the execution span and the coldstart span?  I think with these tests the implementation could drop all resource attributes from every span it processes and they'd still pass.

collector/processor/coldstartprocessor/processor.go
codeboten: Good point, there's now a `bool` flag set once the span has been reported

collector/processor/coldstartprocessor/processor_test.go
codeboten: good catch, i added some tests to validate the resource and scope attributes. the current implementation gives preference to the faas execution span's resource+scope details over the coldstart, as it should contain more meaningful data. The coldstart span's timing is the key information.

tsloughter (approved):
After discussing on Slack I understand what is going on here, so I'd suggest at least a top level comment in `processor.go` that gives some level of details. I think specifically it should call out how this would update the resource/scope of all spans from the telemetryreceiver but that as of this time there is only the coldstart span.

codeboten (commented):
Thanks @tsloughter, I've added details in both godoc and the readme based on the conversation we had.

Aneurysm9 (approved):

PR #371, opened 2022-12-14
[chore] make all shell scripts pass shellcheck

johnbley:
Make all shell scripts pass `shellcheck`:
- use `set -e` for all scripts and drop `command || exit` syntax
- Fix all default shellcheck warnings/errors

Happy to also contribute a PR to do shellcheck in CI if folks want.


Aneurysm9 (approved):
Adding shellcheck to the PR build/CI process would be good.

johnbley:
> Adding shellcheck to the PR build/CI process would be good.

I'll start working on a PR for that, then!

codeboten (approved):

PR #399, opened 2023-01-06
Rearrange collector startup to finish extension init sooner

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>


codeboten (approved):
A follow up item here would be good to add tests around the lifecycle manager to prevent regressions

PR #410, opened 2023-01-10
[collector] add tests for lifecycle manager

codeboten:
Improved the error handling in processEvents along the way.

### DIFF
--- a/collector/internal/collector/collector.go
+++ b/collector/internal/collector/collector.go
@@ -33,11 +33,6 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
-type CollectorWrapper interface {
-	Start(ctx context.Context) error
-	Stop() error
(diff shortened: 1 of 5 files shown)


collector/internal/collector/collector.go
```diff
@@ -33,8 +33,13 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
-// Collector implements the OtelcolRunner interfaces running a single otelcol as a go routine within the
-// same process as the test executor.
+type CollectorWrapper interface {
```
Aneurysm9: Should this interface go in the `lifecycle` package since that's where it is used?

codeboten: makes sense, moved.

Aneurysm9 (approved):

PR #414, opened 2023-01-11
[receiver/telemetryapi] refactor code to remove extensionID

codeboten:
This change updates the telemetryapi listener to provide a callback registration mechanism for components. It then calls any registered callback on events it receives. This removes the need for the telemetryapireceiver to run its own listener.


collector/internal/telemetryapi/listener.go
```diff
@@ -114,13 +126,25 @@ func (s *Listener) Shutdown() {
 	}
 }
 
-func (s *Listener) Wait(ctx context.Context, reqID string) error {
+func (s *Listener) handleEvent(ctx context.Context, e Event) error {
+	handlerLock.Lock()
```
mwear: A read lock is probably sufficient here, but it would only be useful if multiple concurrent reads are possible.

mwear (approved):
LGTM

collector/internal/lifecycle/manager.go
```diff
@@ -139,10 +146,21 @@ func (lm *manager) processEvents(ctx context.Context) error {
 				return err
 			}
 
-			err = lm.listener.Wait(ctx, res.RequestID)
-			if err != nil {
-				lm.logger.Error("problem waiting for platform.runtimeDone event", zap.Error(err), zap.String("requestID", res.RequestID))
-			}
+			lm.reqID = res.RequestID
+			lm.wg.Wait()
```
Aneurysm9: This doesn't seem safe, given that this function also calls `defer lm.wg.Done()`.  I'd expect this line to block forever.  This should probably be a separate `WaitGroup` that is incremented and waited on here and decremented in the event handler.

collector/internal/telemetryapi/listener.go
```diff
@@ -132,13 +156,8 @@ func (s *Listener) Wait(ctx context.Context, reqID string) error {
...
 				}
-				s.logger.Debug("Event processed", zap.Any("event", i))
-				if i.Type != "platform.runtimeDone" {
-					continue
-				}
-
-				if i.Record["requestId"] == reqID {
-					return nil
+				if err := s.handleEvent(ctx, i); err != nil {
+					return err
```
Aneurysm9: Should this return on error?  I think this function will only get called once and thus returning on error here will prevent any further events from processing.  Does this eventually result in the collector extension being terminated?

collector/internal/telemetryapi/listener.go
Aneurysm9: This should be the only thing reading from the handler list and ideally nothing is writing to it after initialization, so there should be very little to no contention on this lock.

codeboten: Agreed that lock contention should be minimal, changed the code to use a read lock nonetheless

collector/internal/lifecycle/manager.go
codeboten: You're right, I've updated the code to use a separate waitgroup for HandleEvent, as well as a lock for reading/writing the `reqID`. I've also added unit tests around the execution of processEvents and HandleEvent. PTAL

collector/internal/telemetryapi/listener.go
codeboten: Good point, I've updated this to log a warning instead of returning. The listener's behaviour shouldn't be impacted by what the registered handlers are doing w/ the event.

tylerbenson:
@codeboten What's the status here?  Is this worth revisiting?

codeboten:
@tylerbenson probably not worth doing at this time.

PR #416, opened 2023-01-12
[chore] add install-tools make target

codeboten:
This will allow us to enable linting/fmt in the CI.


Aneurysm9 (approved):

Aneurysm9:
https://github.com/open-telemetry/opentelemetry-lambda/actions/runs/3906316874/jobs/6674356296#step:5:10 looks like an `empty.go` file in `internal/tools` without build tag restrictions might be needed.  Or that module needs to be excluded.

codeboten (commented):
Opted to exclude that directory

PR #427, opened 2023-01-16
[chore] fix linting issues

codeboten:
Follow up to https://github.com/open-telemetry/opentelemetry-lambda/pull/417

### DIFF
--- a/collector/internal/telemetryapi/listener.go
+++ b/collector/internal/telemetryapi/listener.go
@@ -94,7 +94,9 @@ func (s *Listener) httpHandler(w http.ResponseWriter, r *http.Request) {
 	_ = json.Unmarshal(body, &slice)
 
 	for _, el := range slice {
-		s.queue.Put(el)
+		if err := s.queue.Put(el); err != nil {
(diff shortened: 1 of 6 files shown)


collector/internal/telemetryapi/listener.go
```diff
@@ -94,7 +94,9 @@ func (s *Listener) httpHandler(w http.ResponseWriter, r *http.Request) {
 	_ = json.Unmarshal(body, &slice)
 
 	for _, el := range slice {
-		s.queue.Put(el)
+		if err := s.queue.Put(el); err != nil {
+			s.logger.Error("Failed to put event in queue", zap.Error(err))
```
tsloughter, line 98: Should this really be Error level? Not sure what the convention is in the collector but I'd expect a lower level for something that can be safely skipped like this.

Aneurysm9: This could result in failing to export data and/or failing to request the next event from the extension API which can have runtime cost implications.  I think it is probably appropriate to log this at the error level.

Aneurysm9 (approved):

PR #430, opened 2023-01-17
add feature matrix

codeboten:
Adding the initial feature matrix with some descriptions of the features captured in the matrix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # OpenTelemetry Lambda
+
 ![GitHub Java Workflow Status](https://img.shields.io/github/actions/workflow/status/open-telemetry/opentelemetry-lambda/ci-java.yml?branch%3Amain&label=CI%20%28Java%29&style=for-the-badge)
 ![GitHub Collector Workflow Status](https://img.shields.io/github/actions/workflow/status/open-telemetry/opentelemetry-lambda/ci-collector.yml?branch%3Amain&label=CI%20%28Collector%29&style=for-the-badge)
 ![GitHub NodeJS Workflow Status](https://img.shields.io/github/actions/workflow/status/open-telemetry/opentelemetry-lambda/ci-nodejs.yml?branch%3Amain&label=CI%20%28NodeJS%29&style=for-the-badge)
@@ -6,6 +7,7 @@
 ![GitHub Python Pull Request Workflow Status](https://img.shields.io/github/actions/workflow/status/open-telemetry/opentelemetry-lambda/pr-python.yml?branch%3Amain&label=Pull%20Request%20%28Python%29&style=for-the-badge)
 
 ## OpenTelemetry Lambda Layers
+
 The OpenTelemetry Lambda Layers provide the OpenTelemetry (OTel) code to export telemetry asynchronously from AWS Lambdas. It does this by embedding a stripped-down version of [OpenTelemetry Collector Contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) inside an [AWS Lambda Extension Layer](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-extensions-in-preview/).
 
 Some layers include the corresponding OTel language SDK for the Lambda. This allows Lambdas to use OpenTelemetry to send traces and metrics to any configured backend.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,12 +20,53 @@ Some layers include the corresponding OTel language SDK for the Lambda. This all
...
+
+The following is a list of features provided by the OpenTelemetry layers.
+
+### Custom Context Propagation
+
+Context can be propagated through various mechanisms (e.g. http headers (APIGW), message attributes (SQS), ...). In some cases, it may be required to pass a custom context propagation extractor in lambda through configuration, this feature allows this through Lambda instrumentation configuration.
+
+### Disable AWS Context propagation via environment variable
+
+The context propagation when Active Tracing is enabled uses the `_X_AMZN_TRACE_ID` environment variable present in the Lambda runtime. To disable this behaviour, implementations have added the `OTEL_LAMBDA_DISABLE_AWS_CONTEXT_PROPAGATION` environment variable which can be set to `true`.
```
tsloughter: This isn't in the spec is it?

README.md
```diff
@@ -18,12 +20,53 @@ Some layers include the corresponding OTel language SDK for the Lambda. This all
...
+| Python 3.9    |                            |                         |                      |
+| Python 3.8    |                            |                         |                      |
+| Python 3.7    |                            |                         |                      |
+| Java 11       |                            |                         |                      |
+| Java 8        |                            |                         |                      |
+| .NET Core 3.1 |                            |                         |                      |
+| .NET 6        |                            |                         |                      |
+| .NET 5        |                            |                         |                      |
+| Go 1.x        |                            |                         |                      |
+| Ruby 2.7      |                            |                         |                      |
```
tsloughter: Are there actually differences in whether they support each of these things between the version of the languages?

I think breaking out the semantic conventions to at least `Trace General`, `Trace Incoming`, `Trace Outgoing` and `Metrics` and transposing the matrix so the languages are the headers would be better -- partly assuming languages can be combined so there aren't 11 columns.

README.md
codeboten: It is not, there are two language implementations I'm aware of that are close to or have added it: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1227 https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1507

README.md
codeboten: I don't know if there *should* be differences between the language versions, I want to treat this matrix as a bit of a  test plan, to ensure we've at least confirmed this works in all the supported languages. I worry that otherwise we would say "X works in node.js 16 therefore will work in all node.js, we won't test it", only to find out there is a slight difference in some library's implementation or dependency causing X not to work

codeboten: We could always combine the languages w/ the understanding that something wouldn't be marked as "supported" in any language until its been tested in all supported runtimes?

avzis (commented):
should this matrix also include a section about auto-instrumentation?

README.md
tsloughter: I'd support that (combine them and only mark it if all are supported). Or could mark with a different character, plus have a footnote if 1 version has a specific reason it will never support this.

codeboten:
@tsloughter @avzis updated the matrix based on your feedback, please take a look

README.md
tsloughter: OK, should probably leave it out until it is finalized as a feature?

README.md
```diff
@@ -18,12 +20,53 @@ Some layers include the corresponding OTel language SDK for the Lambda. This all
...
 
 ## Design Proposal
+
 To get a better understanding of the proposed design for the OpenTelemetry Lamda extension, you can see the [Design Proposal here.](docs/design_proposal.md)
+
+## Features
+
+The following is a list of features provided by the OpenTelemetry layers.
+
+### Custom Context Propagation
```
Aneurysm9:
```suggestion
### Custom Context Propagation Carrier Extraction
```

All SDKs support custom context propagation.  What I think we're discussing here is the functionality to provide a hook to extract the carrier (or context, but we should be changing those implementations) from the incoming event.

README.md
Aneurysm9: I think we should  change this to be an option to link a context extracted from the Lambda runtime environment to the instrumentation-generated span rather than disabling that context extraction entirely.

README.md
```diff
@@ -18,12 +20,53 @@ Some layers include the corresponding OTel language SDK for the Lambda. This all
...
+
+### Auto instrumentation
+
+The Lambda layer includes support for automatically instrumentation code via the use of instrumentation libraries.
+
+### Support matrix
+
+The table below captures the state of various features and their levels of support different runtimes.
+
+| Feature                    | Node | Python | Java | .NET | Go   | Ruby |
```
Aneurysm9, line 62: We don't have Go, .NET, or Ruby layers.  They use the stock collector-only layer.  Should this refer to the Lambda instrumentations in each language instead?

codeboten: I think there was discussion from @tsloughter about adding a ruby layer based on the current splunk implementation.

I would expect the other layers to be empty until they're implemented, but i see what you're saying, why have them at all if they're not supported yet.

Aneurysm9: It's not so much that there are empty layers that will eventually not be empty.  Go will never have a layer that is more than the bare collector as the instrumentation needs to be compiled into the application.  I think .NET is similar.

README.md
codeboten: @Aneurysm9 could this be documented in a separate line? i'd like to capture what's there today

codeboten: I've added a separate line for span link context propagation

README.md
codeboten: Fair enough, in that case I would capture the collector as a separate "feature" of the layer, to make clear that those other supported runtimes weren't forgotten. Otherwise I would expect users to see the dotnet layer in this repository and ask "what features does it support"

tsloughter: Not exactly the case for Go, there is the auto-instrumentation Go implementation now. So there may be a layer.

But yea, Ruby isn't there yet and I can add it when I PR the layer.

README.md
tylerbenson: (FWIW, I'm also in favor of removing the need for the setting by shifting the env var context to a span link instead of formalizing a setting for disabling it. Documenting where the setting is currently implemented is fine though.)

codeboten: I've removed the line for disabling context propagation via env variable.

README.md
```diff
@@ -18,12 +20,65 @@ Some layers include the corresponding OTel language SDK for the Lambda. This all
...
+| -------------------------- | :--: | :----: | :--: | :--: | :--: | :--: |
+| OpenTelemetry collector    |  +   |  +     |  +   |  +   |      |      |
```
tsloughter-splunk, line 64:
```suggestion
| Layer                                      |       |         |      |      |      |      |
| - OpenTelemetry collector    |  +   |  +     |  +   |  +   |      |      |
| - Auto-instrumentation         |       |        |      |      |      |      |
```

This is what I was suggesting to convey that there will still be layers for languages like Go but they only have the collector.

tsloughter-splunk (approved):
I had one more comment but its not blocking.

Aneurysm9 (approved):

tsloughter (approved):
Was on the wrong account in my first approval :)

README.md
codeboten: right I guess in my mind everything on this list pertains to the layer :) happy to talk about this in a follow up pr 👍🏻

PR #435, opened 2023-01-20
Add configureInstrumentations for users to customize instrumentations

lpsinger:


codeboten (commented):
Thanks for the contribution! Please sign the CLA

lpsinger:
> Thanks for the contribution! Please sign the CLA

It's with my organization right now.

lpsinger:
It's signed now!

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -69,6 +70,7 @@ const instrumentations = [
   new NetInstrumentation(),
   new PgInstrumentation(),
   new RedisInstrumentation(),
```
Aneurysm9: Should these all be moved into a default implementation of `configureInstrumentations()` such that users have a way to ask for less auto-instrumentation and not just more?  Minimizing the impact of auto-instrumentation on cold start times is a recurring theme we're hearing from customers.

lpsinger: Suits me. Done.

Note, I have no good way to test this.

Aneurysm9 (approved):

lpsinger:
Thank you! When will this change take effect in the Lambda layers on AWS that are listed at https://aws-otel.github.io/docs/getting-started/lambda/lambda-js?

lpsinger:
Hi, when are you going to upload a new Lambda layer?

PR #436, opened 2023-01-20
feat: add faas.instance resource attribute for python

povilasv:
Hey folks!

I've would like to add faas.instance resource attribute to the python autoinstrumentation. This is defined in: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/faas.md#faas-resource-attributes

AWS docs for that environment variable:

https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html

### DIFF
--- a/python/src/otel/otel_sdk/otel-instrument
+++ b/python/src/otel/otel_sdk/otel-instrument
@@ -133,7 +133,7 @@ fi
 #       )
 #   )
 
-export LAMBDA_RESOURCE_ATTRIBUTES="cloud.region=$AWS_REGION,cloud.provider=aws,faas.name=$AWS_LAMBDA_FUNCTION_NAME,faas.version=$AWS_LAMBDA_FUNCTION_VERSION";
+export LAMBDA_RESOURCE_ATTRIBUTES="cloud.region=$AWS_REGION,cloud.provider=aws,faas.name=$AWS_LAMBDA_FUNCTION_NAME,faas.version=$AWS_LAMBDA_FUNCTION_VERSION,faas.instance=$AWS_LAMBDA_LOG_STREAM_NAME";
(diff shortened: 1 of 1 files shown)


python/src/otel/otel_sdk/otel-instrument
```diff
@@ -133,7 +133,7 @@ fi
 #       )
 #   )
 
-export LAMBDA_RESOURCE_ATTRIBUTES="cloud.region=$AWS_REGION,cloud.provider=aws,faas.name=$AWS_LAMBDA_FUNCTION_NAME,faas.version=$AWS_LAMBDA_FUNCTION_VERSION";
+export LAMBDA_RESOURCE_ATTRIBUTES="cloud.region=$AWS_REGION,cloud.provider=aws,faas.name=$AWS_LAMBDA_FUNCTION_NAME,faas.version=$AWS_LAMBDA_FUNCTION_VERSION,faas.instance=$AWS_LAMBDA_LOG_STREAM_NAME";
```
codeboten, line 136: The change looks ok, I'm a bit surprised there isn't a better environment variable to use to identify the specific instance.

codeboten (commented):
The change looks ok, would be good to have someone from aws to confirm this is best variable to use @Aneurysm9

python/src/otel/otel_sdk/otel-instrument
Aneurysm9: I think this serves the purpose effectively and I'm not aware of any better alternative.  I don't think it's a one-way door as any useful alternative would also provide a unique ID.

Aneurysm9 (approved):

codeboten (approved):

PR #471, opened 2023-02-06
Support for node18.x runtime and remove nodejs12.x

vasireddy99:
This PR is created to add support of nodejs18.x version and removes the support of node12.x/node 12 and below.

Supported [Node Runtimes](https://github.com/open-telemetry/opentelemetry-js#supported-runtimes) in OTel-JS

node js 18.x support https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/382

### DIFF
--- a/java/integration-tests/aws-sdk/agent/main.tf
+++ b/java/integration-tests/aws-sdk/agent/main.tf
@@ -10,7 +10,7 @@ resource "aws_lambda_layer_version" "collector_layer" {
   count               = var.enable_collector_layer ? 1 : 0
   layer_name          = var.collector_layer_name
   filename            = "${path.module}/../../../../collector/build/collector-extension.zip"
-  compatible_runtimes = ["nodejs12.x", "nodejs14.x", "nodejs16.x"]
+  compatible_runtimes = ["nodejs14.x", "nodejs16.x", "nodejs18.x"]
   license_info        = "Apache-2.0"
   source_code_hash    = filebase64sha256("${path.module}/../../../../collector/build/collector-extension.zip")
 }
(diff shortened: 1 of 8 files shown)


java/integration-tests/aws-sdk/agent/main.tf
```diff
@@ -10,7 +10,7 @@ resource "aws_lambda_layer_version" "collector_layer" {
   count               = var.enable_collector_layer ? 1 : 0
   layer_name          = var.collector_layer_name
   filename            = "${path.module}/../../../../collector/build/collector-extension.zip"
-  compatible_runtimes = ["nodejs12.x", "nodejs14.x", "nodejs16.x"]
+  compatible_runtimes = ["nodejs14.x", "nodejs16.x", "nodejs18.x"]
```
Aneurysm9, line 13: Why does this claim to be compatible with node in the Java tests.  Same elsewhere with python.

vasireddy99: Yeah, per my understanding. these runtimes are for compatibility with the [collector-extension](https://github.com/open-telemetry/opentelemetry-lambda/pull/471/files#diff-54ad11a2f295cbd18f1d7e86050dd31b21688a8e93c3a0fddf45a65894279154L12) part. which would be later combined with [sdk laye](https://github.com/open-telemetry/opentelemetry-lambda/pull/471/files#diff-04e41d13ae1896a5a9925ce061290ccc6b47c908eacb3b4f53a98c7aa58ea861L6)r into a layer like [here](https://github.com/aws-observability/aws-otel-lambda/blob/main/java/build.sh#L26)

Aneurysm9 (approved):

PR #473, opened 2023-02-06
fix: format shell script and set option

bpgould:
Remove extra space and set pipefail option

set -o pipefail
This setting prevents errors in a pipeline from being masked. If any command in a pipeline fails, that return code will be used as the return code of the whole pipeline. By default, the pipeline's return code is that of the last command even if it succeeds.

### DIFF
--- a/utils/sam/run.sh
+++ b/utils/sam/run.sh
@@ -2,6 +2,7 @@
 
 set -e
 set -u
+set -o pipefail
 
 echo_usage() {
 	echo "usage: Deploy Lambda layer/application by SAM"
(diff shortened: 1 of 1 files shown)


bpgould:
I signed the CLA, but I am not sure if I can re-run the CI check

tsloughter:
Was there something failing that led you to add this? I may be missing it but I don't see any pipes being used in this script. Does something get translated to a pipeline?

bpgould:
> Was there something failing that led you to add this? I may be missing it but I don't see any pipes being used in this script. Does something get translated to a pipeline?

I was browsing the repo and noticed the whitespace, thought I would make a quick PR. Pipefail is generally a best practice for any shell script, thus I don't think YAGNI applies.

tylerbenson:
Unfortunately quick PRs still require the CLA.  Do you want to go through that process or should we close the PR?

I do appreciate the consideration though.  For future reference, here's the recommended settings for bash scripts:
https://sipb.mit.edu/doc/safe-shell/

tylerbenson:
Fixed in #1226

PR #477, opened 2023-02-07
Change OTEL_METRICS_EXPORTER env variable default from none to otlp

### LINKED ISSUE #475: Metric exporting is turned-off by default
**Describe the bug**
[Metric exporting is turned-off by default](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/otel_sdk/otel-instrument#L99)

This is super confusing and un-documented.

Caused me to spend multiple hours to understand why my metrics are not reaching the embedded collector.

Related to the issue https://github.com/aws-observability/aws-otel-lambda/issues/439

humivo:
Will fix [Issue #475](https://github.com/open-telemetry/opentelemetry-lambda/issues/475) by changing the default value for the OTEL_METRICS_EXPORTER from none to otlp.

### DIFF
--- a/python/src/otel/otel_sdk/otel-instrument
+++ b/python/src/otel/otel_sdk/otel-instrument
@@ -96,7 +96,7 @@ fi
 # - Set the metrics exporter
 
 if [ -z "${OTEL_METRICS_EXPORTER}" ]; then
-    export OTEL_METRICS_EXPORTER=none;
+    export OTEL_METRICS_EXPORTER=otlp;
 fi
 
 # - Set the service name


vasireddy99 (approved):

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -96,7 +96,7 @@ fi
 # - Set the metrics exporter
 
 if [ -z "${OTEL_METRICS_EXPORTER}" ]; then
-    export OTEL_METRICS_EXPORTER=none;
+    export OTEL_METRICS_EXPORTER=otlp;
```
rapphil, line 99: should we avoid setting anything explicitly so that we rely on what is on the spec/sdk? https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/#exporter-selection

rapphil: Not setting is more future proof than setting in the unlikely event of default is changed.

Aneurysm9: I agree that this wrapper probably shouldn't change the default behavior in this way.

humivo: Ok so setting this OTEL_METRICS_EXPORTER environment variable should be removed overall so the default behavior in the sdk can be used?

humivo:
A new [PR #478](https://github.com/open-telemetry/opentelemetry-lambda/pull/478) was made to track this issue as well as fix other environment variable defaults and so the discussion shall continue there.  This PR will now be closed.

PR #478, opened 2023-02-09
Set OTEL_EXPORTER_OTLP_PROTOCOL to `http/protobuf`

Kausik-A:
This PR addresses the [deprecation of otlp_proto_grpc and otlp_proto_http](https://github.com/open-telemetry/opentelemetry-python-contrib/commit/83f37eddf2c7f2dac77128b6b27bef57ba167e90) in the python SDK `v0.36b0`. `OTEL_EXPORTER_OTLP_PROTOCOL` is set default to `grpc`. We set default for traces as `http/protobuf`

Edit: As per the suggestion using `OTEL_EXPORTER_OTLP_PROTOCOL` instead of `OTEL_EXPORTER_OTLP_PROTOCOL`


tsloughter:
Why do we need to set the default instead of using python's default?

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -89,14 +89,9 @@ export PYTHONPATH="$LAMBDA_RUNTIME_DIR:$PYTHONPATH";
...
-    export OTEL_TRACES_EXPORTER=otlp_proto_http;
-fi
-
-# - Set the metrics exporter
-
-if [ -z "${OTEL_METRICS_EXPORTER}" ]; then
-    export OTEL_METRICS_EXPORTER=none;
+# - If OTEL_EXPORTER_OTLP_TRACES_PROTOCOL is not set by user, the default exporting protocol for traces is http/protobuf.
+if [ -z "${OTEL_EXPORTER_OTLP_TRACES_PROTOCOL}" ]; then
+    export OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=http/protobuf
```
Aneurysm9:
```suggestion
# - If OTEL_EXPORTER_OTLP_PROTOCOL is not set by user, the default exporting protocol is http/protobuf.
if [ -z "${OTEL_EXPORTER_OTLP_PROTOCOL}" ]; then
    export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
```

I think we should set the signal-agnostic protocol variable if not set and leave signal-specific settings to the user.  [The spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol) says this should be the default in the absence of explicit configuration.

Kausik-A:
> Why do we need to set the default instead of using python's default?

The current default behavior of the lambda instrumentation is to use OTLP/HTTP for traces due to that setting in the [wrapper](https://github.com/open-telemetry/opentelemetry-lambda/blob/f2a3a901603edd0491af61e27566260b1dfbec45/python/src/otel/otel_sdk/otel-instrument#L93). Leaving that to  default `gRPC` may be a breaking change for users who had a collector configuration listening for HTTP and not gRPC.

python/src/otel/otel_sdk/otel-instrument
Kausik-A: Would this not change the exporting protocol for metrics too?

tsloughter:
Ok, I think we should consider that breaking change. But that shouldn't hold up this PR since it isn't changing the behaviour, I'll bring this up separately.

python/src/otel/otel_sdk/otel-instrument
tsloughter: I don't think we should be overriding the default at all, so would argue leaving it to only traces is better. But this means adding back `export OTEL_METRICS_EXPORTER=none;` in this PR.

Aneurysm9: > But this means adding back `export OTEL_METRICS_EXPORTER=none;` in this PR.

Why?  This seems to be clearly wrong and the need to explicitly set a configuration option that is supposed to be the default has been a point of confusion for a number of users.

Aneurysm9 (approved):

python/src/otel/otel_sdk/otel-instrument
tsloughter: It is technically a breaking change. I think both should be using the defaults. So a separate PR where both exports are removed and the fact it is a breaking change can be debated.

Kausik-A: Okay, I will open up another PR to discuss this breaking change.

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -89,14 +89,9 @@ export PYTHONPATH="$LAMBDA_RUNTIME_DIR:$PYTHONPATH";
 
 # - Set the trace exporter
```
vasireddy99: Could we change/remove this line, would be easy to understand

vasireddy99: to understand the script

tsloughter (approved):

python/src/otel/otel_sdk/otel-instrument
Kausik-A: done

PR #531, opened 2023-02-27
Disable CGO

Kausik-A:
Update `CGO_ENABLED=0`

This is to avoid the below error message while trying to invoke the `go sample app` using `provided.al2 runtime` in AWS lambda

`/var/task/bootstrap: /lib64/libc.so.6: version GLIBC_2.34' not found (required by /var/task/bootstrap)`

### DIFF
--- a/go/sample-apps/function/build.sh
+++ b/go/sample-apps/function/build.sh
@@ -4,6 +4,7 @@ set -e
 GOARCH=${GOARCH-amd64}
 
 mkdir -p build
+export CGO_ENABLED=0 
 GOOS=linux go build -o ./build/bootstrap .
 cd build
 zip bootstrap.zip bootstrap
--- a/go/sample-apps/function/build.sh
+++ b/go/sample-apps/function/build.sh
@@ -4,7 +4,6 @@ set -e
 GOARCH=${GOARCH-amd64}
 
 mkdir -p build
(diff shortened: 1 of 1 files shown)


go/sample-apps/function/build.sh
```diff
@@ -4,6 +4,7 @@ set -e
 GOARCH=${GOARCH-amd64}
 
 mkdir -p build
+export CGO_ENABLED=0 
 GOOS=linux go build -o ./build/bootstrap .
```
Aneurysm9:
```suggestion
CGO_ENABLED=0 GOOS=linux go build -o ./build/bootstrap .
```

This should probably include `GOARCH` as well, though I'm not sure the setting of `GOARCH` on line 4 is correct.

Kausik-A: It does pick the env var `GOARCH` even-though it is set above. Looks like `GOARCH` is set to `amd64` if it is not defined. When building the layer, we explicitly declare the value of `GOARCH` before running `./build.sh`. One of such example is [here](https://github.com/aws-observability/aws-otel-lambda/blob/5870d5cde57134f25b211bbf67f9b60fc91fccfc/.github/workflows/pr-build.yml#L80). So we should be good right?

Aneurysm9 (approved):

Issue #534, opened 2023-03-01
Python lambda layer does not support GRPC


rapphil:
**Describe the bug**
The python lambda layer does not support grpc. Support for it was removed in this PR: https://github.com/open-telemetry/opentelemetry-lambda/pull/181

This is a bit misleading and not in parity with other languages.

Other questions:
* Is there any performance benefit/considerations in using GRPC over http/protobuf. I feel that since the collector in the lambda layers is in the same "host" as the application, we should be very opinionated and hard code this settingin every language. I.e.: what option will give us the best performance/smallest footprint given that we are on the same "host"? Reason for hard coding is because it is one less knob for users to deal with.

**What did you expect to see?**
I'd expect that grpc OTLP exporter is supported or that we explicitly call out that it is not supported and why.

I also expect that we provide opinionated hard coded defaults for every language and for the default collector configuration.

**What language layer did you use?**
Python

**Additional context**
The main issue with grpc is that it requires access to native libraries. We have to take special care when building the layer for the arm64 architecture.

However building for arm and x86 is achievable if we use docker buildx.

PR #535, opened 2023-03-01
Add support to grpc in the python lambda layer

rapphil:
Add support to building lambda layers with grpc in Python.

this is an alternative for solving #534

It will assumes that you have buildx properly installed and configured.

I'm creating this a draft so that we can evaluate this.

### DIFF
--- a/python/src/build.sh
+++ b/python/src/build.sh
@@ -1,6 +1,10 @@
 #!/bin/sh
 set -e
 
+## Valid platform values are "amd64" and "arm64"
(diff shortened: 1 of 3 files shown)


tylerbenson:
@rapphil Do you still want to pursue this or should we close it?

jriverauala (approved):

rapphil:
Closing since this did not get much traction. It is a nice to have but I'm not 100% confident on the approach. We can reopen if anyone is interested in having support to grpc.

PR #600, opened 2023-03-29
[chore] add release github action

codeboten:
This action is following https://github.com/aws-observability/aws-otel-lambda/blob/main/.github/workflows/release.yml, minus the integration tests.

### DIFF
--- a/collector/Makefile
+++ b/collector/Makefile
@@ -44,17 +44,16 @@ package: build
 	@echo Package zip file for collector extension layer
 	mkdir -p $(BUILD_SPACE)/collector-config
 	cp config* $(BUILD_SPACE)/collector-config
-	cd $(BUILD_SPACE) && zip -r collector-extension.zip collector-config extensions
+	cd $(BUILD_SPACE) && zip -r collector-extension-$(GOARCH).zip collector-config extensions
+
+.PHONY: publish
+publish:
+	aws lambda publish-layer-version --layer-name $(LAYER_NAME) --zip-file fileb://$(BUILD_SPACE)/collector-extension-$(GOARCH).zip --compatible-runtimes nodejs14.x nodejs16.x nodejs18.x java11 python3.8 python3.9 --query 'LayerVersionArn' --output text
 
 .PHONY: publish-layer
 publish-layer: package
 	@echo Publishing collector extension layer...
-	aws s3 mb s3://$(BUCKET_NAME)
-	aws s3 cp $(BUILD_SPACE)/collector-extension.zip s3://$(BUCKET_NAME)
-	aws lambda publish-layer-version --layer-name $(LAYER_NAME) --content S3Bucket=$(BUCKET_NAME),S3Key=collector-extension.zip --compatible-runtimes nodejs14.x nodejs16.x nodejs18.x java11 python3.8 python3.9 --query 'LayerVersionArn' --output text
-	@echo Clearing cached files...
-	aws s3 rm s3://$(BUCKET_NAME)/collector-extension.zip
-	aws s3 rb s3://$(BUCKET_NAME)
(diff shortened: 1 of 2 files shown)


collector/Makefile
```diff
@@ -29,28 +29,31 @@ $(GOMODULES):
...
 	@echo Package zip file for collector extension layer
 	mkdir -p $(BUILD_SPACE)/collector-config
 	cp config* $(BUILD_SPACE)/collector-config
-	cd $(BUILD_SPACE) && zip -r collector-extension.zip collector-config extensions
+	cd $(BUILD_SPACE) && zip -r collector-extension-$(GOARCH).zip collector-config extensions
 
-publish-layer: package
+.PHONY: publish
+publish:
+	aws lambda publish-layer-version --layer-name $(LAYER_NAME) --zip-file fileb://$(BUILD_SPACE)/collector-extension-$(GOARCH).zip --compatible-runtimes nodejs14.x nodejs16.x nodejs18.x java11 python3.8 python3.9 --query 'LayerVersionArn' --output text
```
tsloughter, line 51: Should this include the custom runtime?

collector/Makefile
```diff
@@ -29,28 +29,31 @@ $(GOMODULES):
...
+
+.PHONY: publish-layer
+publish-layer: package publish
```
tsloughter: Why is there `publish` and `publish-layer`. They appear to do the same thing?

tsloughter (commented):
Looks good, just had a couple questions first.

tsloughter:
Is this a draft because you plan to add the tests before making it ready to review?

collector/Makefile
codeboten: custom runtime? you mean in the list of compatible-runtimes?

collector/Makefile
codeboten: I didn't want to change the `publish-layer` as it's mentioned in the docs and calls package to build the collector. I needed something that only did the publishing aspect

codeboten:
> Is this a draft because you plan to add the tests before making it ready to review?

This is a draft because i still had a couple of things to edit :) 

I'll add issues around the other work that's needed (ie. create a release, add any tests)

collector/Makefile
codeboten:
```suggestion
publish-layer: package
```

tylerbenson (approved):

collector/Makefile
tsloughter: Right, to the list of runtimes.

tsloughter: Shouldn't it be `publish-layer: package publish` instead of duplicating the `aws lambda publish-layer-version` command?

collector/Makefile
tsloughter: Shouldn't it be publish-layer: package publish instead of duplicating the aws lambda publish-layer-version command?

tsloughter (approved):
Had the one last suggestion/comment about the Makefile.

collector/Makefile
codeboten: According to the [aws docs](https://docs.aws.amazon.com/cli/latest/reference/lambda/publish-layer-version.html), this is the list of valid values, not seeing custom in there:

```
  nodejs
  nodejs4.3
  nodejs6.10
  nodejs8.10
  nodejs10.x
  nodejs12.x
  nodejs14.x
  nodejs16.x
  java8
  java8.al2
  java11
  python2.7
  python3.6
  python3.7
  python3.8
  python3.9
  dotnetcore1.0
  dotnetcore2.0
  dotnetcore2.1
  dotnetcore3.1
  dotnet6
  nodejs4.3-edge
  go1.x
  ruby2.5
  ruby2.7
  provided
  provided.al2
  nodejs18.x
```

Is it supported but not documented?

collector/Makefile
codeboten: Right, the only difference is with the output from the `echo` calls in `publish-layer`. If theres no problem losing those, i'm happy to change it in a follow up PR. It's possible this isn't being used

collector/Makefile
tsloughter: I think `provided` and `provided.al2` are the custom runtimes. No idea why they are named provided or what that means :)

tsloughter: Hm, I guess it probably means "user provided".

PR #635, opened 2023-04-17
Modify TF script to get the multi-arch extension

vasireddy99:
This PR is created to modify the tf script to accept the multi arch extensions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,5 @@ terraform.*
 __pycache__/*
 .pyc
 
-.json
 .dll
 .cache
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ terraform.*
 __pycache__/*
 .pyc
 
+.json
+.dll
+.cache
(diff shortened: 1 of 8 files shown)


tsloughter:
I don't see the variables being set anywhere to make use of the arm option?

vasireddy99:
> I don't see the variables being set anywhere to make use of the arm option?

This test suite is used in the distribution of [ADOT](https://github.com/aws-observability/aws-otel-lambda/blob/main/.github/workflows/main-build.yml#L29) with [`amd64` and `arm64`](https://github.com/aws-observability/aws-otel-lambda/blob/main/.github/workflows/main-build.yml#L175) for integ testing and publishing.

codeboten (commented):
Are these terraform files used anywhere in this repo? I dont think we're using them for any integrations at this time, maybe we can remove them entirely? If adot is making use of them, could they live in that repo instead?

vasireddy99:
> Are these terraform files used anywhere in this repo? I dont think we're using them for any integrations at this time, maybe we can remove them entirely? If adot is making use of them, could they live in that repo instead?

yeah, these files are not used in this repo, maybe we can use these in the repo if there are any plans for the integ testing. I can create an issue to get this discussion in place to remove from these repo or to how they can be used in the repo. meanwhile, would like to get this PR reviewed to unblock the downstream(ADOT) distributions  🙂.

vasireddy99:
Created this Issue: https://github.com/open-telemetry/opentelemetry-lambda/issues/643

vasireddy99:
Closing this PR and will continue the discussion on the issue.

PR #646, opened 2023-04-20
Release workflow for nodejs layer

martinkuba:
This is similar to the existing Java release workflow but for Node.js.

I have also added the `version.js` script in order to get the Javascript SDK version.

### DIFF
--- a/.github/workflows/layer-publish.yml
+++ b/.github/workflows/layer-publish.yml
@@ -53,7 +53,7 @@ jobs:
             LAYER_NAME=$LAYER_NAME-${{ inputs.release-group }}
           fi
           
-          LAYER_VERSION=$(echo "$GITHUB_REF_NAME" | sed -r 's/.*\/[^\d\.]*//g')
+          LAYER_VERSION=$(echo "$GITHUB_REF_NAME" | sed -r 's/.*\/[^0-9\.]*//g')
           LAYER_VERSION_CLEANED=$(echo "$LAYER_VERSION" | sed -r 's/\./_/g')
           
           LAYER_NAME=$LAYER_NAME-$LAYER_VERSION_CLEANED
--- a/.github/workflows/release-layer-nodejs.yml
+++ b/.github/workflows/release-layer-nodejs.yml
@@ -1,0 +1,76 @@
+name: "Release NodeJS Lambda Layer"
+
+on:
+  # (Using tag push instead of release to allow filtering by tag prefix.)
+  push:
+    tags:
+      - layer-nodejs/**
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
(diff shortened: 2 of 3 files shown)


.github/workflows/layer-publish.yml
```diff
@@ -53,7 +53,7 @@ jobs:
             LAYER_NAME=$LAYER_NAME-${{ inputs.release-group }}
           fi
           
-          LAYER_VERSION=$(echo "$GITHUB_REF_NAME" | sed -r 's/.*\/[^\d\.]*//g')
+          LAYER_VERSION=$(echo "$GITHUB_REF_NAME" | sed -r 's/.*\/[^0-9\.]*//g')
```
martinkuba, line 56: The previous regex was leaving out the first digit for some reason, e.g. `layer-nodejs/1.2.3` resulted in version `.2.3`.

tylerbenson: Yeah, I added `\d` not realizing that sed doesn't recognize it.  Thanks for fixing.

.github/workflows/release-layer-nodejs.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Save Node SDK Version
+        id: save-node-sdk-version
+        run: |
+          SDK_VERSION=$(node version.js)
+          echo "SDK_VERSION=$SDK_VERSION" >> $GITHUB_OUTPUT
+        working-directory: nodejs/packages/layer/scripts
+
+      - name: Rename zip file
+        run: mv layer.zip opentelemetry-nodejs-layer.zip
+        working-directory: nodejs/packages/layer/build
```
tylerbenson, line 39: I don't think this step is necessary.  Just rename the upload artifact path below to `layer.zip`

tylerbenson (approved):

.github/workflows/release-layer-nodejs.yml
martinkuba: If I don't rename it, then the downloadable artifact of the completed GH action job will be called `layer.zip`. I was trying to make it consistent with the Java action.

tylerbenson: Really?  I thought the name was set by the upload artifact.  🤷

martinkuba: It does, but the artifact name and the zip file name must be the same with how the publish workflow is written right now. We are passing the artifact name to the publish workflow, and the name is used to 1) download the artifact and 2) as the name of the zip file to upload to AWS. If I use different values, then when you download the artifact in the workflow, you will get the `layer.zip` file (not `[artifact-name].zip`).

codeboten (approved):

PR #667, opened 2023-05-02
Add layer compatibility metadata

tylerbenson:
I haven't actually tested this, so let me know if anything looks off...

### DIFF
--- a/.github/workflows/layer-publish.yml
+++ b/.github/workflows/layer-publish.yml
@@ -55,6 +55,12 @@ jobs:
           else
             ARCH="x86_64 arm64"
           fi
+          echo "ARCH=$ARCH" >> $GITHUB_ENV
+          
+          if [[ -n "${{ inputs.runtimes }}" ]]; then
+            RUNTIMES="--compatible-runtimes ${{ inputs.runtimes }}"
+          fi
+          echo "RUNTIMES=$RUNTIMES" >> $GITHUB_ENV
           
           if [[ "${{ inputs.release-group }}" != "prod" ]]; then
             LAYER_NAME=$LAYER_NAME-${{ inputs.release-group }}
@@ -65,7 +71,6 @@ jobs:
           
           LAYER_NAME=$LAYER_NAME-$LAYER_VERSION_CLEANED
           echo "LAYER_NAME=$LAYER_NAME" >> $GITHUB_ENV
-          echo "ARCH=$ARCH" >> $GITHUB_ENV
           
           echo GITHUB_ENV:
           cat $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/release-layer-nodejs.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       layer-name: opentelemetry-nodejs
       component-version: ${{needs.build-layer.outputs.NODEJS_VERSION}}
       # architecture:
+      runtimes: nodejs14.x nodejs16.x nodejs18.x
```
tylerbenson, line 74: @martinkuba I included all the supported versions... Should this be extended further?

martinkuba: If this is intended for the Lambda runtime selection, then the only other one there is Node.js 12.x, which is not supported in OTel. So, I think this is the right list.

.github/workflows/layer-publish.yml
```diff
@@ -47,6 +51,9 @@ jobs:
           
           if [[ -n "${{ inputs.architecture }}" ]]; then
             LAYER_NAME=$LAYER_NAME-${{ inputs.architecture }}
+            ARCH="${{ inputs.architecture }}"
+          else
+            ARCH="x86_64 arm64"
```
rapphil, line 56: do we have any layer that is ARCH agnostic, i.e.: supports both x86_64 and arm64? As far I'm aware all the layers include the collector, which is architecture dependent.

tylerbenson: The collector is an independent layer from the automatic instrumentation. So both layers must be declared. As a result, most (all?) auto instrumentation layers are arch independent.

rapphil: you are right. I was confused by the aws lambda layers. 

Still on the `ARCH` environment variable, I think you need to echo the env var into `$GITHUB_ENV` for it to be accessible in subsequent steps, similarly to how it is done to `LAYER_NAME`.

reference: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

rapphil: nit: what do you think about making the `architecture` github input obligatory and defining explicitly in each workflow that calls `layer-publish.yaml` what are the supported architectures? this would simplify a tiny bit this workflow by removing the need of the the intermediate `ARCH` env var.

tylerbenson: Good catch on `$GITHUB_ENV`.  I've updated that now.

As for making `architecture` required, I think it would be a tradeoff.  It would shift the complexity from here to the layer naming.  When we support both, we don't want it included in the name. I think it's easier to keep as is.

rapphil (approved):

.github/workflows/layer-publish.yml
```diff
@@ -80,6 +88,8 @@ jobs:
             aws lambda publish-layer-version \
               --layer-name $LAYER_NAME \
               --license-info "Apache 2.0" \
+              --compatible-architectures $ARCH
+              --compatible-runtimes ${{ inputs.runtimes }}
```
codeboten: what happens if this is empty? won't the command fail here?

tylerbenson: Oops, good catch. Also missing the trailing `\`.  I'll fix that.

codeboten (approved):
Thanks for addressing my comments

PR #669, opened 2023-05-02
ensure collector version is produced by amd64

### LINKED ISSUE #652: Layer publishing for collector fails
**Describe the bug**
The layer publishing for the collector fails with the following:

```
Run COLLECTOR_VERSION=$( /home/runner/work/opentelemetry-lambda/opentelemetry-lambda/collector/build/extensions/collector -v | sed 's/^.* version //')
  COLLECTOR_VERSION=$( /home/runner/work/opentelemetry-lambda/opentelemetry-lambda/collector/build/extensions/collector -v | sed 's/^.* version //')
  echo "COLLECTOR_VERSION=$COLLECTOR_VERSION" >> $GITHUB_OUTPUT
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
/home/runner/work/_temp/47ec576f-363e-4798-99d5-5049b5846af9.sh: line 1: /home/runner/work/opentelemetry-lambda/opentelemetry-lambda/collector/build/extensions/collector: cannot execute binary file: Exec format error
```

This happens because the action runner is trying to run an arm64 build of the collector. See https://github.com/open-telemetry/opentelemetry-lambda/actions/runs/4801391934/jobs/8543532054#step:6:7

codeboten:
Otherwise this step fails as the gh runner can't run arm64. We only need this version to be set once anyways.

Fixes #652

### DIFF
--- a/.github/workflows/release-layer-collector.yml
+++ b/.github/workflows/release-layer-collector.yml
@@ -33,6 +33,7 @@ jobs:
           path: ${{ github.workspace }}/collector/build/*.zip
 
       - name: Save Collector Version
+        if: ${{ matrix.architecture == 'amd64' }}
         id: save-collector-version
         shell: bash
         # `./collector -v` output is in the form `otelcol-contrib version 0.75.0`


.github/workflows/release-layer-collector.yml
```diff
@@ -33,6 +33,7 @@ jobs:
           path: ${{ github.workspace }}/collector/build/*.zip
 
       - name: Save Collector Version
+        if: ${{ matrix.architecture == 'amd64' }}
```
tylerbenson, line 36: Do the builds run in order, or do we need to have a dependency to ensure the `amd64` build runs first and not in parallel?

tylerbenson (approved):

.github/workflows/release-layer-collector.yml
codeboten: Since publish `needs: [build-layer]` i think it should be fine

PR #678, opened 2023-05-04
update collector workflow to publish a prod layer

codeboten:
Also adding a fairly minimal release document.

### DIFF
--- a/.github/workflows/release-layer-collector.yml
+++ b/.github/workflows/release-layer-collector.yml
@@ -49,20 +49,20 @@ jobs:
           - amd64
           - arm64
         aws_region:
-         - ap-northeast-1
-         - ap-northeast-2
-         - ap-south-1
-         - ap-southeast-1
-         - ap-southeast-2
-         - ca-central-1
-         - eu-central-1
-         - eu-north-1
-         - eu-west-1
-         - eu-west-2
-         - eu-west-3
-         - sa-east-1
-         - us-east-1
-         - us-east-2
+          - ap-northeast-1
+          - ap-northeast-2
+          - ap-south-1
+          - ap-southeast-1
+          - ap-southeast-2
+          - ca-central-1
+          - eu-central-1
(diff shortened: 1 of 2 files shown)


.github/workflows/release-layer-collector.yml
```diff
@@ -49,27 +49,27 @@ jobs:
...
+         - ap-southeast-2
+         - ca-central-1
+         - eu-central-1
+         - eu-north-1
+         - eu-west-1
+         - eu-west-2
+         - eu-west-3
+         - sa-east-1
+         - us-east-1
+         - us-east-2
```
tylerbenson: Indentation inconsistency with below.

RELEASE.md
```diff
@@ -0,0 +1,7 @@
+# OpenTelemetry Lambda Layer Release Procedure
+
+The release process is almost entirely managed by [GitHub actions](https://github.com/open-telemetry/opentelemetry-lambda/tree/main/.github/workflows). To publish a new layer:
+
+1. Create a new tag for the layer to publish. For example, to create a new collector layer, the following command is used:
+        `git tag layer-collector/0.0.8`
+2. Push the tag to [opentelemetry-lambda](https://github.com/open-telemetry/opentelemetry-lambda) repository to trigger the publish action.
```
tylerbenson, line 7: Should we encourage creating a github release with the proper ARN's for each region?  (We should probably automate this somehow.)

tylerbenson (approved):
Approved assuming indentation is fixed.

.github/workflows/release-layer-collector.yml
codeboten: Fixed, thanks for catching!

RELEASE.md
codeboten: Yeah i wasn't clear what the process should be. If we define it we can add it in the doc in a follow up

PR #730, opened 2023-05-31
update python build to use src instead of sample-apps

ocelotl:
Remove the collector as part of the build as well.


codeboten (approved):

ocelotl:
@codeboten it is not working fix

PR #736, opened 2023-06-02
Add sfx exporter to collector lambda layer components

ashish-chide:


tylerbenson:
Can you elaborate why you need this?  I assume Signalfx/Splunk works with OTLP so I'm inclined to close this PR.

PR #849, opened 2023-08-15
[chore] bump collector deps to v0.83.0

codeboten:


collector/receiver/telemetryapireceiver/go.mod
```diff
@@ -6,14 +6,14 @@ replace github.com/open-telemetry/opentelemetry-lambda/collector => ../../
 
 require (
 	github.com/golang-collections/go-datastructures v0.0.0-20150211160725-59788d5eb259
-	github.com/open-telemetry/opentelemetry-lambda/collector v0.81.0
+	github.com/open-telemetry/opentelemetry-lambda/collector v0.83.0
```
Aneurysm9: Should this be bumped later when a tag exists for this module at this version?

codeboten: This is overwritten by a replace anyways, but i don't feel strongly one way or the other.

This question makes me realize that we don't currently publish any tags for this package :(

tylerbenson (approved):

collector/receiver/telemetryapireceiver/go.mod
Aneurysm9: Can we leave it unchanged for now, then, and bump it at the next release when tags are created?

codeboten: Done, i'll create a ticket to ensure module tags are published

Aneurysm9 (approved):

Issue #867, opened 2023-08-23
Add CodeQL for csharp


codeboten:
Related to #862

rapphil:
I ended up working on this one as well. :)

codeboten:
Fixed by https://github.com/open-telemetry/opentelemetry-lambda/pull/876

PR #876, opened 2023-08-29
Feat: Add support to csharp CodeQL analysis

rapphil:
Solves https://github.com/open-telemetry/opentelemetry-lambda/issues/867

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -114,7 +114,8 @@ jobs:
       if: ${{ matrix.target.language == 'csharp' }}
 
     - name: build dotnet
-      run: dotnet build
+      # Build all dotnet applications from this directory
+      run: find . -name '*.sln' -exec dotnet build '{}' \;
       working-directory: ${{ matrix.target.directory }}
       if: ${{ matrix.target.language == 'csharp' }}
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -108,7 +108,7 @@ jobs:
       if: ${{ matrix.target.language == 'java' }}
 
     - name: setup dotnet
-      uses: actions/setup-dotnet@3
+      uses: actions/setup-dotnet@v3
       with:
         dotnet-version: 6.x
       if: ${{ matrix.target.language == 'csharp' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -65,7 +78,10 @@ jobs:
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
       uses: github/codeql-action/autobuild@v2
-      if: ${{ matrix.language != 'java' }}
+      with:
+        working-directory: ${{ matrix.target.directory }}
```
rapphil, line 82: this is to make sure that autobuild will start in the correct subdirectory to try to build the code.

tylerbenson (approved):

.github/workflows/codeql.yml
codeboten: Is there an instance where that's not the case?

.github/workflows/codeql.yml
```diff
@@ -82,16 +98,28 @@ jobs:
...
+    - name: setup dotnet
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: 6.x
+      if: ${{ matrix.target.language == 'csharp' }}
+
+    - name: build dotnet
+      # Build all dotnet applications from this directory
+      run: find . -name '*.sln' -exec dotnet build '{}' \;
+      working-directory: ${{ matrix.target.directory }}
```
codeboten, line 119: Is this still needed w/ line 82?

codeboten (commented):
Thanks @rapphil, just a couple of questions

.github/workflows/codeql.yml
rapphil: [In a previous run when autobuild was executed from the root of the repository](https://github.com/open-telemetry/opentelemetry-lambda/actions/runs/5930233093/job/16079525511#step:4:50), the autobuild action tried to run the build using an unrelated python script for the case it was running for the csharp language. Explicitly setting the working directory will ensure that autobuild will only find build scripts related to the language that is set in line 68.

.github/workflows/codeql.yml
rapphil: this could be removed. I preferred to leave it here just to be consistent with the other languages. (the find command will be executed from the dotnet directory).

codeboten (approved):

PR #908, opened 2023-09-21
Improve Python tests workflow

rapphil:
* Create requirements file for dependencies used in tests
* Update tox file to use dependencies file with stable dependencies versions

This has to be done because currently it is trying to use whatever is set in the main branch, which is problematic. This will add more stability and reproducibility to the tests.


rapphil:
After this merge, this dependabot PR should work: https://github.com/open-telemetry/opentelemetry-lambda/pull/902

codeboten (approved):

PR #959, opened 2023-10-20
Add Decouple and Batch Processors to Collector

adcharre:
# Problem
The introduction of force flush before a Lambda function finishes causes an increase in responses times due to the need to wait for the collector pipelines to complete before the lambda function returns.

# Solution
By adding a new processor that decouples the receiver and exporter sides of the pipeline and is aware of lambda lifecycle events, the response time of the function is unaffected by the time it takes to export opentelemetry data. 

In addition, adding the Batch processor to the list of available processors can reduce the cost of lambda function invocation by not needing to send opentelemetry data on every run. This is at the expense of data being delayed.


collector/processor/decoupleprocessor/README.md
```diff
@@ -0,0 +1,25 @@
+# Decouple Processor
+
+| Status                   |                       |
+| ------------------------ |-----------------------|
+| Stability                | [in development]      |
+| Supported pipeline types | traces, metrics, logs |
+| Distributions            | [extension]           |
+
+This processor decouples the receiver and exporter ends of the pipeline allowing the lambda function to finish before traces/metrics/logs are exported by the collector. The processor is aware of the Lambda lifecycle and will prevent the environment from being frozen or shutdown until any pending traces/metrics/logs have been exported.
```
tylerbenson: I suggest including a link to the [lifecycle doc](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#runtimes-extensions-api-lifecycle). (I assume the `shutdown` hook is what you're taking advantage of here.)

tylerbenson:
@adcharre please go through the CLA process.

collector/processor/decoupleprocessor/README.md
adcharre: Yep, I've added the link as suggested.

adcharre:
> @adcharre please go through the CLA process.

Trying to organise that at the moment - unfortunately my company hasn't yet signed the CLA so it's taking a bit of probing internally to get this done.

collector/processor/decoupleprocessor/README.md
```diff
@@ -0,0 +1,25 @@
...
+In this way the response times of the Lambda function is not impacted by the need to export data, however the billed duration will include the time taken to export data as well as runtime of the lambda function. 
+
+When combined with the batch processor, the number of exports required can be significantly reduced and therefore the cost of running the lambda. This is with the trade-off that the data will not be available at your chosen endpoint until some time after the invocation, up to a maximum of 5 minutes (the timeout that the environment is shutdown when no further invocations are received).
```
codeboten, line 14: I wonder if the `decouple processor` would be better written as a lifecycle-aware batch processor. I worry that without this, data would likely end up sitting in the existing batch processor whether it was placed before or after the decouple processor.

Another question that comes to mind is whether the batch processor alone would work here since it would in theory be able to send all the data left in any existing batches through the normal shutdown process.

codeboten (commented):
Thanks for the PR @adcharre! I'm curious if the batch processor alone would work if the collector is able to respond to shutdown events from the environment.

I remember a few years ago, the batch processor was not useful, since batches would get stuck and never sent when lambdas were frozen.

collector/processor/decoupleprocessor/README.md
adcharre: > I wonder if the decouple processor would be better written as a lifecycle-aware batch processor. I worry that without this, data would likely end up sitting in the existing batch processor whether it was placed before or after the decouple processor.

Actually I was going to update the README to indicate that the batch processor must come before the decouple processor otherwise you run into problems (as mentioned below). It would be possible to rewrite this as you say, however the simplest approach was to do one thing and one thing well and re-use the existing batch processor in front of the decouple processor. 
Also it's reasonable to just use the decouple processor on it's own to reduce the response time of the lambda while not delaying the delivery of otel data.

> Another question that comes to mind is whether the batch processor alone would work here since it would in theory be able to send all the data left in any existing batches through the normal shutdown process.

Not on it's on no, I did also consider it. The problem is that the collector is only shutdown when the environment is about to be destroyed. However it's frozen after each function invocation.
Without the ability to prevent the environment being frozen until the exports have successfully completed you will end up with data occasionally being lost due network interruptions.

tylerbenson: @adcharre I'm not very familiar with extensions (or the collector)... could you highlight for me how the code you added handles the frozen scenario? (How does it prevent from being frozen.) Perhaps some additional comments in your code would be helpful here.

collector/internal/lifecycle/manager.go
```diff
@@ -142,10 +145,37 @@ func (lm *manager) processEvents(ctx context.Context) error {
...
 			}
 
+			lm.notifyFunctionInvoked()
+
 			err = lm.listener.Wait(ctx, res.RequestID)
 			if err != nil {
 				lm.logger.Error("problem waiting for platform.runtimeDone event", zap.Error(err), zap.String("requestID", res.RequestID))
 			}
+
+			// Check other components are ready before allowing the freezing of the environment.
```
adcharre, line 155: @tylerbenson  - here's a comment to explain that we notifier other components that the function has finished and therefore need to be ready to freeze.

The new code I've added hooks into the existing lifecycle code,

collector/lambdalifecycle/notifier.go
```diff
@@ -0,0 +1,37 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package lambdalifecycle
+
+type Listener interface {
+	FunctionInvoked()
+	FunctionFinished()
```
adcharre, line 23: Probably this is where some additional comments are required, I'll look at adding them tomorrow.

adcharre:
@codeboten & @tylerbenson - finally through the CLA and I've updated the readme and code with some more comments around lifecyle.

collector/internal/lifecycle/manager.go
tylerbenson: I don't have a good understanding of the extension lifecycle.  If the lambda function returns and starts waiting for the next event, are you sure it waits for events to finish processing before freezing? Is this documented somewhere? How did you verify this behavior?

tsloughter: Lifecycle docs say it waits until the extension also calls `Next`.

But does the lambda function get the next event only after all extensions have called `Next`? If that were the case then wouldn't it still be waiting until the extension finishes, unless the extension calls `Next` before doing the export, but then it might be frozen before actually exporting.

adcharre: @tylerbenson according to the lifecycle documentation (https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#runtimes-extensions-api-lifecycle)

> Each phase starts with an event from Lambda to the runtime and to all registered extensions. The runtime and each extension signal completion by sending a Next API request. Lambda freezes the execution environment when each process has completed and there are no pending events.

@tsloughter the lambda runtime will only get the next invocation once all extensions have also signalled they are ready. 
The sequence of events is:
1. Lambda and Extensions receive `Next` event
2. Lambda function completes and signals it's ready for the next event.
3. Collector receives notification from Telemetry API that the function has finished.
4. Decouple processor finishes forwarding data.
5. Collector signals it's ready for the next event.
6. Back to 1.
 
Also see the "Invoke phase" in the same link:

> After receiving the function response from the runtime, Lambda returns the response to the client, even if extensions are still running.

and 

> The Invoke phase ends after the runtime and all extensions signal that they are done by sending a Next API request.

As to the comment about verifying this behaviour - from Cloudwatch logs to confirm what's documented.

collector/processor/decoupleprocessor/README.md
adcharre: I've added additional information to the Listener interface here https://github.com/adcharre/opentelemetry-lambda/blob/2a8812c4aa78cbe7a27149532ec78ce6081a7234/collector/lambdalifecycle/notifier.go#L21

collector/internal/lifecycle/manager.go
tylerbenson: Thanks for the detailed response.

tsloughter: Ok, cool, this is great.

codeboten (approved):
Thanks for the submission @adcharre, would you consider submitting the decouple processor to the collector contrib repo? minor change i would suggest is calling it the queueingprocessor

specifically i think it would be beneficial to have a processor that can decouple the export pipeline from the batch processor into queues. many collector exporters do this today via the queueing supported in the exporter helper. a processor would take that responsibility out of the exporter altogether

codeboten:
Please take a look at the failing test:

```
--- FAIL: TestLifecycle (0.40s)
    --- FAIL: TestLifecycle/full_lifecycle_with_data_from_function (0.20s)
        testing.go:1465: race detected during execution of test
    --- FAIL: TestLifecycle/full_lifecycle_with_data_before_shutdown (0.20s)
        testing.go:1465: race detected during execution of test
    testing.go:1465: race detected during execution of test
```

adcharre:
> Please take a look at the failing test:
> 
> ```
> --- FAIL: TestLifecycle (0.40s)
>     --- FAIL: TestLifecycle/full_lifecycle_with_data_from_function (0.20s)
>         testing.go:1465: race detected during execution of test
>     --- FAIL: TestLifecycle/full_lifecycle_with_data_before_shutdown (0.20s)
>         testing.go:1465: race detected during execution of test
>     testing.go:1465: race detected during execution of test
> ```

I've reworked the 2 tests and tested locally and no more errors from the race detector.

adcharre:
> Thanks for the submission @adcharre, would you consider submitting the decouple processor to the collector contrib repo? minor change i would suggest is calling it the queueingprocessor
> 
> specifically i think it would be beneficial to have a processor that can decouple the export pipeline from the batch processor into queues. many collector exporters do this today via the queueing supported in the exporter helper. a processor would take that responsibility out of the exporter altogether

I'll have a go if you think it'll be useful. I had assumed there would be no need as the batch processor can do the breaking the receive -> export pipeline. It should just be a matter of removing some code and renaming a few things...

tylerbenson:
@adcharre If you'd like to apply @codeboten's suggested naming change of `queueingprocessor` to this PR, we can move forward with merging. This way in the future when that processor is in the collector repo we can migrate without breaking users.

Thanks for your effort here.  This is great!  I really appreciate the explanations you've provided.

adcharre:
> @adcharre If you'd like to apply @codeboten's suggested naming change of `queueingprocessor` to this PR, we can move forward with merging. This way in the future when that processor is in the collector repo we can migrate without breaking users.
> 
> Thanks for your effort here. This is great! I really appreciate the explanations you've provided.

@tylerbenson I don't think that makes sense as there will need to be 2 separate plugins! 
This plugin is lambda lifecycle aware and makes use of packages that are not available in the main collector. The main collector plugin would just be the simple queuing mechanism to break the link between receiver and exporter.

Stepping back a bit, there _could_ be 2 separate implementations of the `queueingprocessor` one here with the lambda lifecycle awareness and a separate implementation in opentelemetry-collector-contrib which doesn't include the lambda stuff?

codeboten:
Right, the proposal to submit the queueing portion of the processor (`queueingprocessor`) as its own processor was specifically around giving the processing pipeline the ability to decouple the exporting from the batch processor (as I mentioned earlier this is largely accomplished via queueing in the exporter helper today)

If in the future there was a facility for registering the lambda lifecycle mechanism in the `queueingprocessor` (which could be doable if it were done generically enough, not sure if theres a use-case beyond lambda), then the `decoupleprocessor` could be deprecated in favour of this `queueingprocessor`. 

In the short term though, i think it's fine for them to be separate processors.

tylerbenson:
Ok, sorry for the confusion.  Please resolve the conflicts from main then I can merge.

tylerbenson (approved):

codeboten:
@adcharre please ignore my comment suggesting opening an issue/pr to the main collector repository. I was thankfully reminded that this used to exist in the core repository as the queuedprocessor and was deprecated in November 2021 😄 https://github.com/open-telemetry/opentelemetry-collector/commit/e82037015b83c094f339cc108b3554a4f8d1e918

adcharre:
@tylerbenson  conflicts should now be resolved and I've updated the decoupleprocessors dependencies to be inline with the other processors.

tylerbenson:
@adcharre one thing I forgot to add here... Please update the [readme](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/README.md) with updated guidelines for using this new feature.

adcharre:
> @adcharre one thing I forgot to add here... Please update the [readme](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/README.md) with updated guidelines for using this new feature.

@tylerbenson - Will do! I'll try and get round to it next week.

PR #1035, opened 2023-12-07
Display Python SDK version

### LINKED ISSUE #1034: Python component-version is not displayed in logs
It should appear [here](https://github.com/open-telemetry/opentelemetry-lambda/actions/runs/7133459715/job/19426262691#step:1:30).

ocelotl:
Fixes #1034

### DIFF
--- a/.github/workflows/release-layer-python.yml
+++ b/.github/workflows/release-layer-python.yml
@@ -14,7 +14,7 @@ jobs:
   build-layer:
     runs-on: ubuntu-latest
     outputs:
-      PYTHON_OPENTELEMETRY_SDK_VERSION: ${{ steps.save-python-sdk-version.outputs.PYTHON_OPENTELEMETRY_SDK_VERSION}}
+      PYTHON_OPENTELEMETRY_SDK_VERSION: ${{ steps.save-python-opentelemetry-sdk-version.outputs.PYTHON_OPENTELEMETRY_SDK_VERSION}}
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


python/src/otel/otel_sdk/requirements.txt
```diff
@@ -1,3 +1,4 @@
+opentelemetry-sdk==1.21.0
```
ocelotl, line 1: Dependabot should update this with every new release.

.github/workflows/release-layer-python.yml
```diff
@@ -14,30 +14,40 @@ jobs:
...
+        shell: bash
+
+      - name: Display PYTHON_OPENTELEMETRY_SDK_VERSION
+        run: echo "PYTHON_OPENTELEMETRY_SDK_VERSION=${{ env.PYTHON_OPENTELEMETRY_SDK_VERSION }}"
+        shell: bash
+
+      - name: Save Python SDK Version
+        id: save-python-sdk-version
+        run: |
+          echo "PYTHON_OPENTELEMETRY_SDK_VERSION=$PYTHON_OPENTELEMETRY_SDK_VERSION" >> $GITHUB_OUTPUT
```
tylerbenson: Can these three steps be combined into one?  I think I'd prefer that.

python/src/otel/otel_sdk/requirements-nodeps.txt
```diff
@@ -27,4 +27,3 @@ opentelemetry-instrumentation-sqlite3==0.42b0
 opentelemetry-instrumentation-starlette==0.42b0
 opentelemetry-instrumentation-tornado==0.42b0
 opentelemetry-instrumentation-wsgi==0.42b0
-
```
tylerbenson, line 30: What about the `0.42b0` portion of the version... does that matter?  What is the difference?

.github/workflows/release-layer-python.yml
ocelotl: I merged 2 steps, the displaying of that environment variable needs to be in a separate step, if not its value is not shown in the logs.

tylerbenson (approved):

python/src/otel/otel_sdk/requirements-nodeps.txt
ocelotl: I'll use only `1.21.0` as the version of the layer, we also use `0.42b0` for some packages that are not stable yet, but `1.21.0` is a sufficient id for a release.

PR #1088, opened 2024-01-10
feat(python): add custom ctx extractor for sqs)

mat-rumian:
PR adds possibility to extract context `traceparent/tracestate/baggage` from SQS Message attribute. This solves issue related to lack of context propagation between `lambd` -> `SQS` -> `lambda`. Now everything should be presented as single trace.


python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -48,8 +53,35 @@ def modify_module_name(module_name):
...
+        if 'traceparent' in records:
+            sqs_ctx['traceparent'] = records['traceparent']['stringValue']
+        if 'tracestate' in records:
+            sqs_ctx['tracestate'] = records['tracestate']['stringValue']
+        if 'baggage' in records:
+            sqs_ctx['baggage'] = records['baggage']['stringValue']
+    except (TypeError, KeyError):
+        logger.debug("Extracting context from Lambda Event records failed.")
+
+    if sqs_ctx is not None and 'traceparent' in sqs_ctx:
```
ocelotl:
```suggestion
    if 'traceparent' in sqs_ctx:
```

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -48,8 +53,35 @@ def modify_module_name(module_name):
 class HandlerError(Exception):
     pass
 
-
-AwsLambdaInstrumentor().instrument()
+def _headers_and_sqs_context_extractor(lambda_event: Any) -> Context:
+    headers = None
```
ocelotl:
```suggestion
    headers = {}
```

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -48,8 +53,35 @@ def modify_module_name(module_name):
...
+            sqs_ctx['tracestate'] = records['tracestate']['stringValue']
+        if 'baggage' in records:
+            sqs_ctx['baggage'] = records['baggage']['stringValue']
+    except (TypeError, KeyError):
+        logger.debug("Extracting context from Lambda Event records failed.")
+
+    if sqs_ctx is not None and 'traceparent' in sqs_ctx:
+        return get_global_textmap().extract(sqs_ctx)
+           
+    if not isinstance(headers, dict):
```
ocelotl:
```suggestion
```

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -48,8 +53,35 @@ def modify_module_name(module_name):
...
+        headers = {}
```
ocelotl:
```suggestion
```

ocelotl (approved):
I think some code can be fixed a bit

mat-rumian:
> I think some code can be fixed a bit

yeah, thanks for pointing :)

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -48,8 +53,33 @@ def modify_module_name(module_name):
...
+    try:
+        headers = lambda_event["headers"]
+    except (TypeError, KeyError):
+        logger.debug(
+            "Extracting context from Lambda Event for headers failed."
+        )
+
+    sqs_ctx = {}
+    try:
+        records = lambda_event["Records"][0]['messageAttributes']
```
tylerbenson, line 67: I'm not familiar with the python implementation but I assume it doesn't align with [the spec](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/aws-lambda.md#sqs) and this PR seems to make it diverge even further. Specifically, it only captures headers from the first message and doesn't add span links.

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -48,8 +53,33 @@ def modify_module_name(module_name):
...
+        if 'traceparent' in records:
+            sqs_ctx['traceparent'] = records['traceparent']['stringValue']
+        if 'tracestate' in records:
+            sqs_ctx['tracestate'] = records['tracestate']['stringValue']
+        if 'baggage' in records:
+            sqs_ctx['baggage'] = records['baggage']['stringValue']
```
tylerbenson, line 73: This assumes the w3c propagator is configured. Can you convert the whole `records` object to something that can generically be passed into the extract so it can work with whichever propagator is configured?

python/src/otel/otel_sdk/otel_wrapper.py
```diff
@@ -48,8 +53,33 @@ def modify_module_name(module_name):
 class HandlerError(Exception):
     pass
 
-
-AwsLambdaInstrumentor().instrument()
+def _headers_and_sqs_context_extractor(lambda_event: Any) -> Context:
```
tylerbenson, line 56: To me it seems this logic should belong in the `AwsLambdaInstrumentor` instead of splitting it into separate repos?

tylerbenson (changes requested):

python/src/otel/otel_sdk/otel_wrapper.py
mat-rumian: Yeah, you're right. Instrumentation should be the best place to have it. I will apply changes in the instrumentation :) Thanks

mat-rumian:
SQS context extraction should be applied on the instrumentation side.

PR #1117, opened 2024-01-23
Fix: Only include layer in SAM template for Python src  directory

rapphil:
Only include layer in SAM template for Python src  directory so that the [pr build does not fail ](https://github.com/open-telemetry/opentelemetry-lambda/actions/runs/7629607482/job/20783370755?pr=1076#step:6:195)trying to find some source code. It seems that the template copied at some point from the sample lambda.


tylerbenson (approved):
Any way to build the layer without SAM or the AWS cli like we do for other languages?

PR #1118, opened 2024-01-25
feat: switch from node 14 to 16

mat-rumian:
PR is switching from `node v14` to `node v16` according to [supported runtimes in AWS](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html). PR also fixes linter usage.

### DIFF
--- a/.github/workflows/release-layer-nodejs.yml
+++ b/.github/workflows/release-layer-nodejs.yml
@@ -71,7 +71,7 @@ jobs:
       layer-name: opentelemetry-nodejs
       component-version: ${{needs.build-layer.outputs.NODEJS_VERSION}}
       # architecture:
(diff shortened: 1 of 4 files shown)


tylerbenson (approved):

.github/workflows/release-layer-nodejs.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 14
+          node-version: 16
 
       - name: Build
         run: |
```
tylerbenson, line 26: Do you want to add Node 20 to the runtimes list on line 74?

mat-rumian: Yeah, thanks for catching it :) Added

tylerbenson (approved):

PR #1172, opened 2024-02-29
feat: allow to configure AwsInstrumentation

### LINKED ISSUE #1171: How to disable AwsInstrumentation?
While wanting to use open telemetry on aws lambda I don't want aws sdk spans and metrics, I would like to have only my custom (business) metrics sent to the server.

As of now I see the wrapper is hardcoded to include the AwsInstrumentation always, but I wonder if there is a way to disable that without opening a PR

onhate:
fix #1171


tylerbenson:
As I mentioned in #1171 I think I would rather encourage users to remove the lambda layer instead.  WDYT @rapphil?

tylerbenson:
Closing this for now.  Can revisit if we decide to add this consistently to all languages.

PR #1180, opened 2024-03-04
Add fast startup option for Java agent by enabling JIT tiered compilation to stop at level 1 …

serkan-ozal:
… and disabling bytecode verification.

Fast startup mode is disabled by default but can be enabled by setting `OTEL_JAVA_AGENT_FAST_STARTUP_ENABLED`
 environment variable to `true`.

For more information about the idea behind this optimization, you can check the following resources:
- https://aws.amazon.com/tr/blogs/compute/optimizing-aws-lambda-function-performance-for-java/
- https://aws.amazon.com/tr/blogs/compute/increasing-performance-of-java-aws-lambda-functions-using-tiered-compilation/

Here are some coldstart (init duration) numbers (I have tested with simple hello world Lambda function which does nothing other than printing hello world):

## Memory Size: 3072 MB
| Configuration                                       | Init Duration (ms)   |
| ------------------------------------ | ------------------: |
| No Agent                                              |                          379 |
| With Agent                                           |                       6384  |
| With Agent (Fast startup enabled)    |                       2590  |

When fast startup enabled, Lambda function starts **2.46x** faster.

## Memory Size: 2048 MB
| Configuration                                       | Init Duration (ms)   |
| ------------------------------------ | ------------------: |
| No Agent                                              |                          448 |
| With Agent                                           |                       8819  |
| With Agent (Fast startup enabled)    |                       2907  |

When fast startup enabled, Lambda function starts **3.03x** faster.

## Memory Size: 1536 MB
| Configuration                                       | Init Duration (ms)   |
| ------------------------------------ | ------------------: |
| No Agent                                              |                          500 |
| With Agent                                           |  > 10000 (timeout) |
| With Agent (Fast startup enabled)    |                       3492  |

## Memory Size: 1024 MB
| Configuration                                       | Init Duration (ms)   |
| ------------------------------------ | ------------------: |
| No Agent                                              |                          503 |
| With Agent                                           |  > 10000 (timeout) |
| With Agent (Fast startup enabled)    |                       3500  |

### DIFF
--- a/java/README.md
+++ b/java/README.md
@@ -21,6 +21,23 @@ Note, automatic instrumentation has a notable impact on startup time on AWS Lamb
 generally need to use this along with provisioned concurrency and warmup requests to serve production
 requests without causing timeouts on initial requests while it initializes.
 
+#### Fast startup for Java agent
+
+Fast startup mode is disabled by default but can be enabled by specifying the `OTEL_JAVA_AGENT_FAST_STARTUP_ENABLED=true`
+in your Lambda configuration.
+
+When fast startup mode is enabled, **JIT** (Just-In-Time) **Tiered compilation** is configured to stop at level 1 
+and bytecode verification is disabled. So, the JVM uses the **C1** compiler which is optimized for fast start-up time. 
+This compiler (**C1**) quickly produces optimized native code 
+but it does not generate any profiling data and never uses the **C2** compiler 
+which optimized for the best overall performance but uses more memory and takes a longer time to achieve it.
+Therefore, this option is not enabled by default and needs to be enabled by the user explicitly 
+by taking care of the behavioural change mentioned above.
+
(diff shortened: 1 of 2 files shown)


tylerbenson (approved):
Thanks for the contribution.  This definitely looks like an interesting option for lambdas that are short lived.

tylerbenson:
One request... Would you please update [the readme](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/java/README.md) with some documentation so this new setting is discoverable and users can understand what it does?

serkan-ozal:
@tylerbenson Just updated the README with this new option.

serkan-ozal:
Thanks for merging @tylerbenson.
When are you planning to release new Java agent layer with this addition?

tylerbenson:
We are overdue... I can try to do a release this week.

PR #1225, opened 2024-03-25
bump protobuf,zap,testify in /collector

nslaughter:
These updates align the `./collector` modules with dependency versions now on main in the collector-contrib repo. This includes changes in the following dependabot PRs:
* #1200 
* #1202 
* #1203 
* #1204 

For the future, it makes sense to reconsider groupings and automation that will group more related changes and make each more likely to merge.


tylerbenson:
I think this might be unnecessary now.  For example, check out #1288.

tylerbenson:
I think this is no longer needed.  Reopen if I'm missing something.

PR #1254, opened 2024-04-15
Update typescript-eslint version to latest

martinkuba:
This is an attempt to address the dependabot update in https://github.com/open-telemetry/opentelemetry-lambda/pull/1246, which is bumping only one package.

### DIFF
--- a/nodejs/package.json
+++ b/nodejs/package.json
@@ -30,8 +30,8 @@
   "author": "OpenTelemetry Authors",
   "license": "Apache-2.0",
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


tylerbenson:
Grouping all the updates caused a failure #1259

martinkuba:
Looks like `@typescript/eslint-parser` is not compatible with `eslint@9`, so need to hold off with eslint update for now.

tylerbenson:
Should we pin the version range to 8.x?

tylerbenson (approved):

rapphil (approved):

PR #1255, opened 2024-04-16
Add decoupleafterbatch converter to ensure decouple processor follows batch processor

nslaughter:
## Problem
When using the batch processor in a Lambda Collector pipeline, it's important to always have the decouple processor come after the batch processor. This is because the batch processor introduces delays in processing some spans/metrics/logs in order to batch them with later arriving items. If the batch processor's effect isn't offset by the decouple processor then the delay can cause issues when the Lambda environment gets frozen, potentially leading to loss of data and higher costs. In short, the current situation places a burden on the user to apply knowledge specialized to the Lambda distribution to realize the results they're most likely expecting.

## Solution
This PR introduces a new `decoupleafterbatch` converter which automatically modifies the config, potentially adding a decouple processor to the processor chains as defined in the user's config. The behavior is that checks if a batch processor is defined and if there is no decouple processor following it, then it adds it to the end.

The key aspects are:

1. When a user configures a batch processor without following it with a batch processor, the converter will automatically add a decouple processor to the end of the pipeline.

2. If a decouple processor is already present after the last occurrence of the batch processor then nothing happens.

3. If batch processors are defined in multiple pipelines (logs, traces, etc), each one will be appended according to the state of that processor pipeline only.

## Rationale
The decouple processor allows the lambda function to return while the Collector continues exporting the observability data asynchronously. By ensuring it always comes after the batch processor, we ensure that users avoid potential data loss scenarios, delays, and unnecessary expense caused when the Lambda environment gets frozen in the middle of a batch processor timeout.

## Alternatives Considered

1. **Document the requirement and rely on users to always add the decouple processor.** This is likely to be overlooked and places a burden on the user to get better results.

2. **Disable the batch processor when running in Lambda.** This would ensure preventing bad performance caused by the batch processor in lambda, but it would also prevent users from using the batch processor's capabilities.

The automated config by converters approach is simplest for users and the most robust. It restores the benefit of the batch processor, which is recommended in OpenTelemetry best practices, while ensuring that users don't run into unexpected problems with the unique behaviors in the lambda environment.

## Testing

Unit tests have been added to verify the converter works as expected, both when a user hasn't defined any processors as well as when they've defined just a batch processor, just a decouple processor, or both in different valid and invalid orders.

The rest is in Github Actions.

Please take a look and let me know if you have any feedback or suggestions!


collector/internal/confmap/converter/decoupleafterbatchconverter/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The `DecoupleAfterBatch` converter automatically modifies the collector's configuration for the Lambda distribution. Its purpose is to ensure that a decouple processor is always present after a batch processor in a pipeline, in order to prevent potential data loss due to the Lambda environment being frozen.
+
+## Behavior
+
+The converter scans the collector's configuration and makes the following adjustments:
+
+1. If a pipeline contains a batch processor with no decouple processor defined after it, the converter will automatically add a decouple processor to the pipeline immediately after the batch processor.
+
+2. If a pipeline contains a batch processor with a decouple processor already defined after it or there is no batch processor defined, the converter will not make any changes to the pipeline configuration.
```
adcharre, line 11: What's the expected behaviour if the user gets it wrong and adds a decouple processor before a batch processor?
i.e.
```
input -> decouple -> batch -> output
```

collector/internal/confmap/converter/decoupleafterbatchconverter/converter.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	return nil
+}
+
+// The appendDecouple is the filter predicate for the Convert function action. It tells whether
+// (bool) there was a decouple processor after the last
+// batch processor, which Convert uses to decide whether to append the decouple processor.
+func appendDecouple(processors []interface{}) bool {
```
adcharre: The name `appendDecouple` suggests that it's doing the actual appending, I'd suggest changing the name to something like `shouldAppendDecouple`

collector/processor/decoupleprocessor/README.md
```diff
@@ -7,22 +7,33 @@
...
 
-When combined with the batch processor, the number of exports required can be significantly reduced and therefore the cost of running the lambda. This is with the trade-off that the data will not be available at your chosen endpoint until some time after the invocation, up to a maximum of 5 minutes (the timeout that the environment is shutdown when no further invocations are received). 
+When combined with the batch processor, the number of exports required can be significantly reduced and therefore the cost of running the lambda. This is with the trade-off that the data will not be available at your chosen endpoint until some time after the invocation, up to a maximum of 5 minutes (the timeout that the environment is shutdown when no further invocations are received).
+
+## Auto-Configuration
+
+Due to the significance of performance improvement with this approach, the OpenTelemetry Lambda Layer automatically configures the decouple processor when the batch processor is used. This provides the best performance by default.
```
adcharre: Minor suggestion to make this easier to read
```suggestion
Due to the significant performance improvements with this approach, the OpenTelemetry Lambda Layer automatically configures the decouple processor when the batch processor is used. This ensures the best performance by default.
```

collector/internal/confmap/converter/decoupleafterbatchconverter/converter_test.go
```diff
@@ -0,0 +1,114 @@
...
+    }{
+		// This test is first, because it illustrates the difference in making the rule that when
+		// batch is present the converter appends decouple processor to the end of chain versus
+		// the approach of this code which is to do this only when the last instance of batch
+		// is not followed by decouple processor.
+		{
+			name: 	   "batch then decouple in middle of chain",
+			processors: []interface{}{"processor1", "batch", "decouple", "processor2"},
+			expectedProcessors: []interface{}{"processor1", "batch", "decouple", "processor2"},
+		},
```
adcharre: Nit. Formatting seems a bit off here, I think you might have a combination of tabs and space formatting.

collector/internal/confmap/converter/decoupleafterbatchconverter/converter_test.go
```diff
@@ -0,0 +1,114 @@
...
+        {
+            name:       "no pipelines",
+            processors: nil,
+            expectedProcessors: nil,
+        },
+        {
+            name:       "no processors in chain",
+            processors: nil,
+            expectedProcessors: nil,
+        },
```
adcharre: How are these test cases different as they all pass in the same nil value?

collector/internal/confmap/converter/decoupleafterbatchconverter/converter_test.go
```diff
@@ -0,0 +1,114 @@
...
+                        "processors": processors,
+                    },
+                },
+            },
+        })
+    }
+
+    testCases := []struct {
+        name        string
+        processors  []interface{}
```
adcharre: I think you actually want these to be confmap.Conf values which you populate using the `baseConf` call. 
That way you can test the cases on lines 55-64 where you actual want a confmap.Conf structure without services/pipelines etc

collector/internal/confmap/converter/decoupleafterbatchconverter/converter_test.go
```diff
@@ -0,0 +1,114 @@
...
+            processors: []interface{}{"processor1", "batch", "processor2"},
+            expectedProcessors: []interface{}{"processor1", "batch", "processor2", "decouple"},
+        },
+
+        {
+            name:       "batch and decouple processors already present in correct position",
+            processors: []interface{}{"processor1", "batch", "processor2", "decouple"},
+            expectedProcessors: []interface{}{"processor1", "batch", "processor2", "decouple"},
+        },
+    }
```
adcharre: Missing test case for decouple then batch processor.

collector/go.mod
```diff
@@ -69,6 +69,7 @@ require (
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
+	github.com/google/go-cmp v0.6.0 // indirect
```
adcharre: Should there also be an update to `collector/go.sum`?

collector/internal/confmap/converter/decoupleafterbatchconverter/README.md
nslaughter: > What's the expected behaviour if the user gets it wrong and adds a decouple processor before a batch processor? i.e.
> 
> ```
> input -> decouple -> batch -> output
> ```

Good question. To describe the processor chain rewrite rule in pseudo-regex I'd say `decouple -> .* -> batch -> [^decouple]*` becomes `decouple -> .* -> batch -> [^decouple]* -> decouple`. I've added 3 test cases to clarify the results (including this example) when the chain fits this pattern.

collector/internal/confmap/converter/decoupleafterbatchconverter/converter.go
nslaughter: [x] updated code

collector/internal/confmap/converter/decoupleafterbatchconverter/converter_test.go
nslaughter: Oh, wow. Surprised that `gofmt` wasn't in my editing pipeline.

collector/internal/confmap/converter/decoupleafterbatchconverter/converter_test.go
nslaughter: Thanks for pointing that out.

collector/internal/confmap/converter/decoupleafterbatchconverter/converter_test.go
nslaughter: Thanks. I added a few that deal with this.

tylerbenson:
Thanks for the thorough review @adcharre!

collector/internal/confmap/converter/decoupleafterbatchconverter/README.md
```diff
@@ -0,0 +1,11 @@
+# DecoupleAfterBatch Converter
+
+The `DecoupleAfterBatch` converter automatically modifies the collector's configuration for the Lambda distribution. Its purpose is to ensure that a decouple processor is always present after a batch processor in a pipeline, in order to prevent potential data loss due to the Lambda environment being frozen.
+
+## Behavior
+
+The converter scans the collector's configuration and makes the following adjustments:
+
+1. If a pipeline contains a batch processor with no decouple processor defined after it, the converter will automatically add a decouple processor to the pipeline immediately after the batch processor.
```
adcharre: Re-reading this after you added the additional tests case I note that this isn't strictly true - it doesn't add the decouple process _immediately_ after the batch process, in fact it adds it to the end of the pipeline.

So perhaps this is a better description:

```suggestion
1. If a pipeline contains a batch processor with no decouple processor defined after it, the converter will automatically add a decouple processor to the end of the pipeline.
```

adcharre (commented):
Looking good, just one pedantic comment about the readme.

nslaughter:
> Looking good, just one pedantic comment about the readme.

I appreciate the details. Thank you.

adcharre (approved):
LGTM

tylerbenson (approved):
Nice work!

rapphil (approved):

collector/processor/decoupleprocessor/README.md
```diff
@@ -7,22 +7,33 @@
...
 
-When combined with the batch processor, the number of exports required can be significantly reduced and therefore the cost of running the lambda. This is with the trade-off that the data will not be available at your chosen endpoint until some time after the invocation, up to a maximum of 5 minutes (the timeout that the environment is shutdown when no further invocations are received). 
+When combined with the batch processor, the number of exports required can be significantly reduced and therefore the cost of running the lambda. This is with the trade-off that the data will not be available at your chosen endpoint until some time after the invocation, up to a maximum of 5 minutes (the timeout that the environment is shutdown when no further invocations are received).
+
+## Auto-Configuration
+
+Due to the significant performance improvements with this approach, the OpenTelemetry Lambda Layer automatically configures the decouple processor when the batch processor is used. This ensures the best performance by default.
```
tylerbenson, line 18: Please also update `collector/README.md` with similar info.

nslaughter: Agree. I pushed the changes. Please let me know your thoughts.

PR #1276, opened 2024-04-24
Configure java runtime in the sample apps

vasireddy99:
This PR allows to configure java runtime in the sample apps at the time of deploy.


rapphil (approved):

rapphil:
LGTM, but I would like to hear from @tylerbenson as well.

tylerbenson (approved):
Looks reasonable to me.  Please update the readme's with the new parameter option.

rapphil (approved):

PR #1283, opened 2024-04-30
Try out the new multi-directory support in dependabot.

tylerbenson:
https://github.blog/changelog/2024-04-29-dependabot-multi-directory-configuration-public-beta-now-available/

(Renamed `nodeps-requirements.txt` so it can also get picked up by dependabot. https://stackoverflow.com/a/75909900/210800)


rapphil (approved):

python/src/otel/Makefile
```diff
@@ -3,7 +3,7 @@ export SDK=$(shell pwd)/otel_sdk
 build-OTelLayer:
 	mkdir -p $(ARTIFACTS_DIR)/python
 	python3 -m pip install -r $(SDK)/requirements.txt -t $(ARTIFACTS_DIR)/python
-	python3 -m pip install -r $(SDK)/requirements-nodeps.txt -t $(ARTIFACTS_DIR)/tmp --no-deps
+	python3 -m pip install -r $(SDK)/nodeps-requirements.txt -t $(ARTIFACTS_DIR)/tmp --no-deps
```
tylerbenson, line 6: haha, oops... turns out dependabot was already working with the previous name. I could revert this part of the change or just leave it.

PR #1307, opened 2024-05-02
Add java wrapper layer build

tylerbenson:
This approach contains a few arbitrary decisions.
- I chose to tie it to the same release job as java as they will likely always have changes together (sdk version).
- No independent tagging for the wrapper release.
- The underlying reported version for the wrapper layer will report `--`. (Since the wrapper won't have a separate release page, I think this is ok.)

### DIFF
--- a/.github/workflows/release-layer-java.yml
+++ b/.github/workflows/release-layer-java.yml
@@ -31,13 +31,19 @@ jobs:
       - name: Execute Gradle build
         run: |
           cd java
-          ./gradlew :layer-javaagent:assemble --scan --stacktrace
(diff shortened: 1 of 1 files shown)


rapphil:
> I chose to tie it to the same release job as java as they will likely always have changes together (sdk version).

I'm not sure if this will always be true. The java agent also includes instrumentation for a bunch of libraries no while the wrapper will only contain aws-lambda instrumentation + Otel SDK.

Having said that I think it is fine to release them together for now.

rapphil (approved):

PR #1312, opened 2024-05-10
Add go sample app path in dependabot

vasireddy99:
Add sample app path in dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,6 +21,7 @@ updates:
       - "/collector/processor/coldstartprocessor"
       - "/collector/processor/decoupleprocessor"
       - "/collector/receiver/telemetryapireceiver"
+      - "/go/sample-apps/function"
     schedule:
       interval: "weekly"
     groups:


.github/dependabot.yml
```diff
@@ -21,6 +21,7 @@ updates:
       - "/collector/processor/coldstartprocessor"
       - "/collector/processor/decoupleprocessor"
       - "/collector/receiver/telemetryapireceiver"
+      - "/go/sample-apps/function"
```
tylerbenson, line 24: I didn't include this initially because I didn't want to be responsible for fixing it if something breaks.

nslaughter: This line is just for `gomod`, which only manages Go module dependencies. That directory just has a hundred lines of Go and outside of OTel depends on a handful of AWS SDK packages. So hopefully that means extra notifications will be minimal.

My thinking is that this visibility will be good for maintenance.. and when an update needs more effort than the project can support at the time a maintainer/approver could dismiss dependabot's alert or add a rule to auto-triage it.

tylerbenson: ok, very well.

tylerbenson (approved):

PR #1346, opened 2024-05-24
Bug fixing: TelemetryAPI receiver

jerrytfleung:
- Decoupled API version and Schema version
- Updated schema version to "2022-12-13"
- Fixed Record datatype. It should be `any`. Record could be a string in [function](https://docs.aws.amazon.com/lambda/latest/dg/telemetry-schema-reference.html#telemetry-api-function) type.

### DIFF
--- a/collector/internal/telemetryapi/client.go
+++ b/collector/internal/telemetryapi/client.go
@@ -27,8 +27,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


tylerbenson (approved):
LGTM.  @rapphil @nslaughter PTAL.

rapphil:
Late to the party, but LGTM as well.

PR #1347, opened 2024-05-25
Feature: Generate function and extension logs via Telemetry API receiver

jerrytfleung:
## Motivation

Telemetry API can provide function and extension log. It is better to forward logs so as to provide more observability. 

## Changes 
- Changed `Subscribe` function so it takes event type slice in `telemetryapi/client.go`
- Added logs support
- Added sharedcomponent (copied from opentelemetry-collector-contrib) so as to share the http server among receivers
- Added unit test cases for logs


collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -64,7 +69,7 @@ func (r *telemetryAPIReceiver) Start(ctx context.Context, host component.Host) e
 	}()
 
 	telemetryClient := telemetryapi.NewClient(r.logger)
-	_, err := telemetryClient.Subscribe(ctx, r.extensionID, fmt.Sprintf("http://%s/", address))
+	_, err := telemetryClient.Subscribe(ctx, []telemetryapi.EventType{telemetryapi.Platform, telemetryapi.Function, telemetryapi.Extension}, r.extensionID, fmt.Sprintf("http://%s/", address))
```
jerrytfleung: I wonder if it is a good idea to setup a config that users can choose which type of events to subscribe.
Something like: 
```
receivers:
  telemetryapi:
    types: [platform, function, extension]
```
But after I take a look to `config.go`. There is a field extensionID. Is it possible to let `internal/lifecycle/manager` setup the extensionID and users setup the `types` in the collector configuration?

tylerbenson:
@rapphil @nslaughter PTAL.

collector/receiver/telemetryapireceiver/receiver.go
jerrytfleung: @tylerbenson @rapphil @nslaughter Can you please help to advise the above question? Thanks!

nslaughter: @jerrytfleung
1. Sorry about the delay in getting you this review. I appreciate the work you're doing on this. Great stuff!
2. I'm also not familiar enough with the extension API to say exactly how that will translate to the config from the top of mind. 
3. The config change that you mention seems to me like it **_will benefit some users and be implemented in a way that's unobjectionable to others_**, but I think these are questions we'd be best to hash out ~~in a Github issue - ideally accompanied by a very focussed PR that includes the proposed implementation~~ with maintainers, but this comment thread is fine for that discussion.

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -161,26 +180,133 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
...
+					case "FATAL2":
+						logRecord.SetSeverityNumber(22)
+					case "FATAL3":
+						logRecord.SetSeverityNumber(23)
+					case "FATAL4":
+						logRecord.SetSeverityNumber(24)
+					default:
+					}
+					logRecord.SetSeverityText(logRecord.SeverityNumber().String())
+				}
```
nslaughter, line 217: I haven't implemented much using this part of the OTel logging code, but I think there's some answer to severity parsing challenges [in contrib/pkg/stanza](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/types/severity.md). There are also useful constants [in plog](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/plog/severity_number.go).

If you don't find a callable solution in the contrib code referenced then it could be sufficient to replace this switch with something like a call to a new helper function `func severityTextToNumber(severityText string) SeverityNumber`, which just looks up on map[string]SeverityNumber.

collector/receiver/telemetryapireceiver/factory.go
```diff
@@ -39,14 +40,32 @@ func NewFactory(extensionID string) receiver.Factory {
...
 
 func createTracesReceiver(ctx context.Context, params receiver.CreateSettings, rConf component.Config, next consumer.Traces) (receiver.Traces, error) {
 	cfg, ok := rConf.(*Config)
 	if !ok {
 		return nil, errConfigNotTelemetryAPI
 	}
+	r := receivers.GetOrAdd(cfg, func() component.Component {
+		return newTelemetryAPIReceiver(cfg, params)
+	})
+	r.Unwrap().(*telemetryAPIReceiver).registerTracesConsumer(next)
```
nslaughter: I would want more data protection in this file than I see in the top-level variable for `receivers`, but it looks like the code doesn't read from `receivers`. As I traced my steps through this code, I started thinking we wouldn't get a benefit from the call to `GetOrAdd` receivers. What if we changed that to the following?
```suggestion
	r := newTelemetryAPIReceiver(cfg, params)
	r.registerTracesConsumer(next)
```
And if that works we don't have to rely on copying `internal/sharedcomponent`.

nslaughter (commented):
Let me know what you think of the changes suggested in `factory.go`. That's a key to this PR I think.

collector/receiver/telemetryapireceiver/factory.go
rapphil: and actually this is a bit fragile. this means that you cannot use this receiver in lets say, more than one pipeline.

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -161,26 +180,133 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
...
+					logRecord.Body().SetStr(line)
+				}
+			}
+		}
+	}
+	return log, nil
+}
+
+func (r *telemetryAPIReceiver) registerTracesConsumer(next consumer.Traces) {
+	r.nextTraces = next
```
rapphil, line 273: since we are trying to reuse the http server and it seems that we want to have a single instance of traces consume and logs consumer, would it make sense for `nextTraces` to be an array?

This would allow to have multiple receivers in different pipelines.

collector/receiver/telemetryapireceiver/receiver.go
rapphil: >But after I take a look to config.go. There is a field extensionID. Is it possible to let internal/lifecycle/manager setup the extensionID and users setup the types in the collector configuration

I think we will need to refactor the code a bit so that the extension id is passed to the factory functions `createTracesReceiver` and `createLogsReceiver` through the scope. 

However, if we are adding configuration, we will also need to be able to support multiple instances of the receivers, which does not seem to be the case with the usage of this `SharedComponents` abstraction.

rapphil:
My overall comment is that I'm trying to understand why we need to have this:

> A dded sharedcomponent (copied from opentelemetry-collector-contrib) so as to share the http server among receivers

Can't we have multiple subscriptions using the telemetry client, each using a different port to subscribe to the events so that we don't have to have this singleton receiver? This will open the door for adding the configurations that you suggested [here](https://github.com/open-telemetry/opentelemetry-lambda/pull/1347/files#r1614280650), which I think is a good idea.

collector/receiver/telemetryapireceiver/factory.go
jerrytfleung: I was thinking to reuse the http server. `internal/sharedcomponent` can help to reuse the http server by looking up the receiver using cfg as a key. 
I was assuming we could have configuration for `telemetryapi`

collector/receiver/telemetryapireceiver/receiver.go
jerrytfleung: After trying out, I found that the `extensionID` is preserved if I add a field. I updated the code so as to support multiple instances of telemetry api receiver.
I think `SharedComponents` could help to reuse the http server by using `cfg` as a key to select the corresponding receiver

collector/receiver/telemetryapireceiver/factory.go
jerrytfleung: I think `sharedcomponent` could play a role so as to reuse http server and to support multiple instances of telemetry api receiver. 

Can you take a look to the updated change and let me know your feedback? Thanks!

collector/receiver/telemetryapireceiver/receiver.go
jerrytfleung: My thinking is to reuse http server within a receiver instance and a receiver instance has a single instance of traces consumer and logs consumers. 

I don't know the benefit of making `nextTraces` to be an array. Can you please elaborate more on that?

jerrytfleung:
> My overall comment is that I'm trying to understand why we need to have this:
> 
> > Added sharedcomponent (copied from opentelemetry-collector-contrib) so as to share the http server among receivers
> 
> Can't we have multiple subscriptions using the telemetry client, each using a different port to subscribe to the events so that we don't have to have this singleton receiver? This will open the door for adding the configurations that you suggested [here](https://github.com/open-telemetry/opentelemetry-lambda/pull/1347/files#r1614280650), which I think is a good idea.

I managed to support multiple instances of telemetry api receiver by allowing user to configure `port` in the telemetry API receiver configuration. Please help to take a look and let me know your feedback. Thanks!

example config
```
receivers:
    telemetryapi:
    telemetryapi/1:
      port: 4326
    telemetryapi/2:
      port: 4326
      types:
        - platform
        - function
    telemetryapi/3:
      port: 4326
      types: ["platform", "function"]
```

collector/receiver/telemetryapireceiver/receiver.go
jerrytfleung: Done.
I don't find a callable function in the contrib so I added the function you suggested.

jerrytfleung:
@rapphil @nslaughter PTAL. Thanks!

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -161,26 +184,118 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
...
+		"WARN4":    plog.SeverityNumberWarn4,
+		"ERROR":    plog.SeverityNumberError,
+		"ERROR2":   plog.SeverityNumberError2,
+		"ERROR3":   plog.SeverityNumberError3,
+		"ERROR4":   plog.SeverityNumberError4,
+		"FATAL":    plog.SeverityNumberFatal,
+		"FATAL2":   plog.SeverityNumberFatal2,
+		"FATAL3":   plog.SeverityNumberFatal3,
+		"FATAL4":   plog.SeverityNumberFatal4,
+		"CRITICAL": plog.SeverityNumberFatal,
```
jerrytfleung, line 261: Adding `CRITICAL` here. It is because Python `logging` library can log `CRITICAL` (50) and it is mapped to Fatal in opentelmetry-python https://opentelemetry-python.readthedocs.io/en/latest/_modules/opentelemetry/_logs/severity.html#std_to_otel

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -161,26 +184,118 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
...
+		"ALL":      plog.SeverityNumberTrace,
```
jerrytfleung, line 262: Adding `ALL` here. It is because Java `log4j` can log `ALL` and it is mapped to Trace in opentelemetry-java https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/a87245b842a03bc42af6b388bb837fe8884aadae/instrumentation/log4j/log4j-appender-2.17/library/src/main/java/io/opentelemetry/instrumentation/log4j/appender/v2_17/internal/LogEventMapper.java#L201

jerrytfleung (commented):
Added comments to explain the decision of add `CRITICAL` and `ALL` severity text conversion.

jerrytfleung:
> My overall comment is that I'm trying to understand why we need to have this:
> 
> > A dded sharedcomponent (copied from opentelemetry-collector-contrib) so as to share the http server among receivers
> 
> Can't we have multiple subscriptions using the telemetry client, each using a different port to subscribe to the events so that we don't have to have this singleton receiver? This will open the door for adding the configurations that you suggested [here](https://github.com/open-telemetry/opentelemetry-lambda/pull/1347/files#r1614280650), which I think is a good idea.

I am thinking the following configuration so users can configure the types of subscription per each receivers. Also they can put the receiver to the pipeline according to their requirements.
```
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: "localhost:4317"
      http:
        endpoint: "localhost:4318"
  telemetryapi/1:
    port: 4325
    types: [platform]
  telemetryapi/2:
    port: 4326
    types: [function]
exporters:
  logging:
    loglevel: debug

service:
  pipelines:
    traces:
      receivers: [otlp,telemetryapi/1,telemetryapi/2]
      exporters: [logging]
    metrics:
      receivers: [otlp]
      exporters: [logging]
    logs/1:
      receivers: [telemetryapi/1]
      exporters: [logging]
    logs/2:
      receivers: [telemetryapi/2]
      exporters: [logging]
  telemetry:
    metrics:
      address: localhost:8888


```

bhaskarbanerjee:
@jerrytfleung I see that the config you shared above is supporting metric in the pipeline as well? metrics:
```
metrics:
      receivers: [otlp]
      exporters: [logging]
```
But as far I see [this](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/receiver/telemetryapireceiver/receiver.go#L48) I can only see it supporting traces.

Do you mean that with this PR, the collector will start receiving and exporting metric as well as logs, in addition to the traces?

jerrytfleung:
> @jerrytfleung I see that the config you shared above is supporting metric in the pipeline as well? metrics:
> 
> ```
> metrics:
>       receivers: [otlp]
>       exporters: [logging]
> ```
> 
> But as far I see [this](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/receiver/telemetryapireceiver/receiver.go#L48) I can only see it supporting traces.
> 
> Do you mean that with this PR, the collector will start receiving and exporting metric as well as logs, in addition to the traces?

@bhaskarbanerjee Solarwinds has plan to extract information from platform event and convert them into metrics. e.g. When we see `phase=init` from initReport, we can increase the counter of `fass.coldstarts` and it is a FaaS metrics. Currently, it is still WIP from Solarwinds due to some internal discussion.

This PR covers logs only.

tylerbenson:
@rapphil @jerrytfleung What's the status of this PR? Are there more changes needed, or is resolving the merge conflict enough?

@jerrytfleung maybe you could come to the next SIG meeting and present about these changes?  I'm having a hard time understanding and I think that would help.

jerrytfleung:
> @rapphil @jerrytfleung What's the status of this PR? Are there more changes needed, or is resolving the merge conflict enough?
> 
> @jerrytfleung maybe you could come to the next SIG meeting and present about these changes? I'm having a hard time understanding and I think that would help.

I resolved the merge conflict. I don't have other changes to add. This PR targets to extract the log messages in telemetry API and export them.

I am happy to join the SIG meeting to present the idea / result and answer questions. Will see you then.

collector/receiver/telemetryapireceiver/receiver_test.go
```diff
@@ -161,3 +170,301 @@ func TestCreatePlatformInitSpan(t *testing.T) {
...
+		},
+		{
+			level:  "ALL",
+			number: plog.SeverityNumberTrace,
+		},
+		{
+			level:  "UNKNOWN",
+			number: plog.SeverityNumberUnspecified,
+		},
+	}
```
nslaughter, line 388: I think this test simply specifies the severity map in another form. If you think it's beneficial I'd rather see the condensed format. Could move that map to top-level and create a separate map here as the oracle/golden.

collector/receiver/telemetryapireceiver/config_test.go
```diff
@@ -15,11 +15,122 @@
...
+			cfg := factory.CreateDefaultConfig()
+			sub, err := cm.Sub(tt.id.String())
+			require.NoError(t, err)
+			require.NoError(t, sub.Unmarshal(cfg))
+			require.NoError(t, component.ValidateConfig(cfg))
+			require.Equal(t, tt.expected, cfg)
+		})
+	}
+}
+
```
nslaughter, line 118: There's an approach in other Go tests in the project that would reduce some duplication and improve the specification of these tests. It looks more like this:

```
func TestLoadConfig(t *testing.T) {
    t.Parallel()

    // Helper function to create expected Config
    createExpectedConfig := func(types ...string) *Config {
        return &Config{
            extensionID: "extensionID",
            Port:        12345,
            Types:       types,
        }
    }

    tests := []struct {
        name     string
        id       component.ID
        expected component.Config
    }{
        {
            name:     "default",
            id:       component.NewID(component.MustNewType("telemetryapi")),
            expected: NewFactory("extensionID").CreateDefaultConfig(),
        },
        {
            name:     "all types",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "1"),
            expected: createExpectedConfig(platform, function, extension),
        },
        {
            name:     "platform only",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "2"),
            expected: createExpectedConfig(platform),
        },
        {
            name:     "function only",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "3"),
            expected: createExpectedConfig(function),
        },
        {
            name:     "extension only",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "4"),
            expected: createExpectedConfig(extension),
        },
        {
            name:     "platform and function",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "5"),
            expected: createExpectedConfig(platform, function),
        },
        {
            name:     "platform and extension",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "6"),
            expected: createExpectedConfig(platform, extension),
        },
        {
            name:     "function and extension",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "7"),
            expected: createExpectedConfig(function, extension),
        },
        {
            name:     "empty types",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "8"),
            expected: createExpectedConfig(),
        },
        {
            name:     "function and extension (alternative syntax)",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "9"),
            expected: createExpectedConfig(function, extension),
        },
        {
            name:     "function and extension (another syntax)",
            id:       component.NewIDWithName(component.MustNewType("telemetryapi"), "10"),
            expected: createExpectedConfig(function, extension),
        },
    }

    for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
            cm, err := confmaptest.LoadConf(filepath.Join("testdata", "config.yaml"))
            require.NoError(t, err)

            factory := NewFactory("extensionID")
            cfg := factory.CreateDefaultConfig()

            sub, err := cm.Sub(tt.id.String())
            require.NoError(t, err)
            require.NoError(t, sub.Unmarshal(cfg))
            require.NoError(t, component.ValidateConfig(cfg))
            
            require.Equal(t, tt.expected, cfg)
        })
    }
}

**Review by nslaughter — APPROVED** [review]
Good test coverage. I'm going to approve now. Would prefer to see those changes implemented for maintainability, but don't want to slow this feature down for that.

File: collector/receiver/telemetryapireceiver/config_test.go (hunk 8 above)
Comment (jerrytfleung) [reply collector/receiver/telemetryapireceiver/config_test.go:118]: Applied. Thanks for recommendation!

File: collector/receiver/telemetryapireceiver/receiver_test.go (hunk 7 above)
Comment (jerrytfleung) [reply collector/receiver/telemetryapireceiver/receiver_test.go:388]: I am not sure what do you mean by `Could move that map to top-level and create a separate map here as the oracle/golden.`. Can you help to elaborate by an simple example? Thanks!

**Review by tylerbenson — APPROVED** [review]

File: collector/receiver/telemetryapireceiver/receiver_test.go (hunk 7 above)
Comment (jerrytfleung) [reply collector/receiver/telemetryapireceiver/receiver_test.go:388]: Done. Added the golden mapping table and out of boundary cases.

File: collector/receiver/telemetryapireceiver/receiver.go [hunk 9]
```diff
@@ -260,6 +260,7 @@ func severityTextToNumber(severityText string) plog.SeverityNumber {
 		"FATAL4":   plog.SeverityNumberFatal4,
 		"CRITICAL": plog.SeverityNumberFatal,
 		"ALL":      plog.SeverityNumberTrace,
+		"WARNING":  plog.SeverityNumberWarn,
```
jerrytfleung, line 263: Adding `WARNING` here. It is because Python logger can log `WARNING` level when the function record is in JSON format. 

Example
```
{
    "timestamp": "2024-07-12T20:06:53Z",
    "level": "WARNING",
    "message": "30 warning {'filters': [], 'name': 'root', 'level': 0, 'parent': None, 'propagate': True, 'handlers': [<LambdaLoggerHandlerWithFrameType (NOTSET)>], 'disabled': False, '_cache': {10: True, 20: True, 30: True}}",
    "logger": "root",
    "requestId": "dcbdc55b-25bb-49dd-8273-64a96181ca2a",
    "warning": [
        "from-extra",
        30
    ]
}
```

jerrytfleung:
@tylerbenson @nslaughter I updated the test cases and polished the code again. Please help to proceed. Thanks!

PR #1362, opened 2024-06-03
feat: ruby prod release

xuan-cao-swi:


ruby/README.md
```diff
@@ -9,6 +9,7 @@ Requirement:
 * [Go](https://go.dev/doc/install)
 * [Docker](https://docs.docker.com/get-docker)
 
+Current ruby layer only support ruby 3.2.0
```
tylerbenson:
```suggestion
Current ruby layer only supports ruby 3.2.0
```

.github/workflows/release-layer-ruby.yml
```diff
@@ -78,6 +78,6 @@ jobs:
       layer-name: opentelemetry-ruby
       component-version: ${{needs.build-layer.outputs.RUBY_SDK_VERSION}}
       runtimes: ruby3.2
```
tylerbenson, line 80: What's preventing you from adding `ruby3.3`?

xuan-cao-swi: I think current implementation that use template.yaml to build layer restrict multiple runtimes; especially with default [GEM_PATH](https://docs.aws.amazon.com/lambda/latest/dg/ruby-package.html#ruby-package-dependencies-layers) (e.g. ruby/gems/3.2.0). If use template.yaml and sam to build the layer, it will fetch the image 
`public.ecr.aws/sam/build-<language>-<language_version>` (e.g. `public.ecr.aws/sam/build-python3.9`; `public.ecr.aws/sam/build-ruby3.2`). 

I will create an issue to track the issue and find way to support multiple runtimes under one layer (if possible).

tylerbenson (approved):

PR #1375, opened 2024-06-11
feat(nodejs): configureAwsInstrumentation for downstream

mzl-md:
Enables downstream repos to configure [@opentelemetry/instrumentation-aws-sdk](https://www.npmjs.com/package/@opentelemetry/instrumentation-aws-sdk) using the global function `configureAwsInstrumentation()`. This is required for setting custom user attributes as described in the [documentation](https://www.npmjs.com/package/@opentelemetry/instrumentation-aws-sdk).


tylerbenson (commented):
Would you mind adding a unit test to validate this code path is working properly and doesn't have regressions in the future?

mzl-md:
Currently there are no unit tests at all. Setting up a test framework is not really the scope of this PR.

Also, I can't find a good way to do this. Sinon fails to stub `registerInstrumentations()` or the constructor of `AwsInstrumentation` and I see no other way to check what arguments are passed to AwsInstrumentation. I added a trivial test that checks if the global function is called, but I think that is not really helpful.

In my opinion, meaningful tests require restructuring the wrapper.ts in separate files.

tylerbenson (approved):
That's a reasonable criticism. I'd love to see contributions to help us improve testing across each language, but I understand that's not reasonable scope in a PR like this.  I appreciate you making an effort to at least add something.

PR #1381, opened 2024-06-12
fix(docs): Clarify that pre-built layers are provided

gshpychka:


tylerbenson (approved):
Haha, oops.  I missed this when we added the community layers. Thanks for fixing.

PR #1399, opened 2024-06-18
Move Tristan to Emeritus and add @nslaughter as Approver

tylerbenson:
I've also removed employer from those with Emeritus status as it's easy to become outdated.

Welcome to @nslaughter and thanks to @tsloughter for his previous contributions on the project.


nslaughter (approved):
Glad about this. Thanks.
I'm especially glad we found an approver with just one extra stroke in their surname.

rapphil (approved):

PR #1401, opened 2024-06-21
Autobump Collector VERSION to latest on build - 

nslaughter:
The purpose of the PR is to automate the Lambda Layer Collector build and release process by extracting version information from dependency files for the `LDFLAGS` in `go build` and store the information in the `VERSION` file where it can be used by processes (such as crafting release notes) where consistent version information is needed.


collector/Makefile
```diff
@@ -4,8 +4,7 @@ BASE_SPACE:=$(shell pwd)
 BUILD_SPACE:=$(BASE_SPACE)/build
 BUCKET_NAME:=lambda-artifacts-$(shell dd if=/dev/random bs=8 count=1 2>/dev/null | od -An -tx1 | tr -d ' \t\n')
 LAYER_NAME:=otel-collector
-
-VERSION=$(shell cat VERSION)
+VERSION=$(shell curl -s https://raw.githubusercontent.com/open-telemetry/opentelemetry-collector-contrib/main/versions.yaml | grep 'version:' | head -1 | awk '{print $$2}')
```
tylerbenson: What about grepping `/collector/go.mod` directly for `go.opentelemetry.io/collector ` (with space at end) and parsing the version number from there?

nslaughter: We can certainly switch to that approach.

nslaughter: @tylerbenson to confirm, you mean getting [this line ](https://github.com/open-telemetry/opentelemetry-lambda/blob/c55ad8742d9d4714a950fe4eaf11ed3246673577/collector/go.mod#L135) right?

tylerbenson: That's what I was thinking of... I'm concerned with the approach you proposed as it expects we're always on the latest release, but that's not necessarily the case.  Open to other ideas though.

collector/Makefile
```diff
@@ -4,14 +4,13 @@ BASE_SPACE:=$(shell pwd)
 BUILD_SPACE:=$(BASE_SPACE)/build
 BUCKET_NAME:=lambda-artifacts-$(shell dd if=/dev/random bs=8 count=1 2>/dev/null | od -An -tx1 | tr -d ' \t\n')
 LAYER_NAME:=otel-collector
-
-VERSION=$(shell cat VERSION)
```
tylerbenson, line 8: Is the `VERSION` file being used for anything but this? If not, I think the file should be deleted and report the version through the environment variable.

collector/Makefile
nslaughter: made changes

collector/Makefile
tylerbenson: That is for Javascript... I don't think it's pulling the VERSION from the collector folder.

nslaughter: All I can say at the moment is that after reviewing the search, I'm not comfortable deleting it. I did some more looking, but I didn't want to keep confirming that one more thing didn't rely on that file. I would rather settle for the small but certain victory on this case.

nslaughter: You were right that my comment pointed to the javascript. I found another oddity in there that I would like to explore where version is 0.0.0 in one of the configs, but in both cases I'd take a bit of focused time on the issue.

tylerbenson: I understand being cautious about it... if you think it's important to keep it updated, that's fine. Perhaps it's better to have it be local only and not checked into git?

tylerbenson (approved):
Thanks for doing this.

rapphil (approved):

collector/Makefile
```diff
@@ -4,14 +4,13 @@ BASE_SPACE:=$(shell pwd)
...
+OTELCOL_VERSION="UNSET" # run recipe set-otelcol-version to get this
 GIT_SHA=$(shell git rev-parse HEAD)
 GOARCH ?= amd64
 GOBUILD=GO111MODULE=on CGO_ENABLED=0 installsuffix=cgo go build -trimpath
 BUILD_INFO_IMPORT_PATH=main
 
-LDFLAGS=-ldflags "-s -w -X $(BUILD_INFO_IMPORT_PATH).GitHash=$(GIT_SHA) -X $(BUILD_INFO_IMPORT_PATH).Version=$(VERSION) \
+LDFLAGS=-ldflags "-s -w -X $(BUILD_INFO_IMPORT_PATH).GitHash=$(GIT_SHA) -X $(BUILD_INFO_IMPORT_PATH).Version=$(OTELCOL_VERSION) \
```
rapphil, line 13: TIL variables in Makefiles are dynamically expanded.

rapphil: https://www.gnu.org/software/make/manual/html_node/Using-Variables.html#:~:text=Variables%20and%20functions%20in%20all,at%20the%20time%20of%20expansion.

Issue #1196, opened 2024-03-11
Add `decouple` processor to the default config


nslaughter:
**Is your feature request related to a problem? Please describe.**
Without the `decouple` processor in the pipeline there have been lengthy delays waiting on the Collector layer to flush telemetry. The [README explains](https://github.com/open-telemetry/opentelemetry-lambda/blob/collector/README.md) the situation.

**Describe the solution you'd like**
Add `decouple` processor to the default pipeline config. This is just adopting a sensible default that's more likely to benefit the community as a whole and will not limit the opportunity of users who disagree with the approach to override for the present state.

**Describe alternatives you've considered**
The current approach is to document for users how to override default config. This doesn't seem like the right approach given the significant benefit of using the `decouple` processor.

**Additional context**
I don't currently have the comparison data accessible, however this has been the subject of some research in the community and I can add details for discussion.

xThomo:
You linked the Makefile in the issue description, I think you meant this [README](https://github.com/open-telemetry/opentelemetry-lambda/blob/597b0d31405d28083d4100f96aef3c0ca1e1c096/collector/README.md#improving-lambda-responses-times)

Any idea when #1197 will be merged? It looks like a simple change.

nslaughter:
> You linked the Makefile in the issue description, I think you meant this [README](https://github.com/open-telemetry/opentelemetry-lambda/blob/597b0d31405d28083d4100f96aef3c0ca1e1c096/collector/README.md#improving-lambda-responses-times)
> 
> Any idea when #1197 will be merged? It looks like a simple change.

Thanks for the correction and the interest in #1197. What's happened here is an intervening idea where the decoupleprocessor could be added to the end of pipelines which contain a batch processor. I've written up the [code for that in my fork](https://github.com/nslaughter/opentelemetry-lambda/pull/1) and will be documenting it better before sharing for approval next week.

nslaughter:
Addressed in #1255

PR #1427, opened 2024-07-15
Add custom build support for lambdacomponents

its-felix:
Adding support for customized builds of the collector.

## Motivation
I hoped to use the `spanmetricsconnector` using this collector, but sadly found it not to be part of this distribution.

I also tried doing so in a customized fork, but found it to be extremely cumbersome to do with how go handles `internal`-packages together with forks. With how the project is structured, I saw myself adding a lot of `replace` instructions to the different `go.mod` files to make a successful build of my fork.

I came to the conclusion that others might also want to use more specialized versions for their usecases and figured out this could be done without affecting the distribution for everyone else by making use of build tags.

### Customized fork of this repository
As mentioned above, I initially considered forking this repository and adding my desired components to my fork.
However, I faced issues building from the fork because of how go handles the `internal`-Package. I was able to solve the issues I was facing, but only by adding a lot of `replace` instructions to several `go.mod` files.

This might be just due to my limited knowledge of go's build system.

### Customized collector, utilizing this repositories collector as a library
Similar to https://medium.com/opentelemetry/building-your-own-opentelemetry-collector-distribution-42337e994b63

I tried building my own collector layer, utilizing the lambda-lifecycle of this project to integrate the resulting collector with AWS Lambdas extension API.

Trying to import the following packages:
```golang
import (
	"github.com/open-telemetry/opentelemetry-lambda/collector/internal/lifecycle"
	"github.com/open-telemetry/opentelemetry-lambda/collector/lambdalifecycle"
)
```

Results in the following error on a fresh go project:
```
go: github.com/gw2auth/custom-opentelemetry-lambda-collector imports
	github.com/open-telemetry/opentelemetry-lambda/collector/internal/lifecycle imports
	github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents: reading github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents/go.mod at revision collector/lambdacomponents/v0.98.0: unknown revision collector/lambdacomponents/v0.98.0
```

It seems not as easy to do this as I initially thought it would be.

Again, this might be just due to my limited knowledge of go's build system.

### [OCB](https://opentelemetry.io/docs/collector/custom-collector/)
OCB works great for using a customized set of components not yet covered in any of the existing distributions or even adding fully custom componnents. However, this results in a "standard" collector which doesn't include the AWS Lambda extension API communication.

### Conclusion
I checked previous PRs and Issues of this project and found that the usual approach was to add components (https://github.com/open-telemetry/opentelemetry-lambda/issues/102 , https://github.com/open-telemetry/opentelemetry-lambda/issues/1196) as long as these additional components are expected to be useful for a reasonable number of users of this specialized collector.

However, this approach does not work well for more niche components, which might only be useful for few usecases.

This PR adds an in-between method to include components, without inflating the collector binary with components which would only be useful for a minority.
Additionally, this PR enables everyone to reduce their self-built collector to only the set of components they actually need, potentially reducing the size of the resulting binary compared to the default set of components.

## Usage
If the collector is built without build tags, the `default.go` Components will be used.

If the collector is built with `lambdacomponents.custom`, the `custom.go` Components will be used.
This allows a customized set of components to be included, for example:

```shell
go build -tags "lambdacomponents.custom,lambdacomponents.receiver.all,lambdacomponents.processor.all,lambdacomponents.exporter.otlphttp,lambdacomponents.connector.spanmetrics"
```

This would include:
- All receivers
- All processors
- No extensions
- Only the otlphttp exporter
- Only the spanmetrics connector

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ Some layers include the corresponding OTel language SDK for the Lambda. This all
 
 * **What exporters/receivers/processors are included from the OpenTelemetry Collector?**
     > You can check out [the stripped-down collector's imports](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/lambdacomponents/default.go#L18) in this repository for a full list of currently included components.
+  
+    > Self-built binaries of the collector have **experimental** support for a custom set of connectors/exporters/receivers/processors. For more information, see [(Experimental) Customized collector build](#experimental-customized-collector-build)
 * **Is the Lambda layer provided or do I need to build it and distribute it myself?**
     > This repository provides pre-built Lambda layers, their ARNs are available in the [Releases](https://github.com/open-telemetry/opentelemetry-lambda/releases). You can also build the layers manually and publish them in your AWS account. This repo has files to facilitate doing that. More information is provided in [the Collector folder's README](collector/README.md).
 
@@ -98,6 +100,45 @@ The following are runtimes which are no longer or not yet supported by this repo
 [2]: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/faas-spans.md#incoming-invocations
 [3]: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/faas-spans.md#outgoing-invocations
 [4]: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/faas-metrics.md
+
+## (Experimental) Customized collector build
+The collector can be built with a customized set of connectors/exporters/receivers/processors. This feature is **experimental** and is only supported for self-built binaries of the collector.
+
+### Build Tags
+The build-tag `lambdacomponents.custom` must always be provided to opt-in for a custom build.
+Once this build-tag is present, you need provide additional build-tags to include your desired components in the resulting binary:
+
+- `lambdacomponents.all` includes all available components
+- `lambdacomponents.connector.all` includes all available connectors
+- `lambdacomponents.exporter.all` includes all available exporters
+- `lambdacomponents.extension.all` includes all available extensions
+- `lambdacomponents.processor.all` includes all available processors
+- `lambdacomponents.receiver.all` includes all available receivers
+
+Each available component can also be included explicitly by using its specific build-tag. For a full-list of available components, have a look into the [lambdacomponents](./collector/lambdacomponents) package.
+
+As an example, the full build command including the following components:
+- All receivers
+- All processors
+- No extensions
+- Only the otlphttp exporter
+- Only the spanmetrics connector
+
+would be the following:
+```shell
+go build -tags "lambdacomponents.custom,lambdacomponents.receiver.all,lambdacomponents.processor.all,lambdacomponents.exporter.otlphttp,lambdacomponents.connector.spanmetrics"
+```
+
+### Adding additional options
+To add more options for a customized build, you can add your desired component to the [lambdacomponents](./collector/lambdacomponents) package.
+Make sure to always restrict your addition using the appropriate build-tags.
+
+For example, if you want to add the extension `foo`, the file providing this extension should be located in the [extension](./collector/lambdacomponents/extension) directory have to following build restriction:
+```
+//go:build lambdacomponents.custom && (lambdacomponents.all || lambdacomponents.extension.all || lambdacomponents.extension.foo)
+```
+
+You can provide your addition as a pull-request to this repository. Before doing so, please also read through the details of [Contributing](#contributing) to this project.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ go build -tags "lambdacomponents.custom,lambdacomponents.receiver.all,lambdacomp
 To add more options for a customized build, you can add your desired component to the [lambdacomponents](./collector/lambdacomponents) package.
 Make sure to always restrict your addition using the appropriate build-tags.
 
-For example, if you want to add the extension `foo`, the file providing this extension should be located in the [extension](./collector/lambdacomponents/extension) directory have to following build restriction:
+For example, if you want to add the extension `foo`, the file providing this extension should be located in the [extension](./collector/lambdacomponents/extension) directory have the following build restriction:
 ```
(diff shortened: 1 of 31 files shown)


rapphil:
I like this idea of making the life of user who want to extend the collector layer easier. However I'm not sure if we are ready to fully support this long term. I'm proposing to clearly state that this feature/characteristic of the repository is experimental for now and we highly advertise it so that users can start experimenting. WDYT @tylerbenson ?

collector/lambdacomponents/custom.go
```diff
@@ -0,0 +1,127 @@
...
+func (fn FactoryWithExtensionIdFn[T]) Get(extensionId string) T {
+	return fn(extensionId)
+}
+
+var (
+	receiverFactories  []factory[receiver.Factory]
+	processorFactories []factory[processor.Factory]
+	exporterFactories  []factory[exporter.Factory]
+	extensionFactories []factory[extension.Factory]
+	connectorFactories []factory[connector.Factory]
```
rapphil: can we simplify this? can this be a list of functions (providers) that receive a extensionid as parameter and then return a factory as return value?

```
	receiverFactories  []func(extensionid string) receiver.Factory
	processorFactories []func(extensionid string) processor.Factory
	exporterFactories  []func(extensionid string)  exporter.Factory
	extensionFactories []func(extensionid string)  extension.Factory
	connectorFactories []func(extensionid string) connector.Factory
```

Then in each init block in the modules you can use the `AddReceiverFactoryProvider` & etc to append to the slices.

e.g.:
```
func init() {
	lambdacomponents.AddReceiverFactoryProvider(func(extensionId string) receiver.Factory { return telemetryapireceiver.NewFactory(extensionID) })
}
```
then in `Components` you just iterate over the elements invoking the provider to get the factories and pass that to the appropriate `MakeFactoryMap`.

It think what you did is very nice but I prefer to keep the code simpler so that it is simpler to maintain and more welcoming for people trying to understand the code to extend the collector layer using your approach.

rapphil:
Thank you for your PR. This is a very neat idea.

How would be the workflow for users who want to extend the collector layer? users will need to fork the repository and add code to instantiate new factories with new modules or is your idea to use the layer as a library somehow.

Can you create a section in the README detailing how to extend the collector layer? We should make it clear that this is experimental.

its-felix:
Hey, thanks for the feedback. I'm looking forward to implement it as suggested.
Just want to let you know that I've read your comments, I am just pretty busy right now and will need a bit to get back onto this PR.

its-felix:
Hey @rapphil ,

I have addressed your comments and rebased my fork. For the build to work, I had to move the definition of each of the Factory-Slices to each package, because otherwise there would've been a cyclic dependency.

Please let me know if there's anything you'd like to see adjusted :)

its-felix:
Collector binary sizes built with these commands:
```
GOOS=linux GOARCH=amd64 go build -tags "lambdacomponents.custom,lambdacomponents.all" -o "collector.full.bin"
GOOS=linux GOARCH=amd64 go build -tags "lambdacomponents.custom" -o "collector.none.bin"
GOOS=linux GOARCH=amd64 go build -o "collector.default.bin"
```

```
52M collector.default.bin
53M collector.full.bin
41M collector.none.bin
```

README.md
```diff
@@ -99,6 +101,45 @@ The following are runtimes which are no longer or not yet supported by this repo
 [3]: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/faas-spans.md#outgoing-invocations
 [4]: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/faas/faas-metrics.md
 
+## (Experimental) Customized collector build
```
tylerbenson: These sections might be better in the `collector/README.md` instead, especially since that documents how to publish the resulting layer.

tylerbenson (approved):
Aside from moving the readme changes to the collector readme, I'm ok with this change.

README.md
its-felix: Moved the section to the collector README and updated the link in the main README accordingly

collector/README.md
```diff
@@ -13,6 +13,45 @@ Be sure to:
...
+As an example, the full build command including the following components:
+- All receivers
+- All processors
+- No extensions
+- Only the otlphttp exporter
+- Only the spanmetrics connector
+
+would be the following:
+```shell
+go build -tags "lambdacomponents.custom,lambdacomponents.receiver.all,lambdacomponents.processor.all,lambdacomponents.exporter.otlphttp,lambdacomponents.connector.spanmetrics"
```
rapphil: can we just add more details on how to build and publish this custom layer to the person's account?

we just need to include:
* add details about how to zip.
* how to publish. (the aws cli command that can be used).

I'm looking for end to end instructions in case someone not experienced with the process can follow. This will make the custom builds fully usable by anyone.

nslaughter (changes requested):
First things first, awesome feature and I identify with the motivation expressed in the PR. Here are a few comments:

My one requested change here is the addition of an ALTERNATIVES CONSIDERED section including consideration of employing the Collector’s `ocb` for custom builds. I try to provide some helpful resources, including more detail on the custom build in the notes below for consideration.

See Juraci's article - https://medium.com/opentelemetry/building-your-own-opentelemetry-collector-distribution-42337e994b63 on custom buillds in the Collector.

I noticed in the PR message that imports were causing some issues. If I understand the issues you encountered correctly this problem of patching `go.mod` with `replace` statements can be resolved by developing with Go `workspaces`, and some more about using workspaces for custom Collector development with the Collector is outlined in (the doc on building receivers)[https://opentelemetry.io/docs/collector/building/receiver/]. That's also a good introduction to the custom builder in practice.

Reiterating… awesome feature and I’m excited about the initiative and effort that’s brought us here.

collector/README.md
its-felix: Added this to the `Makefile` making use of the env variable `BUILDTAGS` and adjusted the README accordingly

its-felix:
@nslaughter I added the requested section to the main comment of this PR

rapphil:
This PR looks great! LGTM. @nslaughter  do you have any other concerns?

rapphil (approved):

nslaughter (approved):
LGTM
Thanks for the diligence in updating changes.

PR #1444, opened 2024-07-19
Fix wrapper publish step

tylerbenson:
When I originally added it, I used the wrong file name so build was failing with `Error parsing parameter '--zip-file': Unable to load paramfile fileb://opentelemetry-javawrapper-layer.zip: [Errno 2] No such file or directory: 'opentelemetry-javawrapper-layer.zip'`

### DIFF
--- a/.github/workflows/release-layer-java.yml
+++ b/.github/workflows/release-layer-java.yml
@@ -108,7 +108,7 @@ jobs:
           - us-west-1
           - us-west-2
     with:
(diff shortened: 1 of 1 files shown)


tylerbenson:
For reference: https://github.com/open-telemetry/opentelemetry-lambda/actions/runs/9995205724/job/27626839065

rapphil (approved):

PR #1469, opened 2024-08-14
feat: update nodejs lambda to the latest standards and packages

pragmaticivan:
* Adds support to lock files for predictable builds.
* Update major version of lerna
* Run `npm run lint:fix`


pragmaticivan:
Might need to consider that as a breaking change with the nodejs16.x removal

tylerbenson (approved):
I'm not opposed to removing node 16 support, especially if it's not supported by AWS anymore.  Perhaps better as a separate PR though.  Thanks for the contribution and look forward to more participation.

martinkuba (approved):
LGTM

PR #1481, opened 2024-08-21
as aws will not support Java8, we should remove it

### LINKED ISSUE #1442: Consider removing support to java8 runtime
**Describe the bug**
Java8 support is reaching end of the life. No new function can be created with the Java8 runtime.
https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-deprecated

**Steps to reproduce**
Try to create a new function that uses java8.

That opentelemetry-lambda did not support java8.

**What did you see instead?**
Java8 is still supported. 

**Additional context**
It is ok to deprecate java8 in my opinion because we still will support java8.al2.
https://aws.amazon.com/blogs/compute/announcing-migration-of-the-java-8-runtime-in-aws-lambda-to-amazon-corretto/

knarfli:
Fixes #1442

### DIFF
--- a/.github/workflows/release-layer-java.yml
+++ b/.github/workflows/release-layer-java.yml
@@ -80,7 +80,7 @@ jobs:
       layer-name: opentelemetry-javaagent
       component-version: ${{needs.build-layer.outputs.JAVAAGENT_VERSION}}
       # architecture:
-      runtimes: java8 java8.al2 java11 java17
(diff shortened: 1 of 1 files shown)


tylerbenson (approved):

rapphil (approved):

rapphil:
Thank you for your contribution!

knarfli:
My pleasure

PR #1483, opened 2024-08-21
Install all transitive dependencies for Python instrumentation

### LINKED ISSUE #1482: Python "packaging" is required
When trying to run instrumentation for a python lambda, I get the following error.  It implies I need to install the `packaging` package into my lambda, but why is this not in the otel layer if it is indeed required?  What is leveraging this?  Or am I missing a configuration?  I was following https://opentelemetry.io/docs/faas/lambda-auto-instrument/

```
Instrumenting of sqlalchemy failed
Traceback (most recent call last):
File "/opt/python/opentelemetry/instrumentation/auto_instrumentation/_load.py", line 87, in _load_instrumentors
distro.load_instrumentor(entry_point, skip_dep_check=True)
File "/opt/python/opentelemetry/instrumentation/distro.py", line 62, in load_instrumentor
instrumentor: BaseInstrumentor = entry_point.load()
^^^^^^^^^^^^^^^^^^
File "/var/task/pkg_resources/__init__.py", line 2470, in load
self.require(*args, **kwargs)
File "/var/task/pkg_resources/__init__.py", line 2493, in require
items = working_set.resolve(reqs, env, installer, extras=self.extras)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/var/task/pkg_resources/__init__.py", line 795, in resolve
raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'packaging>=21.0' distribution was not found and is required by the application
Failed to auto initialize opentelemetry
Traceback (most recent call last):
File "/opt/python/opentelemetry/instrumentation/auto_instrumentation/sitecustomize.py", line 39, in initialize
_load_instrumentors(distro)
File "/opt/python/opentelemetry/instrumentation/auto_instrumentation/_load.py", line 91, in _load_instrumentors
raise exc
File "/opt/python/opentelemetry/instrumentation/auto_instrumentation/_load.py", line 87, in _load_instrumentors
distro.load_instrumentor(entry_point, skip_dep_check=True)
File "/opt/python/opentelemetry/instrumentation/distro.py", line 62, in load_instrumentor
instrumentor: BaseInstrumentor = entry_point.load()
^^^^^^^^^^^^^^^^^^
File "/var/task/pkg_resources/__init__.py", line 2470, in load
self.require(*args, **kwargs)
File "/var/task/pkg_resources/__init__.py", line 2493, in require
items = working_set.resolve(reqs, env, installer, extras=self.extras)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/var/task/pkg_resources/__init__.py", line 795, in resolve
raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'packaging>=21.0' distribution was not found and is required by the application
```

rapphil:
Fixes: #1482

This change will install all Python transitive dependencies of the instrumentation supported by the lambda layer. This has to be done to reduce the friction for users, so that they are not required to install extra dependencies for their instrumentation to work.

There are some exceptions that will be handled separately because they are depending on a pinned version of opentelemetry-propagator-aws-xray. We wont need to keep those exceptions in the next release because it is already fixed https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2773

### DIFF
--- a/python/src/otel/otel_sdk/nodeps-requirements.txt
+++ b/python/src/otel/otel_sdk/nodeps-requirements.txt
@@ -1,31 +1,4 @@
-opentelemetry-instrumentation-aiohttp-client==0.47b0
-opentelemetry-util-http==0.47b0
-asgiref~=3.8
-opentelemetry-instrumentation-asgi==0.47b0
-opentelemetry-instrumentation-asyncpg==0.47b0
-opentelemetry-instrumentation-boto==0.47b0
-opentelemetry-instrumentation-boto3sqs==0.47b0
-opentelemetry-instrumentation-botocore==0.47b0
-opentelemetry-instrumentation-celery==0.47b0
-opentelemetry-instrumentation-dbapi==0.47b0
-opentelemetry-instrumentation-django==0.47b0
-opentelemetry-instrumentation-elasticsearch==0.47b0
-opentelemetry-instrumentation-fastapi==0.47b0
-opentelemetry-instrumentation-falcon==0.47b0
-opentelemetry-instrumentation-flask==0.47b0
-opentelemetry-instrumentation-grpc==0.47b0
-opentelemetry-instrumentation-jinja2==0.47b0
-opentelemetry-instrumentation-mysql==0.47b0
-opentelemetry-instrumentation-psycopg2==0.47b0
-opentelemetry-instrumentation-pymemcache==0.47b0
-opentelemetry-instrumentation-pymongo==0.47b0
-opentelemetry-instrumentation-pymysql==0.47b0
-opentelemetry-instrumentation-pyramid==0.47b0
-opentelemetry-instrumentation-redis==0.47b0
-opentelemetry-instrumentation-requests==0.47b0
-opentelemetry-instrumentation-sqlalchemy==0.47b0
-opentelemetry-instrumentation-sqlite3==0.47b0
-opentelemetry-instrumentation-starlette==0.47b0
-opentelemetry-instrumentation-tornado==0.47b0
-opentelemetry-instrumentation-wsgi==0.47b0
+# TODO: move this dependency to requirements.txt when this dependency gets updated to 0.48b0
+# This has to be done because 0.47b0 is pinning a specific version of opentelemetry-propagator-aws-xray
+# causing dependency resolution conflicts.
 opentelemetry-instrumentation-aws-lambda==0.47b0
--- a/python/src/otel/otel_sdk/nodeps-requirements.txt
+++ b/python/src/otel/otel_sdk/nodeps-requirements.txt
@@ -1,4 +1,4 @@
-# TODO: move this dependency to requirements.txt when this dependency gets updated to 0.48b0
-# This has to be done because 0.47b0 is pinning a specific version of opentelemetry-propagator-aws-xray
-# causing dependency resolution conflicts.
+# TODO: move these dependencies to requirements.txt when they stopped relying on a pinned version of
+# opentelemetry-propagator-aws-xray
 opentelemetry-instrumentation-aws-lambda==0.47b0
+opentelemetry-instrumentation-botocore==0.47b0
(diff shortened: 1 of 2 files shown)


python/src/otel/otel_sdk/nodeps-requirements.txt
```diff
@@ -1,31 +1,4 @@
...
-opentelemetry-instrumentation-pymysql==0.47b0
-opentelemetry-instrumentation-pyramid==0.47b0
-opentelemetry-instrumentation-redis==0.47b0
-opentelemetry-instrumentation-requests==0.47b0
-opentelemetry-instrumentation-sqlalchemy==0.47b0
-opentelemetry-instrumentation-sqlite3==0.47b0
-opentelemetry-instrumentation-starlette==0.47b0
-opentelemetry-instrumentation-tornado==0.47b0
-opentelemetry-instrumentation-wsgi==0.47b0
+# TODO: move these dependencies to requirements.txt when they stopped relying on a pinned version of
```
ocelotl, line 1: Since [this](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/2773) is already merged we can make a new Python release and this won't be necessary.

rapphil: yes, this will work for v.0.48b0 (which is not released yet.). being merged does not mean that it is released.

rapphil: for the current release, 0.47.b0, it is still using the the pinned version.

python/src/otel/otel_sdk/requirements.txt
```diff
@@ -2,5 +2,35 @@ opentelemetry-sdk==1.26.0
 opentelemetry-exporter-otlp-proto-http==1.26.0
 opentelemetry-distro==0.47b0
 opentelemetry-instrumentation==0.47b0
-opentelemetry-semantic-conventions == 0.47b0
+opentelemetry-semantic-conventions==0.47b0
 opentelemetry-propagator-aws-xray==1.0.2
+
+# Instrumentation dependencies
+opentelemetry-instrumentation-aiohttp-client==0.47b0
```
tylerbenson, line 9: How much does moving all this to the lambda layer increase the layer size? Is that enough to be concerned about for adding to the cold start time?

rapphil: I simulated locally to check what is the delta:

without installing all dependencies
```
python3 -m pip install -r otel_sdk/requirements.txt -t ./deps && \
  python3 -m pip install -r otel_sdk/nodeps-requirements.txt -t ./deps --no-deps
du -h --max-depth 0 deps
25M     deps
```

installing all transitive dependencies
```
python3 -m pip install -r otel_sdk/requirements.txt -t ./deps && \
  python3 -m pip install -r otel_sdk/nodeps-requirements.txt -t ./deps --no-deps
du -h --max-depth 0 deps
27M     deps
```

The impact on the cold startup will be negligible.

tylerbenson: Looks like the [last version](https://github.com/open-telemetry/opentelemetry-lambda/actions/runs/9995251242) was 7.4 MB.  We can see where things end up with the next release.

tylerbenson (approved):
I assume the no-deps idea was to allow users to only add dependency for the specific instrumentation they want, but bundling it all makes it easier to keep versions in sync.

python/src/otel/otel_sdk/requirements.txt
rapphil: the data presented before regarding size was just for the dependencies and it was not zipped. Here is the resulting layer for reference:
```
8.3M Aug 22 15:51 build/opentelemetry-python-layer.zip
```

tylerbenson: So an added 1mb.... that's not insignificant. 🤷

PR #1486, opened 2024-08-22
Fix Java release workflow - Make the name of the java wrapper artifact consistent

rapphil:
This has to be done so that the release workflow works properly. 

The name of the artifact in the upload action is just an arbitrary name to identify the artifact and it has nothing to do with the contents of it. https://github.com/actions/upload-artifact?tab=readme-ov-file#zip-archives


tylerbenson (approved):
:table-flip:

PR #1493, opened 2024-08-22
build: fix release layer nodejs

pragmaticivan:


rapphil:
Please check the content of the zipped layer file before you submit: 

I tried to use your branch and it looks like it still not including the dependencies: `4.6K Aug 22 16:06 layer.zip`

p.s.: Don't be blocked by lerna and workspaces. It is fine if you want to use a separated npm project for the sample app and for the layer.

pragmaticivan:
@rapphil I think this new change should fix everything, although I made a slight change on the build since it was also shipping `dev dependencies` in the layer.

rapphil:
Thank you! the resulting layer is smaller than before :)

rapphil (approved):

PR #1504, opened 2024-08-30
feat: set global meter provider when configureMeterProvider is not defined

pragmaticivan:
If you want to create a custom metric, there's no way to access the wrapper metric provider in the lambda code.

Usage:

```js
const meter = metrics.getMeterProvider().getMeter('default');
```

`getMeterProvider` will return the globally configured meter provider


tylerbenson (approved):
Seems reasonable to me.

serkan-ozal:
good catch!

serkan-ozal (approved):

pragmaticivan:
Since we have a ticket targeting this as a bug, wondering if this should be released as a hotfix

PR #1521, opened 2024-09-12
Rename OPENTELEMETRY_COLLECTOR_CONFIG_FILE to OPENTELEMETRY_COLLECTOR_CONFIG_URI

### LINKED ISSUE #645: Rename OPENTELEMETRY_COLLECTOR_CONFIG_FILE to appropriate usage
Currently `OPENTELEMETRY_COLLECTOR_CONFIG_FILE` is a environment variable used to set the config file from any of the config map providers such as [http, s3, yaml.](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/internal/collector/collector.go#L60) etc., as per the naming convention it is not exactly a file as the input to this variable could be URI format example [here](https://github.com/open-telemetry/opentelemetry-lambda/tree/main/collector#configuration). Hence creating this issue to see if this can be renamed to `OPENTELEMETRY_COLLECTOR_CONFIG` instead.

gshpychka:
As noted in #645, the name of the environment variable to set the collector config URI is confusing, since it does not only accept a file.

This PR renames it to `OPENTELEMETRY_COLLECTOR_CONFIG_URI` to better reflect what it means, while keeping the old name for backwards compatibility, with the new name taking precedence.

Fixes #645.

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -33,7 +33,7 @@ Alternatively, to configure the OpenTelemetry Lambda Extension via CloudFormatio
 
 ## Configuration
 
-By default, OpenTelemetry Collector Lambda layer exports telemetry data to AWS backends. To customize the collector configuration, add a `collector.yaml` to your function and specify its location via the `OPENTELEMETRY_COLLECTOR_CONFIG_FILE` environment file.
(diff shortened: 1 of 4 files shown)


gshpychka:
Do you squash on merge, or should I do that in the branch?

tylerbenson:
We squash, so you're fine.

tylerbenson (approved):

serkan-ozal (approved):

PR #1527, opened 2024-09-20
Increase stale-action's operations-per-run

tylerbenson:
We keep getting rate limited, so it's not getting through everything.

This number should be plenty for the size of our repo.


nslaughter (approved):
LGTM

serkan-ozal (approved):

PR #1528, opened 2024-09-20
fix: update references to logging exporter

codeboten:
This exporter has been replaced by the debug exporter and will be removed soon. Related to https://github.com/open-telemetry/opentelemetry-collector/pull/11037

### DIFF
--- a/collector/config.yaml
+++ b/collector/config.yaml
@@ -7,17 +7,17 @@ receivers:
         endpoint: "localhost:4318"
 
 exporters:
-  logging:
-    loglevel: debug
+  debug:
+    verbosity: detailed
 
 service:
   pipelines:
     traces:
       receivers: [otlp]
-      exporters: [logging]
+      exporters: [debug]
     metrics:
       receivers: [otlp]
-      exporters: [logging]
+      exporters: [debug]
   telemetry:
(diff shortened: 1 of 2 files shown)


tylerbenson (approved):
Welcome back!  You should visit more often 😉

collector/config.yaml
```diff
@@ -7,17 +7,17 @@ receivers:
         endpoint: "localhost:4318"
 
 exporters:
-  logging:
```
tylerbenson, line 10: Do we need to remove an underlying dependency here to avoid a build failure in the future?

codeboten: In the future yes, I was going to leave that for now (as people transition away from it in their config)

codeboten:
> Welcome back! You should visit more often 😉

Only so much time in the day 😅

collector/config.yaml
tylerbenson: Ok.  I wasn't sure how quickly it was going to be removed.

serkan-ozal (approved):

nslaughter (approved):

nslaughter:
Thanks a bunch for this.

Issue #300, opened 2022-10-17
add check-links to this repo


codeboten:
**Describe the bug**
It would be great to have a CI check for links in this repository. See https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/workflows/check-links.yaml as an example

Anu-MT:
Hi,
I would like to work on this issue. Would you please assign this to me?

codeboten:
@Anu-MT all yours! Thanks

poorvasharma2501:
Hi ,
I would like to start working on it, could you please assign this to me?

RichardChukwu:
Hello @poorvasharma2501 whats your progress on this please, still working on it or should I take it up? @tylerbenson

RichardChukwu:
@tylerbenson submitted a PR for this, hoping to get feedback please. Thanks

RichardChukwu:
Linked in a new PR on this, the other one had alot of inadequacies @tylerbenson I'll appreciate feedback

Thank you

pragmaticivan:
I can pick this one

pragmaticivan:
Fixed here: https://github.com/open-telemetry/opentelemetry-lambda/pull/1721

PR #1558, opened 2024-10-16
add check-links to this repo #300

RichardChukwu:
feat #300 

- Added a new GitHub Actions workflow for checking markdown links
- Configured `check_links_config.json` to include patterns for ignoring certain links (e.g., localhost, example.com)
- Ensured the workflow is triggered on push and pull requests
- Tested the configuration locally and via GitHub Actions to verify correct link checking using README.md as a case study and all links were checked


tylerbenson (changes requested):
It looks like you're missing a git ignore.  There are a ton of files checked in that shouldn't be.  Please delete and ammend.

RichardChukwu:
On it @tylerbenson

PR #1559, opened 2024-10-17
check-links added to repo

RichardChukwu:
feat #300 

- Added a new GitHub Actions workflow for checking markdown links

- Configured check_links_config.json to include patterns for ignoring certain links (e.g., localhost, example.com)

- Ensured the workflow is triggered on push and pull requests

- Tested the configuration locally and via GitHub Actions to verify correct link checking using README.md as a case study and all links were checked


.github/workflows/check-links.yaml
```diff
@@ -0,0 +1,51 @@
...
+  push:
+    branches: [main]
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
```
tylerbenson: Why is this needed?

.github/workflows/check-links.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install markdown-link-check
+        run: npm install
+
+      - name: Run markdown-link-check
+        run: |
+          npx --no -- markdown-link-check \
+            --verbose \
+            --config .github/workflows/check_links_config.json \
+            ${{needs.changedfiles.outputs.md}} \
+            || { echo "Check that anchor links are lowercase"; exit 1; }
```
tylerbenson, line 46: Could you explain this line a bit?

.github/workflows/check-links.yaml
RichardChukwu: Thank you for the review @tylerbenson  

The `cancel-in-progress: true` was included to ensure that any currently running workflows from the same PR would be canceled when a new commit is pushed, thus avoiding redundant workflow runs. This helps streamline the CI process, but I realize that this may not be necessary in this context. 

I can remove it if it's not aligned with the project's workflow policies.

.github/workflows/check-links.yaml
RichardChukwu: Ok,  This line of code runs the markdown-link-check with verbose logging enabled and specifies the custom configuration file (check_links_config.json). The `needs.changedfiles.outputs.md` variable is used to dynamically target only the markdown files that were modified in the PR.

The || `{ echo "Check that anchor links are lowercase"; exit 1; } ` part is a fallback check to ensure that if there are any issues with the anchor links not being lowercase (which is a requirement for valid markdown links), the workflow will fail and print a meaningful message for the contributor to resolve. 

However, if this check is redundant, I can adjust or remove it.

.github/workflows/check-links.yaml
tylerbenson: Assuming this job runs quickly, which I assume it does, I don't think that's necessary.  Go ahead and remove.

.github/workflows/check-links.yaml
tylerbenson: Not objecting to the current implementation, but I wonder if maybe a linter would do a better job of that particular edge case.  What do you think?

.github/workflows/check-links.yaml
RichardChukwu: "Thank you for the feedback, removed now.

.github/workflows/check-links.yaml
RichardChukwu: I agree that a linter might be more appropriate for this edge case, as it would likely cover more general cases of markdown formatting and anchor validation. Not quite sure how to implement exactly for this as I'm new here. I'll appreciate any help @tylerbenson

tylerbenson: I don't have any specific insight here.  Maybe investigate what other OpenTelemetry repos are doing?

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/AwsSdkSample/obj/Debug/net6.0/.NETCoreApp,Version=v6.0.AssemblyAttributes.cs
```diff
@@ -0,0 +1,4 @@
+// <autogenerated />
```
tylerbenson: I think the changes to the `dotnet` folder need to be removed.

RichardChukwu: Ok. I'll do that

RichardChukwu: Done @tylerbenson any further changes?

tylerbenson: I meant the whole `dotnet` folder should not have any diffs...  I think I should only see changes in `.github/workflows`.

RichardChukwu: Yes, I did notice that those files somewhat gets automatically added after cloning the repo and running a build, hence why they are being commited as changes, not sure why though.

I want to try to reclone and commit the github workflow changes again however to see if it fixes that

RichardChukwu: Done, you should be seeing only the changes in the `.github/workflows` now @tylerbenson

dotnet/sample-apps/aws-sdk/wrapper/SampleApps/AwsSdkSample.sln
```diff
@@ -1,25 +0,0 @@
-﻿
-Microsoft Visual Studio Solution File, Format Version 12.00
```
tylerbenson: While I don't know why this file is checked in... I don't think this is the right PR to remove it.

RichardChukwu: Ok, I'll close this PR then to have a fresh start with new clone then

tylerbenson: That shouldn't be necessary.  Just undo deleting this file.

RichardChukwu: Ok, done

tylerbenson (approved):

tylerbenson:
@RichardChukwu I'm seeing an odd failure:
https://github.com/open-telemetry/opentelemetry-lambda/actions/runs/11840804028/job/32995521705?pr=1593

Would you mind taking a look?

RichardChukwu:
> @RichardChukwu I'm seeing an odd failure: https://github.com/open-telemetry/opentelemetry-lambda/actions/runs/11840804028/job/32995521705?pr=1593
> 
> Would you mind taking a look?

Hello @tylerbenson, seems really odd actually.

 From what I can see via the logs, seems like a package.json file seems non-existent and can't be read in the stated path `/home/runner/work/opentelemetry-lambda/opentelemetry-lambda/` thereby preventing `npm install `from running via checks.
 
 Checked the workflow file: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/.github/workflows/check-links.yaml  and it shouldn't be giving that error though

tylerbenson:
Ok, I'll ignore the error for now.  Please investigate.

tylerbenson:
We are still seeing errors here.  Can you please see what's wrong?

PR #1562, opened 2024-10-22
Upgrade common OTEL packages used by other OTEL packages to prevent having multiple versions of the same OTEL packages

serkan-ozal:
Without this fix, it is very likely to hit the package size limit which puts `250 MB` limit (unzipped) for the contents of a deployment package, including layers and custom runtimes. 

This fix also helps to reduce the coldstart delay by reducing the deployment artifact size.

Compressed Size
=====
**Before fix (`v0.10.0`):** `38.5 MB`
**After fix:** `8.5 MB`

`78%` package size improvement.

Uncompressed Size (On Disk)
=====
**Before fix (`v0.10.0`):** `202.1 MB`
**After fix:** `42.6 MB`

`79%` package size improvement.

### DIFF
--- a/nodejs/package-lock.json
+++ b/nodejs/package-lock.json
@@ -2374,9 +2374,9 @@
       }
     },
     "node_modules/@opentelemetry/resource-detector-aws": {
-      "version": "1.6.1",
-      "resolved": "https://registry.npmjs.org/@opentelemetry/resource-detector-aws/-/resource-detector-aws-1.6.1.tgz",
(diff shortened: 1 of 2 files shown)


pragmaticivan (approved):
LGTM

tylerbenson (approved):
Maybe we need a build assertion somewhere to verify this doesn't happen again?

serkan-ozal:
> Maybe we need a build assertion somewhere to verify this doesn't happen again?

@tylerbenson yes, right. I think, at least we can put a hard-coded compressed and un-compressed size limit (for ex. `10 MB` compressed size limit) check step in the Node.js layer release action after building but before publishing the layer.

rapphil (approved):

PR #1568, opened 2024-10-26
build(dev): update all nodejs packages

pragmaticivan:


tylerbenson (approved):
Thanks!

Why does this have to be so hard with JS.

pragmaticivan:
Lol @tylerbenson the more packages we have, the more people managing package lifecycle we need to deal with 😂

PR #1602, opened 2024-11-20
build: use esbuild to generate layer

pragmaticivan:
## Before:

Screenshot 2024-11-20 at 12 44 35 PM

Screenshot 2024-11-20 at 12 45 32 PM

## After:

Screenshot 2024-11-20 at 12 44 05 PM

Screenshot 2024-11-20 at 12 46 24 PM

### DIFF
--- a/nodejs/.gitignore
+++ b/nodejs/.gitignore
@@ -33,7 +33,6 @@ build/Release
 node_modules/
 jspm_packages/
 build/
-build_backup/
 
 # TypeScript v1 declaration files
 typings/
--- a/nodejs/.gitignore
+++ b/nodejs/.gitignore
@@ -33,6 +33,7 @@ build/Release
 node_modules/
 jspm_packages/
 build/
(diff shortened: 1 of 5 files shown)


pragmaticivan:
I might need some help testing this with an alpha layer for the esm module resolution

pragmaticivan:
It turns out that for ESM, will still need the `import-in-the-middle` and I the optional source maps are also missing here, although they are optional, so will only be included if the user pass a nodejs flag.

pragmaticivan:
Screenshot 2024-12-04 at 5 18 49 PM

New Layer Size: 

Some files are not included in the runtime.

pragmaticivan:
Closing this one for some live tests with ESM first

tylerbenson:
Worth revisiting?

pragmaticivan:
@serkan-ozal was workign on a different trial with webpack instead

PR #1603, opened 2024-11-21
honor initial set zap loglevel

prolane:
`os.Getenv("OPENTELEMETRY_EXTENSION_LOG_LEVEL")` returns an empty string when the env var is not set. `zap.ParseAtomicLevel("")` will return info log level.

This means  line `lvl := zap.NewAtomicLevelAt(zapcore.InfoLevel)` is never used.
This PR fixes this.

### DIFF
--- a/collector/main.go
+++ b/collector/main.go
@@ -18,8 +18,9 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"github.com/open-telemetry/opentelemetry-lambda/collector/lambdalifecycle"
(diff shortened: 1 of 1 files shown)


pragmaticivan:
The code LGTM. It indeed fixes the issue.

pragmaticivan (approved):

tylerbenson (approved):

serkan-ozal (approved):

PR #1630, opened 2024-12-18
fix: updated broken links

garysassano:
Docs were moved, so I updated the links to point to the new paths.


tylerbenson (approved):
Thanks!

PR #1653, opened 2025-01-10
feat(nodejs): add ability to be able to enable/disable active instrumentations by env var

serkan-ozal:


RaphaelManke:
I like the idea to give more control of what is loaded and what not. 

This change seems to introduce a configuration option (unless it's just something experimental/internal right now) and therefore should have a documentation section in the readme.

serkan-ozal:
These configurations are already in use ones as mentioned here: https://opentelemetry.io/docs/zero-code/js/configuration/#excluding-instrumentation-libraries

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -126,6 +91,109 @@ declare global {
   function configureInstrumentations(): Instrumentation[];
 }
 
+function getActiveInstumentations(): Set<string> {
+  let emabledInstrumentations: string[] = defaultInstrumentationList;
```
RaphaelManke:
```suggestion
  let enabledInstrumentations: string[] = defaultInstrumentationList;
```

serkan-ozal: fixed

tylerbenson:
It looks like Node already follows this convention, so that's fine.  Just pointing it out that Java does it a bit differently: https://opentelemetry.io/docs/zero-code/java/agent/disable/#enable-only-specific-instrumentation

tylerbenson (approved):
looks ok to me.

PR #1654, opened 2025-01-11
feat(nodejs:exporter): switch to `otlp-http` exporters for traces, metrics and logs ...

serkan-ozal:
… to get rid of initialization overhead of the protobuf modules during coldstart


tylerbenson (approved):
This sounds ok, but prob counts as a breaking change, so we'll need to document very clearly in the next release notes.

PR #1658, opened 2025-01-13
fix(extension): consider requests failed when status code is not 2xx

pkarakal:
This updates the extension api client to consider requests as failed only when their status code is not 2xx. This is mostly required since AWS Lambda has updated the status code returned on successful extension registration from 200 to 201 and thus all lambda invocations fail.

### DIFF
--- a/collector/internal/extensionapi/client.go
+++ b/collector/internal/extensionapi/client.go
@@ -180,7 +180,7 @@ func (e *Client) doRequest(req *http.Request, out interface{}) (*http.Response, 
 	if err != nil {
 		return nil, err
 	}
-	if resp.StatusCode != 200 {
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
 		return nil, fmt.Errorf("request failed with status %s", resp.Status)
 	}
 	defer resp.Body.Close()


tylerbenson (approved):
@rapphil @serkan-ozal does this look ok to you?

collector/internal/extensionapi/client.go
```diff
@@ -180,7 +180,7 @@ func (e *Client) doRequest(req *http.Request, out interface{}) (*http.Response,
 	if err != nil {
 		return nil, err
 	}
-	if resp.StatusCode != 200 {
+	if resp.StatusCode < 200 && resp.StatusCode > 299 {
```
tylerbenson: I don't recall if this has existing unit tests.  If it does, please consider adding a few test cases for this.

serkan-ozal: This must be `resp.StatusCode < 200 || resp.StatusCode > 299` (Replace `&&` with `||`). Because there is no number `< 200` and `> 299` at the same time.

pkarakal: you're completely right, made that change in a hurry :see_no_evil: 

I didn't find any test for the `extensionapi` package that's why I didn't add any. If you'd like I try adding tests to the package

tylerbenson: haha, oops... good catch @serkan-ozal.  I missed that too.

collector/internal/tools/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/open-telemetry/opentelemetry-lambda/collector/internal/tools
 
-go 1.19
+go 1.23
+
+toolchain go1.23.2
```
tylerbenson: This PR now has too many changes.  Please revert everything else.

pkarakal: dropped all irrelevant changes and rebased on top of latest main.

tylerbenson (approved):

collector/internal/extensionapi/client.go
tylerbenson: @serkan-ozal what do you think?  Would you like to see a test here?

serkan-ozal: @tylerbenson I am OK with the PR as is. Of course it would be better if there was a test for this PR. But, since we don't have any test infrastructure for the Lambda collector yet, I don't think missing test should be a blocker.

serkan-ozal (approved):

PR #1664, opened 2025-01-14
feat(nodejs): minimize out-of-the-box supported propagators to reduce coldstart delay

serkan-ozal:
and additionally, this PR
- drops `@opentelemetry/sdk-trace-node` to be not dependent to the `@opentelemetry/propagator-b3` and `@opentelemetry/propagator-jaeger` propagator packages.
- and so (as a result of dropping `@opentelemetry/sdk-trace-node`  dependency) expands the required types can be passed to the `configureTracerProvider` and `configureTracer` functions. Since this is type expansion, it should not break backward compatibility, because lower types (`NodeTraceProvider` and `NodeTracerConfig`) can still be passed to these functions as is.

### DIFF
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -16,15 +16,13 @@ import {
   W3CTraceContextPropagator,
 } from '@opentelemetry/core';
 import {
+  BasicTracerProvider,
   BatchSpanProcessor,
   ConsoleSpanExporter,
   SDKRegistrationConfig,
   SimpleSpanProcessor,
+  TracerConfig,
 } from '@opentelemetry/sdk-trace-base';
-import {
-  NodeTracerConfig,
-  NodeTracerProvider,
-} from '@opentelemetry/sdk-trace-node';
 import {
   MeterProvider,
   MeterProviderOptions,
@@ -60,6 +58,8 @@ import {
 import { AWSXRayPropagator } from '@opentelemetry/propagator-aws-xray';
 import { AWSXRayLambdaPropagator } from '@opentelemetry/propagator-aws-xray-lambda';
 
+import { LambdaTracerProvider } from './LambdaTracerProvider';
+
 const defaultInstrumentationList = [
   'dns',
   'express',
@@ -88,8 +88,8 @@ declare global {
   function configureAwsInstrumentation(
     defaultConfig: AwsSdkInstrumentationConfig,
   ): AwsSdkInstrumentationConfig;
-  function configureTracerProvider(tracerProvider: NodeTracerProvider): void;
-  function configureTracer(defaultConfig: NodeTracerConfig): NodeTracerConfig;
+  function configureTracerProvider(tracerProvider: BasicTracerProvider): void;
+  function configureTracer(defaultConfig: TracerConfig): TracerConfig;
   function configureSdkRegistration(
     defaultSdkRegistration: SDKRegistrationConfig,
   ): SDKRegistrationConfig;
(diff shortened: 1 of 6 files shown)


nodejs/packages/layer/src/wrapper.ts
```diff
@@ -212,6 +226,44 @@ function createInstrumentations() {
   ];
 }
 
+function getPropagator(): TextMapPropagator {
```
tylerbenson, line 229: Consider adding some unit tests.

serkan-ozal: @tylerbenson added tests, could you please check

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -212,6 +226,44 @@ function createInstrumentations() {
...
+  if (
+    process.env.OTEL_PROPAGATORS == null ||
+    process.env.OTEL_PROPAGATORS.trim() === ''
+  ) {
+    return new CompositePropagator({
+      propagators: [
```
pragmaticivan, line 235: Could you clarify why the default propagators will not match the ones listed here? https://github.com/open-telemetry/opentelemetry-lambda/blob/main/nodejs/packages/layer/scripts/otel-handler#L12

Also if the defaults are already set in otel-handler, is there any reason to also set it here? (I actually prefer that to be here instead)

serkan-ozal: @pragmaticivan Because this is the **existing** behavior. 

Currently, propagators are taken from [here](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/metapackages/auto-configuration-propagators/src/utils.ts#L51) in the `@opentelemetry/auto-configuration-propagators` package and hence I used the impl/logic there.

pragmaticivan: But the defaults of `this project` are `tracecontext,baggage,xray` instead. This env variable will always be overwritten by that file, so the code will never hit that block

serkan-ozal: - without this PR, default propagators (tracecontext,baggage,xray) are set by the `otel-handler` script
- with this PR, still default propagators (tracecontext,baggage,xray) are set by the `otel-handler` script

So, this keeps the existing behavior. No need to put the same default list into the newly introduced `getPropagator` method too. `otel-handler` script is responsible for setting the defaults.

serkan-ozal: I really don't understand the issue here. This PR keeps the existing behavior for propagators by default because it uses the existing impl code. Why it was OK before, and NOT OK now?

pragmaticivan: I guess you are trying to replicate the `other package behavior here` when you might not need this code block. All good, though. I understand that this was the default from the other package, yet this is a block of code that will `never` be able to be reached `in this package`, which makes it `dead code` out of the gate

serkan-ozal: `No one can guarantee that this is dead code` even though `otel-handler` is in place. 

Because users might put their own initializers and activate it through the `NODE_OPTIONS` env var. And in this case, their initializers will run before our wrapper. So, potentially, they can clear the `OTEL_PROPAGATORS` env var in their initializer code, hence wrapper should take care of empty `OTEL_PROPAGATORS` env var too.

pragmaticivan: Makes sense, I guess I see the point of this code, but don't see the point of keeping that code in otel-handler 😂

serkan-ozal: If you are still interested in removing that code block, you can send another PR for it and we can discuss. But the main point of this PR is **dropping** b3 and jaeger propagator dependencies **without** changing existing behavior.

tylerbenson (approved):

PR #1679, opened 2025-01-25
feat(nodejs): bundling with webpack

serkan-ozal:
With this PR, core (`wrapper.js` and most) of the OTEL AWS Lambda Node.js layer is bundled (tree-shaken and minified) with Webpack. So, this significantly reduces AWS Lambda coldstart delay of the OTEL AWS Lambda Node.js layer.

Additionally, with this PR, users can also disable metrics and logs by `OTEL_METRICS_EXPORTER=none` and `OTEL_LOGS_EXPORTER=none` to cut some more (`~30 ms`) coldstart delay if they don't need metrics and logs, but just traces.

- :white_check_mark: All tests are passing
- :white_check_mark: Tested with CJS and ESM based Lambda handlers

According to my experiments, 
- the coldstart delay reduces to `~220 ms` from `~550 ms`: **`60%` coldstart reduction** (`190 ms` if metrics and logs are disabled as described above)
- the Lambda layer artifact size reduces to `471 KB` from `8.9 MB`: **`95%` package size reduction**
 
There are still improvements we can do over this:
- There are improvements have been merged into the `opentelemetr-js` and `opentelemetry-js-contrib` repositories, but have not released and used in the Lambda layer yet.
- There are still pending improvement PRs to the `opentelemetr-js` and `opentelemetry-js-contrib` repositories waiting for review.
- Cache compiled V8 bytecode of the core (`wrapper.js`) and reuse it without compiling during coldstart.

Hopefully, with all these improvements, OTEL AWS Lambda Node.js layer coldstart overhead will be down to two digit ms (`< 100 ms`).

Besides that, we also need some improvements in the OTEL Lambda collector extension as it adds additional `270-280 ms` delay to the coldstart.


tylerbenson:
@pragmaticivan @martinkuba would you mind taking a look at this please?

pragmaticivan (approved):
Could not spot anything wrong/major and synced with the PR author in slack with some questions.

tylerbenson (approved):

serkan-ozal:
@tylerbenson @pragmaticivan I am done with this

nodejs/packages/layer/install-externals.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -euf -o pipefail
+
+rm -rf ./build/workspace/node_modules
+
+# Space separated list of external NPM packages
+EXTERNAL_PACKAGES=( "import-in-the-middle" )
```
rapphil, line 8: js newbie here :) 

Instead of having this separated script file with a list of external dependencies, would it work if we had a separated node package with package.json where you define all the dependencies. This would be a "noop package", but it would allow you to explicitly select what version of the external dependencies you want to use. and then you install it with just `npm install --prefix ./build/workspace --production --ignore-scripts`.

rapphil (approved):

nodejs/packages/layer/install-externals.sh
serkan-ozal: The reason why I have used bash script to install external packages is that I want to use the same version of `import-in-the-middle` coming from OTEL packages. So, I have to resolve its version dynamically. That is why I first resolve its version by npm query below and then install it with that specific version.

pragmaticivan (approved):

PR #1693, opened 2025-02-05
build(check-links): remove unnecessary npm install for check-links workflow

pragmaticivan:
This was added by accident, but if the workflow is using npx, it will download the dependency automatically. There is also no `package.json` at the root of the repository either, which is causing errors in some PRs.

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -1,4 +1,4 @@
-name: check-links
+name: "Link Check"
 on:
   push:
     branches: [main]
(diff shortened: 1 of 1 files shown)


.github/workflows/check-links.yaml
```diff
@@ -1,4 +1,4 @@
-name: check-links
+name: "Link Check"
```
pragmaticivan, line 1: FYI, also changed the name of the workflow because this one is the only workflow not following this format.

serkan-ozal (approved):

tylerbenson (approved):

serkan-ozal:
@pragmaticivan seems that still fails:
```
npm error npx canceled due to missing packages and no YES option: ["markdown-link-check@3.13.6"]
```

pragmaticivan:
Screenshot 2025-02-07 at 7 47 55 AM
https://www.npmjs.com/package/markdown-link-check @serkan-ozal 

Screenshot 2025-02-07 at 7 49 02 AM

PR #1694, opened 2025-02-05
docs(nodejs): improve nodejs doc and use node 22 for workflows

pragmaticivan:


serkan-ozal:
I don't think using node v22 is good idea especially for CI workflow. And even using v18 is not good. I believe that as long as we support v18, we should run build and tests in v18 along with other supported node versions (v20, v22).

The reasons is that, we might use an Node API or CLI flag which is 
- available in the earlier versions (for ex. v18), but removed in the later versions (for ex. v22)
- introduced in the newer versions (for ex. v22), but not available in the older versions (for ex. v18).

So, what I am suggesting is, at least for CI, we should use Github workflow job matrix and run the CI on all the supported Node versions (v18, v20, v22 as of now).

pragmaticivan:
> I don't think using node v22 is good idea especially for CI workflow. And even using v18 is not good. I believe that as long as we support v18, we should run build and tests in v18 along with other supported node versions (v20, v22).
> 
> The reasons is that, we might use an Node API or CLI flag which is
> 
> * available in the earlier versions (for ex. v18), but removed in the later versions (for ex. v22)
> * introduced in the newer versions (for ex. v22), but not available in the older versions (for ex. v18).
> 
> So, what I am suggesting is, at least for CI, we should use Github workflow job matrix and run the CI on all the supported Node versions (v18, v20, v22 as of now).

Matrix sounds good to me. I have to check some stuff before, but it appears the version is irrelevant in some cases because even on tests, it's being transpiled to a single target with typescript (es2017) + `"ts-node/register"`

serkan-ozal:
>  I have to check some stuff before, but it appears the version is irrelevant in some cases because even on tests, it's being transpiled to a single target with typescript (es2017) + "ts-node/register"

Not everything can be handled by transpilation. If you use a new API (available here https://nodejs.org/docs/latest-v22.x/api/, but not here https://nodejs.org/docs/latest-v18.x/api/) introduced in v22, since there may not be equivalent of it in the v18, the result js code will fail at runtime and these fails can only be caught by the tests running at v18. 

And additionally, as I have mentioned above, by mistake we might use a Node CLI option in the `otel-handler` script which is not supported in v18 and this will only fail at v20 and v22. 

So, in summary, such cases cannot be handled by just transpilation whatever you do. The only proper way of catching those cases is running the tests in all the supported node versions. That is why `opentelemetry-js` repository runs their tests in all the supported Node environments: https://github.com/open-telemetry/opentelemetry-js/blob/main/.github/workflows/unit-test.yml#L14

PR #1724, opened 2025-03-02
Feat/nodejs/add auto instrumentations

### LINKED ISSUE #1600: Unable to get logs shipped from lambda to Loki
**Describe the bug**
I am unable to get logs to ship using opentelemetry. I tried [pino http transport](https://github.com/pinojs/pino-opentelemetry-transport) but with no avail. 
Is there a way to configure opentelemetry to scrape stdout or lambda telemetry API?

**Steps to reproduce**
Configure a lambda with the instrumentation and collector layer. Log something in the lambda

Logs in Loki.

**What did you see instead?**
Nothing. No errors either.

Instrumentation layer 0.11
Collector 0.12

NodeJS

**Additional context**
Collector config:

```
receivers:
  otlp:
    protocols:
      http: 
  telemetryapi:
exporters:
  otlphttp:
    endpoint: 'https://tracing.internal.tools.company.io/otlp'
  otlphttp/logs:
    endpoint: 'https://loki.tools.company.io/otlp'
  prometheusremotewrite:
    endpoint: 'https://mimir.internal.tools.company.io/api/v1/push'
    resource_to_telemetry_conversion:
      enabled: true
    external_labels:
      source: aws-lambda
    remote_write_queue:
      enabled: true
      queue_size: 100000
      num_consumers: 50
  logging:

processors:
  coldstart:
  decouple:
  batch:
    timeout: 1m

service:
  pipelines:
    traces:
      receivers:
        - otlp
        - telemetryapi
      exporters:
        - otlphttp
      processors:
        - batch
        - decouple
        - coldstart
    metrics:
      receivers:
        - otlp
      exporters:
        - prometheusremotewrite
      processors:
        - batch
        - decouple
    logs:
      receivers:
        - otlp
        - telemetryapi
      exporters:
        - otlphttp/logs
      processors:
        - batch
        - decouple
```

I am getting lambda telemetry metadata in loki. Like faas.name and faas.max_memory hwever, the log line itself is empty. Is my collector config OK? 
What am I doing wrong with application logs?

### LINKED ISSUE #1611: [nodejs] Support for Pino instrumentation
Current implementation using global configureInstrumentations make it a bit hard to configure, especially in our use case where the lambda is bundled using cdk

I'd like to plug @opentelemetry/instrumentation-pino as part of the defaultConfigureInstrumentations

**Describe alternatives you've considered**
I tried to define defaultConfigureInstrumentations in a trace.js file and configure NODE_OPTIONS=--require /var/task/trace.js, but it does not play well if the lambda is packaged with cdk

**Additional context**
I am happy to write a PR

### LINKED ISSUE #1723: Match available instrumentation packages provided by the auto-instrumentations-node metapackage
All available packages should be included and configured via environment variables

No need to change defaults in this ticket, only add async/conditional support to all packages in `auto-instrumentations-node`

There will be an increase in the bundle size, but there should be no increase in latency.

wpessers:
This PR fixes #1723, fixes #1600, fixes #1611 
Added all the instrumentations that exist in the otel auto-instrumentations-node metapackage. Also removed most of the defaults. Right now only defaults (besides aws-lambda and aws sdk instrumentation) left are:
```
dns
http
net
```
I can see 2 possible changes to this:

1. Remove dns and net from the defaults. This was already discussed before in the community, and I personally also question them being enabled by default. I think opt-in for these may make more sense.
2. Add undici to defaults. I'm personally not 100% sure about this one. This would fix users not seeing traces when using nodejs native fetch. But then again, not everyone uses native fetch.

~As of now the PR is still a work in progress. I'm still working on a proposal to refactor the logic for the dynamic imports. As the wrapper code has gotten a bit bloated with all newly added instrumentations: [wrapper.ts](https://github.com/wpessers/opentelemetry-lambda/blob/feat/nodejs/add-auto-instrumentations/nodejs/packages/layer/src/wrapper.ts#L110-L307)~

Also, as this was discussed in the otel faas working group. The changes do indeed not noticably impact package size and cold starts. After adding the extra instrumentation packages, the package size increased by ~40KB. I ran some tests and noticed no difference in cold start times.


wpessers:
Separate comment for readability. After discussion with maintainers I've reverted the change where we alter the default instrumentations. This PR's purpose is only to support more out-of-the-box instrumentation and therefore we don't want it to be breaking. Removing defaults is, however, breaking so that will be a separate PR.

wpessers:
As requested I also ran some quick tests to see impact on the actual OTEL wrapper initialization times.
- In my testing, one lambda uses the `OTEL_NODE_ENABLED_INSTRUMENTATIONS` env var to only load the http instrumentation. I will refer to this lambda as the **light** lambda. 
- The other one does not set the env var and thus loads all instrumentations, I will refer to this lambda as the **heavy** lambda.

These are the results after running 40 times (data in ms):
| Statistic |light_old | light_new | heavy_old | heavy_new |
| -------- | -------- | -------- | -------- | -------- | 
| **Min** | 194 | 196 | 230 | 219 |
| **Max** | 260 | 265 | 358 | 341 |
| **Median** | 211 | 228 | 253 | 270.5 |
| **Mean** | 214.475 | 228.525 | 259.7 | 271 |

Here's the raw data: [otel_wrapper_init_times_ms.csv](https://github.com/user-attachments/files/19078641/otel_wrapper_init_times_ms.csv)

Sample size n=40 may not be a lot. But these results do look assuring that the changes have minimal to no impact on wrapper initialization time. If further testing is required, I can of course alter the parameters and/or sample size to get more data.

nodejs/packages/layer/package.json
```diff
@@ -44,21 +44,41 @@
     "@opentelemetry/exporter-metrics-otlp-http": "^0.57.0",
     "@opentelemetry/exporter-trace-otlp-http": "^0.57.0",
     "@opentelemetry/instrumentation": "^0.57.0",
+    "@opentelemetry/instrumentation-amqplib": "^0.46.1",
     "@opentelemetry/instrumentation-aws-lambda": "^0.50.0",
     "@opentelemetry/instrumentation-aws-sdk": "^0.49.0",
+    "@opentelemetry/instrumentation-bunyan": "^0.45.1",
+    "@opentelemetry/instrumentation-cassandra-driver": "^0.45.1",
+    "@opentelemetry/instrumentation-connect": "^0.43.1",
+    "@opentelemetry/instrumentation-cucumber": "^0.14.1",
```
serkan-ozal: I don't think we need to have `cucumber` instrumentation as it is for automated tests and nothing for Lambda.

wpessers: My bad, I looked through a couple of these packages but assumed we could use everything that was in the meta-package. Let me remove this one and double check all other libs to see if I missed any other irrelevant packages.

wpessers:
Fixed @serkan-ozal's comment about `cucumber` instrumentation. Also removed instrumentation for `generic-pool` and `lru-memoizer` as these don't make sense to use in lambda functions.

serkan-ozal (approved):

pragmaticivan (approved):
LGTM!

pragmaticivan:
Btw, it seems the failures in check-links are legit, so we probably need to fix these links later in a new PR. Not related to this PR though.

tylerbenson (approved):

PR #1734, opened 2025-03-04
feat(nodejs): remove non-generic instrumentations from defaults

### LINKED ISSUE #1735: node.js: Do not enable non-generic instrumentations by default
Currently the node.js instrumentation layer has a lot of [default instrumentations](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/nodejs/packages/layer/src/wrapper.ts#L50-L64). These default instrumentations are all being initialized when the `OTEL_NODE_ENABLED_INSTRUMENTATIONS` environment variable is not set. Meaning the wrapper will take longer to initialize due to a bunch of instrumentations being activated that most users probably don't even need.

An example: express, fastify, hapi and express instrumentations are all initialized by default. This does not make sense to me. Usually **if** people are using these sorts of frameworks they will stick to one. I also don't think we should be making the node instrumentation layer opinionated and that's why I would propose to remove all non-generic instrumentations from the defaults.

wpessers:
This PR reduces the amount of default instrumentations. These are the instrumentations that are initialized by default when a user does not set the `OTEL_NODE_ENABLED_INSTRUMENTATIONS` environment variable on their lambda function.

The current defaults include a lot of instrumentations that are not generic. e.g. the `express` instrumentation is only useful to users who use the `express.js` framework in their lambda, but is currently part of the defaults.

To start I have reduced the list of defaults to contain:
```
dns
http
net
```
Having `http` instrumentation seems like a sensible default, as it is used under the hood by many other packages.

What concerns `dns` and `net` instrumentation, I can see a good case for removing those from the defaults as well. 
- `dns`: Do we need dns lookup etc spans by default?
- `net`: Do we need tcp level spans by default?
I'm personally leaning towards removing these from defaults as well. Any input here is appreciated!

fixes #1735

### DIFF
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -47,11 +47,7 @@ import {
 import { AWSXRayPropagator } from '@opentelemetry/propagator-aws-xray';
 import { AWSXRayLambdaPropagator } from '@opentelemetry/propagator-aws-xray-lambda';
 
-const defaultInstrumentationList = [
-  'dns',
-  'http',
-  'net',
-];
+const defaultInstrumentationList = ['dns', 'http', 'net'];
 
 const propagatorMap = new Map<string, () => TextMapPropagator>([
   ['tracecontext', () => new W3CTraceContextPropagator()],
@@ -93,7 +89,7 @@ declare global {
   function configureLoggerProvider(loggerProvider: unknown): void;
 }
 
-function getActiveInstumentations(): Set<string> {
+function getActiveInstrumentations(): Set<string> {
   let enabledInstrumentations: string[] = defaultInstrumentationList;
   if (process.env.OTEL_NODE_ENABLED_INSTRUMENTATIONS) {
     enabledInstrumentations =
@@ -114,7 +110,7 @@ function getActiveInstumentations(): Set<string> {
 
 async function defaultConfigureInstrumentations() {
   const instrumentations = [];
-  const activeInstrumentations = getActiveInstumentations();
+  const activeInstrumentations = getActiveInstrumentations();
   if (activeInstrumentations.has('amqplib')) {
     const { AmqplibInstrumentation } = await import(
       '@opentelemetry/instrumentation-amqplib'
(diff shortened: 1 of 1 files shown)


pragmaticivan:
I'm honestly content with this list:

```
dns
http
net
```

While not extremely useful, DNS and Net might be relevant for some people and are generic enough that they might be worth keeping as default.

pragmaticivan:
@serkan-ozal @rapphil What are your thoughts on that one? Maybe something to bundle with Otel 2.0 for a `major version` release?

tylerbenson (approved):

serkan-ozal:
Yeah, let's keep this PR for the major release.

tylerbenson:
Are we are intending to "major release" the lambda layers? I'm fine if we decide to just stay at `0.x`.

pragmaticivan:
> Are we are intending to "major release" the lambda layers? I'm fine if we decide to just stay at `0.x`.

Not sure if we can do that on the AWS level (tag) but removing default values is usually a major change because there might be dashboards using that set of instrumentation that will stop working correctly.

pragmaticivan:
@tylerbenson @serkan-ozal have we reached consensus on what's the next step to get this one sorted?

tylerbenson:
I personally don't think we should bother with major versioning similar to how the collector has been operating (still `0.x`).

tylerbenson (approved):

wpessers:
I'll fix the lint issues on this one, whoops.

wpessers:
fixed the formatting of the instrumentation lib and also some random typo that suddenly caught my eye @open-telemetry/lambda-extension-approvers

tylerbenson (approved):

pragmaticivan (approved):

maxday:
LGTM!

serkan-ozal:
I am OK with the proposed list of instrumentations by default, but this seems a breaking change to me. 

For example, traces for a user Lambda function querying data over MySQL will not be available by default anymore, but they were before. 

Therefore, even though we are not bumping major version, I think, it should be clearly highlighted in the documents and release notes.

wpessers:
> I am OK with the proposed list of instrumentations by default, but this seems a breaking change to me.
> 
> For example, traces for a user Lambda function querying data over MySQL will not be available by default anymore, but they were before.
> 
> Therefore, even though we are not bumping major version, I think, it should be clearly highlighted in the documents and release notes.

Definitely agree!

PR #1751, opened 2025-03-25
Use Random port for listener in collector ext

vasireddy99:
Use Random port for listener in collector ext in order to avoid error `bind: address in use`

### DIFF
--- a/collector/internal/telemetryapi/listener.go
+++ b/collector/internal/telemetryapi/listener.go
@@ -17,18 +17,32 @@ package telemetryapi
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"math/rand"
+	"net"
 	"net/http"
 	"os"
+	"syscall"
 	"time"
 
 	"github.com/golang-collections/go-datastructures/queue"
 	"go.uber.org/zap"
 )
 
-const defaultListenerPort = "53612"
-const initialQueueSize = 5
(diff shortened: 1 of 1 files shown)


tylerbenson (approved):
I don't fully understand how the sender knows which port to send data to, but this seems reasonable.

serkan-ozal:
@vasireddy99 You also need to update receiver test too as it uses constant port: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/receiver/telemetryapireceiver/receiver_test.go#L119

serkan-ozal:
> I don't fully understand how the sender knows which port to send data to, but this seems reasonable.

@tylerbenson Its active port is registered here: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/internal/telemetryapi/client.go#L59

and it is called by here: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/internal/lifecycle/manager.go#L78

vasireddy99:
@serkan-ozal good point, I looked into it initially since we are not testing httprequest but only the handler, and  we are also using httptest it generates a valid http request irrespective of what we have provided, so probably be okay in my opinion.

serkan-ozal (approved):

serkan-ozal:
Thanks for the fix @vasireddy99

serkan-ozal (approved):

PR #1761, opened 2025-04-10
Upgrade eslint to be compatible with newer typescript versions

wpessers:
In PR #1750 there's some issues due to eslint-plugin's typescript version range. The outdated eslint-plugin limits typescript to <5.8


serkan-ozal (approved):

serkan-ozal:
Thanks for fixing this quickly.

PR #1789, opened 2025-04-23
Upgrade to otel sdk v2

wpessers:
~~Still work in progress. Most of it should be covered by this PR, but~~ still doing some smoke-testing on my own lambdas


nodejs/packages/layer/scripts/version.js
```diff
@@ -1 +0,0 @@
-console.log(require('@opentelemetry/core').VERSION);
```
serkan-ozal, line 1: Is `VERSION` no longer exported?

nodejs/packages/layer/package.json
```diff
@@ -33,56 +33,57 @@
...
-    "@opentelemetry/sdk-logs": "^0.57.0",
-    "@opentelemetry/sdk-metrics": "^1.30.0",
-    "@opentelemetry/sdk-trace-base": "^1.30.0"
+    "@opentelemetry/api-logs": "^0.200.0",
+    "@opentelemetry/context-async-hooks": "^2.0.0",
+    "@opentelemetry/core": "^2.0.0",
+    "@opentelemetry/exporter-logs-otlp-http": "^0.200.0",
+    "@opentelemetry/exporter-metrics-otlp-http": "^0.200.0",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.200.0",
+    "@opentelemetry/exporter-zipkin": "^2.0.0",
```
serkan-ozal: Why we need `exporter-zipkin` dependency?

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -359,32 +356,71 @@ function getPropagator(): TextMapPropagator {
...
+    process.env.OTEL_TRACES_EXPORTER == null ||
+    process.env.OTEL_TRACES_EXPORTER.trim() === '' ||
+    process.env.OTEL_TRACES_EXPORTER.includes('none')
+  ) {
+    return [new OTLPTraceExporter()];
+  }
+
+  const stringToExporter = new Map<string, () => SpanExporter>([
+    ['otlp', () => new OTLPTraceExporter()],
+    ['zipkin', () => new ZipkinExporter()],
```
serkan-ozal: Why we need to have and support Zipkin exporter by default?

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -359,32 +356,71 @@ function getPropagator(): TextMapPropagator {
...
+  resource: Resource,
 ): Promise<TracerProvider> {
   let config: TracerConfig = {
     resource,
+    spanProcessors: [],
   };
+
   if (typeof configureTracer === 'function') {
     config = configureTracer(config);
+  } else if (process.env.OTEL_TRACES_EXPORTER) {
```
serkan-ozal: I think, instead of `else if (process.env.OTEL_TRACES_EXPORTER)`, we can try to get exporters from env if no span processor is configured by the `configureTracer` function before. 

Because, even though `configureTracer` function is specified, user might not have interest to configure span processor (left its configuration to the Lambda layer here), but configure other properties of the `TracerConfig`.

And also, I think, `OTEL_TRACES_EXPORTER` env var should override the exporters set by the `configureTracer` function (for example, user is playing with different exporters without need to change the code and redeploy by just setting the env var.)

Therefore, I think, the following logic might be better:
- In the `getExportersFromEnv`, if there is no `OTEL_TRACES_EXPORTER` env var, return empty array.
- In the `getExportersFromEnv`, if `OTEL_TRACES_EXPORTER` is specified as `none`, return `null`.
- And here:
```js
// Try to get exporters from env var
const exporters = getExportersFromEnv();

if (!exporters) {
  // `null` exporters means tracing exporter is disabled, so skip trace provider initialization
  return;
}

// If there is specified `configureTracer` function by the user, use it to setup the config
if (typeof configureTracer === 'function') {
  config = configureTracer(config);
}

if (exporters.length) {
  // If there are exporters configured by env var, clear the ones configured before.
  // Because, the exporters configured by env var should override the ones configured by code. 
  config.spanProcessors = []; 
  exporters.map(exporter => {
    if (exporter instanceof ConsoleSpanExporter) {
      config.spanProcessors.push(new SimpleSpanProcessor(exporter));
    } else {
      config.spanProcessors.push(new BatchSpanProcessor(exporter));
    }
  });
}

config.spanProcessors = config.spanProcessors || [];

// If no exporter configured, use the default one
if (config.spanProcessors.length === 0) {
  // Default
  config.spanProcessors.push(
    new BatchSpanProcessor(new OTLPTraceExporter()),
  );
}

// Logging for debug
if (logLevel === DiagLogLevel.DEBUG) {
  config.spanProcessors.push(
    new SimpleSpanProcessor(new ConsoleSpanExporter()),
  );
}

const tracerProvider = new NodeTracerProvider(config);

...
```

nodejs/packages/layer/scripts/version.js
wpessers: yep, see upgrade guide: https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md?plain=1#L253

nodejs/packages/layer/package.json
wpessers: I added this because since the 2.0 upgrade, the tracing sdk does not support creation of exporters through env vars anymore. I then assumed that previously the zipkin exporter was supported by default because of the way it was listed here: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#exporter-selection
But now upon further inspection it does seem that this wasn't supported before this functionality was removed from the tracing sdk: https://github.com/open-telemetry/opentelemetry-js/pull/5239/files

So I will remove this, as we don't need to support it by default.

nodejs/packages/layer/src/wrapper.ts
wpessers: This makes sense to me, will fix

nodejs/packages/layer/src/wrapper.ts
wpessers: Removed this, same reason as stated above. OTEL docs specify that zipkin is a possible value for exporters set through env but it is not strictly something that is required. Also it wasn't supported previously in the tracing sdk, which we relied on for this "exporters from env vars" functionality before the 2.0 version.

nodejs/packages/layer/src/wrapper.ts
wpessers: Added the suggested changes in f2af371

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -359,32 +355,79 @@ function getPropagator(): TextMapPropagator {
...
+    return null;
+  }
+
+  const stringToExporter = new Map<string, () => SpanExporter>([
+    ['otlp', () => new OTLPTraceExporter()],
+    ['console', () => new ConsoleSpanExporter()],
+  ]);
+  const exporters: SpanExporter[] = [];
+  process.env.OTEL_TRACES_EXPORTER.split(',').map(exporterName => {
+    exporterName.toLowerCase().trim();
```
serkan-ozal:
```js
exporterName = exporterName.toLowerCase().trim();
```

rndD:
Hi @wpessers, any plans to merge it soon?

wpessers:
@rndD I will be doing some final testing tonight hopefully. Things should move quickly from there. But be aware that I can't guarantee when the next version for the layer will be released, this might take some time still after the PR has been merged.

rndD:
@wpessers, thanks for the detailed answer. 

Can you tell me the current cycle or strategy for layer releases?

wpessers:
> @wpessers, thanks for the detailed answer.
> 
> Can you tell me the current cycle or strategy for layer releases?

I will have to refer you to @serkan-ozal @tylerbenson @pragmaticivan for this question. As far as I'm aware there's no strict release cycle. It's just that when we are ready to release a certain change for a language specific layer, we want to make sure to also include "low-hanging fruit" (like dependency upgrades) for the different language-specific layers.

And if there happens to be some change in another layer that we want to release, this may hold back the entire release a little bit. So that we don't have to do multiple releases after one another in a short period of time.

nodejs/packages/layer/src/wrapper.ts
wpessers: Fixed so that exporterName now properly gets reassigned

serkan-ozal (approved):

Issue #1802, opened 2025-05-08
Update Go to version >=1.24.2 (CVE-2025-22871)


vw-anton:
Go versions before 1.24.2 contain a critical vulnerability which should be fixed in the published lambda layer for Go. This means as of now Go should be updated to 1.24.2. 

> The net/http package improperly accepts a bare LF as a line terminator in chunked data chunk-size lines. This can permit request smuggling if a net/http server is used in conjunction with a server that incorrectly accepts a bare LF as part of a chunk-ext.

https://pkg.go.dev/vuln/GO-2025-3563

maxday:
I think we can close this one @serkan-ozal  as https://github.com/open-telemetry/opentelemetry-lambda/pull/1812 got merged

serkan-ozal:
Fixed in #1812

PR #1812, opened 2025-05-14
use go version 1.24.2

biddyfiddy:
https://github.com/open-telemetry/opentelemetry-lambda/issues/1802


biddyfiddy:
bump

biddyfiddy:
@xuan-cao-swi is it possible to have this reviewed?

serkan-ozal:
Hi @biddyfiddy,

Sorry for late responses. Actually, there are other places in the collector (https://github.com/open-telemetry/opentelemetry-lambda/tree/main/collector) where we use Go version `1.23.1`. Should those versions need to be updated?

biddyfiddy:
> Hi @biddyfiddy,
> 
> Sorry for late responses. Actually, there are other places in the collector (https://github.com/open-telemetry/opentelemetry-lambda/tree/main/collector) where we use Go version `1.23.1`. Should those versions need to be updated?

thanks!  i updated the other go.mod files.

maxday:
idea: it could be great to automate this kind of bump thanks to a GitHub action? 
Unfortunately dependabot does not support go version upgrade in go.mod files (source: https://github.com/dependabot/dependabot-core/issues/9057) but it seems that this 3-rd party actions does that: https://github.com/030/gomod-go-version-updater-action. WDYT?

serkan-ozal:
Thanks @biddyfiddy, even though caret version is used, can you also update the version here: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/.github/workflows/release-layer-collector.yml#L36

biddyfiddy:
> idea: it could be great to automate this kind of bump thanks to a GitHub action? Unfortunately dependabot does not support go version upgrade in go.mod files (source: [dependabot/dependabot-core#9057](https://github.com/dependabot/dependabot-core/issues/9057)) but it seems that this 3-rd party actions does that: https://github.com/030/gomod-go-version-updater-action. WDYT?

This might be better as an improvement outside of this PR.  My intentions are to quickly bump the version so we can use it in our downstream dependencies.

biddyfiddy:
> Thanks @biddyfiddy, even though caret version is used, can you also update the version here: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/.github/workflows/release-layer-collector.yml#L36

done

maxday:
> This might be better as an improvement outside of this PR. My intentions are to quickly bump the version so we can use it in our downstream dependencies.

Sure! If that sounds good to you @serkan-ozal I can open an issue and draft a PR for this. The less manual work we need to do to stay evergreen the better is it IMO :)

serkan-ozal:
Hi @maxday, sure, sounds good. let's go ahead!

serkan-ozal (approved):

serkan-ozal:
Thanks for taking care of every comment @biddyfiddy

PR #1825, opened 2025-05-22
feat: use sha for actions

### LINKED ISSUE #1769: Use commit SHA for ALL github actions, except the ones provided by Github (actions/*)
For security reasons and to avoid situations like the ones in tj-actions, I would recommend we use commit SHA for most actions. Dependabot supports tag updates as a comment already and there should be no major issues on that support.

maxday:
Fixes https://github.com/open-telemetry/opentelemetry-lambda/issues/1769

The original issue mentioned that it should only apply to non github-vended actions but I think it's safer to apply it everywhere. For instance `actions/cache` has 100+ contributors, and not all of them are from the official GitHub org.
The sha was already set for some github-vended actions in some workflows but it was inconsistent, this PR adds shas for 100% of actions


pragmaticivan:
I'm ok with SHA for `all`, but there is a benefit when we use major versions from GH org (less PRs to review is one of them)

maxday:
I think dependabot is a great ally for this use case, but I'm ok to revert the change for GitHub-vended actions. Let me know!

serkan-ozal:
Actually, both approaches have their own pros and cons, but, yes, using SHA is the recommended approach by Github especially in terms of security. Also, pinning actions by SHA helps to have better OSSF score:
- https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions
- https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

So, I am personally OK with these changes. 
@tylerbenson WDYT?

pragmaticivan (approved):
LGTM

tylerbenson (approved):
Does dependabot maintain the version in comments? I appreciate that and hope it does. The main thing that I don't like about SHAs is the version number is opaque.

pragmaticivan:
> Does dependabot maintain the version in comments? I appreciate that and hope it does. The main thing that I don't like about SHAs is the version number is opaque.

Yes, dependabot will update the comments

serkan-ozal:
Yes, comments are updated too. So, merging this one too.

serkan-ozal (approved):

PR #1830, opened 2025-05-25
fix: nodejs vulnerabilities

maxday:
This PR fixes:

`GHSA-h5c3-5r3r-rr8q`
see: https://osv.dev/vulnerability/GHSA-h5c3-5r3r-rr8q
affected package: `@octokit/plugin-paginate-rest`

`GHSA-rmvr-2pp2-xj38`
https://osv.dev/vulnerability/GHSA-rmvr-2pp2-xj38
affected package: `@octokit/request`

`GHSA-xx4v-prfh-6cgc`
https://osv.dev/vulnerability/GHSA-xx4v-prfh-6cgc
affected package: `@octokit/request-error`

By running npm ls on those three packages, it appears that they're a dependencies of `lerna`
See the output of the three commands:

```
npm ls @octokit/plugin-paginate-rest --all

opentelemetry-lambda@0.0.0 /Users/maxday/git/opentelemetry-lambda/nodejs
└─┬ lerna@8.2.2
  └─┬ @octokit/rest@20.1.2
    └── @octokit/plugin-paginate-rest@11.4.4-cjs.2

npm ls @octokit/request --all

opentelemetry-lambda@0.0.0 /Users/maxday/git/opentelemetry-lambda/nodejs
└─┬ lerna@8.1.9
  └─┬ @octokit/rest@19.0.11
    └─┬ @octokit/core@4.2.4
      ├─┬ @octokit/graphql@5.0.6
      │ └── @octokit/request@6.2.8 deduped
      └── @octokit/request@6.2.8

npm ls @octokit/request-error --all

opentelemetry-lambda@0.0.0 /Users/maxday/git/opentelemetry-lambda/nodejs
└─┬ lerna@8.1.9
  └─┬ @octokit/rest@19.0.11
    └─┬ @octokit/core@4.2.4
      ├── @octokit/request-error@3.0.3
      └─┬ @octokit/request@6.2.8
        └── @octokit/request-error@3.0.3 deduped
```
I think dependabot might not being able to bump it because of the caret, this PR, in addition to bump the `lerna` version, removes the carret.

Those vulnerabilities are reported here: https://scorecard.dev/viewer/?uri=github.com/open-telemetry/opentelemetry-lambda


wpessers:
Upgrading lerna will indeed fix the vulnerabilities. But I'm wondering why dependabot did not pick this up, as I've seen it propose dependency upgrades for versions that use caret before. How certain are we that this will fix that? Anyways it is a devdependency so I'm not opposed to pinning this one.

maxday:
@wpessers interesting! You're absolutely right, the issue is that by default devDependencies are >not< updated by dependabot [(I've just discovered that I have the same issue in my repo)](https://github.com/maxday/lambda-perf/blame/main/result-builder/package.json), you can see that dependabot successfully upgraded dependencies but not dev-dependencies.

Fortunately, there is an option to fix this and allow dev dependencies upgrade: `dependency-type` https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#dependency-type-allow

I'll update the PR

serkan-ozal:
Actually, so far, there were PRs which dependabot updated dev dependencies without having `dependency-type=all`. Not sure why it didn't pick up this one: 
Here are just a few examples:
- https://github.com/open-telemetry/opentelemetry-lambda/pull/1835
- https://github.com/open-telemetry/opentelemetry-lambda/pull/1815
- https://github.com/open-telemetry/opentelemetry-lambda/pull/1809
- https://github.com/open-telemetry/opentelemetry-lambda/pull/1711

maxday:
ah that's interesting! Let me try to build a simple external repro to see what's happening and iterate with different dependabot config

wpessers:
whoops, I was typing out that review but in the meantime you've already responded to Serkan's comment I see. The thing about the versioning-strategy I linked above may be worth looking into still!

tylerbenson:
Is this still relevant?

pragmaticivan:
Can we close this PR @maxday ?

Issue #1838, opened 2025-05-31
Go: auto-bump go version


maxday:
In order to reduce maintenance effort (see: #1812), it would be great to see `go` version to be auto-bumped.

I think there are two sub tasks for this:

1. Remove hardcoded `go-version` in GitHub Actions worflows as it can lead to version mismatch, ie on main today: https://github.com/open-telemetry/opentelemetry-lambda/blob/3af180dc7f9dde4e06a939bd794b0430840dcddd/.github/workflows/ci-collector.yml#L27 and https://github.com/open-telemetry/opentelemetry-lambda/blob/3af180dc7f9dde4e06a939bd794b0430840dcddd/.github/workflows/ci-python.yml#L51 
2. Use a GitHub action to auto-bump go.mod go version. Unfortunately that's not yet supported by Dependabot (source: https://github.com/dependabot/dependabot-core/issues/9057)

I will tackle both of these
Thanks!

maxday:
1. Should be tackled by this: #1839

maxday:
now that 1. is merged, https://github.com/open-telemetry/opentelemetry-lambda/pull/1848 has been opened to tacked 2.

thompson-tomo:
Couldn't the solution be to switch to renovate like a large number of repos have already done?

wpessers:
It definitely could. But low maintainer capacity / low contributor activity means that's currently not being worked on. Feel free to open an issue and/or PR if you'd like to tackle that though.

thompson-tomo:
No worries pr for renovate config available via #2336

PR #1848, opened 2025-06-04
feat: go mod auto version update

maxday:
This is a follow-up PR on this: https://github.com/open-telemetry/opentelemetry-lambda/pull/1839 to complete this: https://github.com/open-telemetry/opentelemetry-lambda/issues/1838

I've tried it on my fork and the PR with a bump is indeed created see the example here: https://github.com/maxday/opentelemetry-lambda/pull/7

Quick notes:
- Since the PR is created from a worflow step, it does NOT auto trigger the collector check to run (This is a documented mechanism to increase security + to prevent loops)
- I've added a step to run the check but there is no easy way to report the status on the PR. I think I can make it work with custom checks but I think it's good enough like this). When a new PR like this is created, we would need to manually check the status of the triggered action, IMO, it's still a huge benefit to have this auto-bump which is already detecting that we're behind one minor version 😅.

Let me know!
Max

### DIFF
--- a/collector/receiver/telemetryapireceiver/receiver.go
+++ b/collector/receiver/telemetryapireceiver/receiver.go
@@ -321,6 +321,10 @@ func newTelemetryAPIReceiver(
 		r.Attributes().PutStr(semconv.AttributeServiceName, "unknown_service")
 	}
 
+	if val, ok := os.LookupEnv("OTEL_SERVICE_NAME"); ok {
+		r.Attributes().PutStr(semconv.AttributeServiceName, val)
+	}
+
 	for env, resourceAttribute := range envResourceMap {
 		if val, ok := os.LookupEnv(env); ok {
 			r.Attributes().PutStr(resourceAttribute, val)


.github/workflows/ci-gomod-auto-upgrade.yaml
```diff
@@ -0,0 +1,15 @@
...
+permissions:
+  contents: read
+jobs:
+  gomod-go-version-updater-action:
+    permissions:
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: 030/gomod-go-version-updater-action@7de28fa4093a877ba8924f7901e6c7b6e5dd3954 # v0.3.3
```
tylerbenson: Is this being used anywhere else within otel?

maxday: @tylerbenson no I could not find other usage. Maybe we can pioneer this and influence other go repos if this is working well for us?
if you look at the code, it's a very simple action: https://github.com/030/gomod-go-version-updater-action/blob/main/main.py

tylerbenson: @trask @carlosalberto @jaronoff97 @codeboten WDYT?  Does this need further review or good to go?

trask: from a security perspective, I personally avoid github actions that are not widely used, since it's one more attack vector you have to worry about

that said, I don't think we have any official guidance for OpenTelemetry repositories, and I suspect that we would first focus on limiting the github token permissions, since that's typically the easier win:

> An action can access the GITHUB_TOKEN through the github.token context even if the workflow does not explicitly pass the GITHUB_TOKEN to the action. As a good security practice, you should always make sure that actions only have the minimum access they require by limiting the permissions granted to the GITHUB_TOKEN. For more information, see [Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

(https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow)

In this particular case though, have you checked if Renovate will automatically update this for you? The majority of OpenTelemetry repos are on Renovate now, and I believe the Go repos were early Renovate adopters.

cc @open-telemetry/sig-security-approvers @open-telemetry/go-approvers

maxday:
oops sorry I removed the branch 😶‍🌫️ , will reopen

PR #1852, opened 2025-06-09
Fix telemetryapi not picking up service name

### LINKED ISSUE #1851: telemetryapi receiver not picking up service name label
**Describe the bug**
When using `telemetryapi`, the service name is always equal to the function name. `OTEL_SERVICE_NAME` is ignored.

**Steps to reproduce**
- Create and instrument a Lambda function
- Set `OTEL_SERVICE_NAME` environment variable
- Enable `telemetryapi` receiver for logs
- Send logs to your OTEL backend of choice and check the `service_name` label. It will be equal to the function name, not the value of `OTEL_SERVICE_NAME`.

Service name label should be equal to value of `OTEL_SERVICE_NAME`.

**What did you see instead?**
Service name label is Lambda function name

0.14.0 (collector layer)

JavaScript, Go

**Additional context**

tarek-sater-pf:
Fixes #1851

### DIFF
--- a/collector/receiver/telemetryapireceiver/receiver.go
+++ b/collector/receiver/telemetryapireceiver/receiver.go
@@ -321,6 +321,10 @@ func newTelemetryAPIReceiver(
 		r.Attributes().PutStr(semconv.AttributeServiceName, "unknown_service")
 	}
 
+	if val, ok := os.LookupEnv("OTEL_SERVICE_NAME"); ok {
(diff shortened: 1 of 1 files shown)


tylerbenson (commented):
Seems ok to me.  @open-telemetry/lambda-extension-approvers any objections?

serkan-ozal (approved):

PR #1861, opened 2025-06-10
fix: remove debug logs and use the diag logger for init timings

maxday:
As discussed during the SIG meeting, let's remove the `Registering OpenTelemetry` log as this is showing in plain text (not JSON formatted) 
see picture below:
Screenshot 2025-06-10 at 11 46 19 PM

The init duration information can definitely be useful so I propose to use the diag logger and set the level to `DEBUG`. This way a customer using `INFO` log, won't see any plain text logs.

### DIFF
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -610,7 +610,9 @@ export async function init() {
   initialized = true;
 }
 
-console.log('Registering OpenTelemetry');
+export function logDebug(message: string, ...args: unknown[]) {
+  diag.debug(message, ...args);
+}
 
 let initialized = false;
 let instrumentations: Instrumentation[];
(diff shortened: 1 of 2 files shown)


nodejs/packages/layer/src/wrapper.ts
```diff
@@ -610,7 +610,9 @@ export async function init() {
   initialized = true;
 }
 
-console.log('Registering OpenTelemetry');
+export function logDebug(message: string, ...args: unknown[]) {
+  diag.debug(message, ...args);
```
tylerbenson, line 614: Any reason this needs to be a separate function rather than calling `debug` directly from the callers?

tylerbenson (approved):

nodejs/packages/layer/src/wrapper.ts
serkan-ozal: When everything is bundled, any OTEL module (including `diag`) is not directly available to the outside of `wrapper`. It can be solved by configuring webpack to bundle `diag` out individually, but it will just complicate the things. So, I think, it is better to have `init.mjs` handling the things over `wrapper` instead of having directt dependency.

serkan-ozal (approved):

nodejs/packages/layer/src/wrapper.ts
maxday: @serkan-ozal exactly!

PR #1902, opened 2025-07-14
docs: fix links

trentm:
The links to opentelemetry-js-contrib.git changed due to
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2928

---

I tested locally with:

```
lychee --verbose --config .github/lychee.toml **/*.md
```

The other links (those *not* related to the opentelemetry-js-contrib.git change) were ... broken for a while I guess? I didn't dig into the history there.


wpessers:
Thank you for fixing, the collector links seem like they should've been broken for a while indeed. No idea why that never passed our radar before...

wpessers (approved):

tylerbenson (approved):

PR #1906, opened 2025-07-21
add invocation ids for plain text logs

adamantal:
In case the Lambda logs are of plain text format, they are not enriched with the `requestId` information ([link to API reference](https://docs.aws.amazon.com/lambda/latest/dg/telemetry-schema-reference.html#telemetry-api-function)).

I am well aware that logging in json format is preferred, but I think manually enriching the logs with this attribute can be beneficial in at least two cases:
- even if the application code logs in json format, static init/teardown of the function may be in plain text format - e.g. an entrypoint that prints out some message which still belongs to the same invocation
- legacy apps (boring, but also our use case)

Tested internally on some of our Lambdas


serkan-ozal (approved):

tylerbenson (approved):
Do you have a screenshot of what the logs before/after look like?

adamantal:
Sure, here's an example snippet of a Lambda invocation:
Screenshot 2025-07-31 at 16 54 09

You can see some [Magnum](https://github.com/Kludex/mangum) logs when the invocation starts.

The rest of the observability stack is quite simple honestly. We send logs to Loki's OTLP endpoint, and browse them through Grafana.

We have a few other changes in our internal fork that we'd like to contribute back over time. For example in the screenshot you can see the `service_instance_id` field. This is not exposed to the Lambda collector from the core Otel SDK by default, so we enrich it manually (which is the equivalent of AWS Cloudwatch log streams).

Let me know if you have any questions. Great project btw

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -217,16 +218,31 @@ func (r *telemetryAPIReceiver) createLogs(slice []event) (plog.Logs, error) {
...
 				// in plain text https://docs.aws.amazon.com/lambda/latest/dg/telemetry-schema-reference.html#telemetry-api-function
 				if line, ok := el.Record.(string); ok {
 					logRecord.Body().SetStr(line)
 				}
 			}
+		} else { // platform events, if subscribed to
+			if el.Type == string(telemetryapi.PlatformStart) {
+				if record, ok := el.Record.(map[string]interface{}); ok {
+					if requestId, ok := record["requestId"].(string); ok {
+						r.currentFaasInvocationID = requestId
```
pragmaticivan, line 240: Not as familiar with this code, but any reason you are not logging the request ID after this line?

adamantal: no special reason. I can add a debug level log to this, if it's helpful

wpessers (approved):

PR #1921, opened 2025-07-31
chore(nodejs): introduce `configureLogger` and deprecate `configureLoggerProvider`

### LINKED ISSUE #1905: Introduce Global `configureLogger` Function
As a follow up action of this PR https://github.com/open-telemetry/opentelemetry-lambda/pull/1903, we need to introduce global `configureLogger` function allow users customizing logger configuration instead of configuring over `configureLoggerProvider`. Because `configureLoggerProvider` doesn't allow adding `LogRecordProcessor`s over it in the latest versions of OpenTelemetry JS packages, but `LoggerConfig` does.

serkan-ozal:
Fixes: #1905

### DIFF
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -46,7 +46,6 @@ import {
 } from '@opentelemetry/instrumentation-aws-lambda';
 import { AWSXRayPropagator } from '@opentelemetry/propagator-aws-xray';
 import { AWSXRayLambdaPropagator } from '@opentelemetry/propagator-aws-xray-lambda';
-import { LogRecordProcessor } from '@opentelemetry/sdk-logs';
 
 const defaultInstrumentationList = [
   'dns',
@@ -88,10 +87,19 @@ declare global {
   // No explicit metric type here, but "unknown" type.
   // Because metric packages are important dynamically.
   function configureMeter(defaultConfig: unknown): unknown;
+  /**
+   * @deprecated please use {@link configureMeter} instead.
+   */
   function configureMeterProvider(meterProvider: unknown): void;
 
+  // No explicit logger type here, but "unknown" type.
+  // Because logger packages are important dynamically.
+  function configureLogger(defaultConfig: unknown): unknown;
   // No explicit log type here, but "unknown" type.
   // Because log packages are important dynamically.
+  /**
+   * @deprecated please use {@link configureLogger} instead.
+   */
   function configureLoggerProvider(loggerProvider: unknown): void;
 }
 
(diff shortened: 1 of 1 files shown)


tylerbenson (approved):

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -88,10 +87,19 @@ declare global {
...
   // Because metric packages are important dynamically.
   function configureMeter(defaultConfig: unknown): unknown;
+  /**
+   * @deprecated please use {@link configureMeter} instead.
+   */
   function configureMeterProvider(meterProvider: unknown): void;
 
+  // No explicit logger type here, but "unknown" type.
+  // Because logger packages are important dynamically.
+  function configureLogger(defaultConfig: unknown): unknown;
```
wpessers, line 97: Can we not make the return type `LoggerProviderConfig`? The `configureLogger` result is always used to instantiate  the `LoggerProvider`
```suggestion
  function configureLogger(defaultConfig: unknown): LoggerProviderConfig;
```

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -504,21 +512,27 @@ async function initializeLoggerProvider(
...
   // Logging for debug
   if (logLevel === DiagLogLevel.DEBUG) {
-    logRecordProcessors.push(
+    loggerConfig.processors.push(
       new SimpleLogRecordProcessor(new ConsoleLogRecordExporter()),
     );
   }
-  const loggerProvider = new LoggerProvider(loggerConfig);
+
+  const loggerProvider = new LoggerProvider(loggerConfig as object);
```
wpessers, line 535: Here we would no longer need to do the type assertion if we define the return type of `configureLogger` as suggested.

```suggestion
  const loggerProvider = new LoggerProvider(loggerConfig);
```

wpessers (commented):
One minor suggestion, rest looks good to me!

nodejs/packages/layer/src/wrapper.ts
serkan-ozal: @wpessers The reason is why I have used `unknown` type (instead of actual `LoggerProviderConfig` type) is preventing eager loading of the logger related modules in the webpack bundle (for ex. when `OTEL_LOGS_EXPORTER` env var is `none`) to reduce coldstart delay.

wpessers: Ah that makes sense. In that case the PR looks good to me.

wpessers (approved):

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -504,21 +512,27 @@ async function initializeLoggerProvider(
   );
 
   const logExporter = new OTLPLogExporter();
-  const logRecordProcessors: LogRecordProcessor[] = [];
-  const loggerConfig = {
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+  let loggerConfig: any = {
```
pragmaticivan, line 516: Any reason why you would not use https://github.com/mfernd/opentelemetry-js/blob/main/experimental/packages/sdk-logs/src/index.ts#L18 here?

pragmaticivan (approved):

PR #1940, opened 2025-08-22
fix: retain configured span processors

### LINKED ISSUE #1923: SpanProcessors defined by the custom trace config would be overriden if exports are set
https://github.com/open-telemetry/opentelemetry-lambda/blob/2945d3555cc8f9025059ac382cf71558ae738ea3/nodejs/packages/layer/src/wrapper.ts#L407

The line above simply overrides any span processors that could have been custom configured via ```config = configureTracer(config);``` just a few lines above in case there were exporters set as well. 

That means that a scenario in which I would want a custom span processor that would add some attributes to spans while also setting custom otlp exports is not possible.

artkhod:
Fixes #1923

Previously defining OTEL exporter would have removed any spanProcessors defined by configureTracer

### DIFF
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -412,7 +412,7 @@ async function initializeTracerProvider(
   }
 
   if (exporters.length) {
-    config.spanProcessors = [];
+    config.spanProcessors = config.spanProcessors || [];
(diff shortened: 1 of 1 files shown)


serkan-ozal:
@artkhod Thanks for the fix!

serkan-ozal (approved):

Issue #1963, opened 2025-09-15
Optional: Signing the Lambda Layers with AWS Code Signing


andyloughran:
**Is your feature request related to a problem? Please describe.**

https://github.com/open-telemetry/opentelemetry-lambda/blob/main/.github/workflows/publish-layer-collector.yml

Running the above workflow to produce private layers.

If a team is running an AWS CodeSigner, then the publish job should also include an optional capability to sign the layers to confirm the provenance.

https://docs.aws.amazon.com/lambda/latest/dg/governance-code-signing.html

**Describe the solution you'd like**

I'd like the GitHub action to include an artefact attestation for the zip binaries and each of the language-specific binaries released. 
https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations and couple that with AWS's codesigner for the zip files.

```
aws signer start-signing-job \
--source 's3={bucketName=xxxx,key=collector-layer.zip,version=xxxxx}' \
--destination 's3={bucketName=xxxx,prefix=signed/}' \
--profile-name SigningProfile_XXXXX
--client-request-token XXXUNIQUE_FROM_GITHUB_JOB
```

The additional parameter would need to be the SigningProfile, that would need to be present in the account that the GHA is uploading to.

Prerequisites

- The Github OIDC role would also need permission to be able to sign.
- The S3 bucket needs to have versioning enabled.

**Describe alternatives you've considered**

I think the GitHub attestation step is optional, but could allow extra confidence that the artefacts have not been tampered with, especially if the signing job connects directly to the GitHub attestation.

**Additional context**

I'm keen to raise a PR to demonstrate that this could be an optional extra; but wanted to raise the issue first as I couldn't find evidence of where this has been/not-been considered.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

tylerbenson:
This sounds like a reasonable approach.  Thanks for working on it!

PR #1964, opened 2025-09-15
Add build attestations to all uploaded/released artifacts.

andylockran:
As a step suggested in #1963 , add the GH attestations for binaries produced by the Github workflows on this repository, so that all artefacts generated as a result of the workflows have provenance checks.

It may be that more work needs to be done on deciding how to manage the lifecycle of historic attestations.

https://github.com/actions/attest-build-provenance

### DIFF
--- a/.github/workflows/publish-layer-collector.yml
+++ b/.github/workflows/publish-layer-collector.yml
@@ -54,7 +54,6 @@ on:
 
 permissions:
   contents: read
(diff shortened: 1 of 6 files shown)


tylerbenson:
@maxday could you take a look at this PR.  I'm curious if you think this would be a better solution than your proposal.

PR #1965, opened 2025-09-18
feat: update ruby layer that fix aws_lambda issue

xuan-cao-swi:
Update ruby layer to fix the instrumentation aws_lambda issue (see pr https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1645)


pragmaticivan (approved):
I confirmed the updated package is already supporting ruby 3.4

wpessers:
@xuan-cao-swi this looks good to me. Just want to check, have you built the ruby layer from source and tested this out as well? Also I'm not knowledgeable on ruby, but this PR also includes releasing the layer for ruby 3.4 runtime. Does anything else need to happen to support this. e.g. see: https://github.com/open-telemetry/opentelemetry-lambda/blob/14bd5bf410fa7ba19f08447447769fdfce9cb321/ruby/src/template.yml#L17

xuan-cao-swi:
@wpessers , yes, it works with 3.4 since last release (0.9.0), I just forgot to add the runtime in release-layer-ruby.yml. The template.yml is not used for build the layer; it's only for testing (or for user who like to use sam to build the layer.

wpessers (approved):

PR #1972, opened 2025-09-26
chore(actions): Upgrade java to version 17 for codeql workflow

wpessers:
The dependabot update for the spotless plugin in #1917 causes the codeql workflow to fail due to an incompatible java version. Since this new major 8.0.0 spotless requires java 17.


pragmaticivan (approved):
LGTM, same version used in the existing workflow https://github.com/open-telemetry/opentelemetry-lambda/blob/main/.github/workflows/ci-java.yml#L28-L31

tylerbenson:
Good find!

PR #1982, opened 2025-10-08
NodeJS: update mongodb instrumentation version to `0.59.0`

rndD:
There was an [issue](https://github.com/open-telemetry/opentelemetry-js-contrib/issues/2281) with the default `dbStatementSerializer` blocking us from using MongoDB instrumentation in lambdas. Since there is no way to set the serializer, we locked with the default one in the lambda OTEL layer.

It was fixed in https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3131 and released in https://github.com/open-telemetry/opentelemetry-js-contrib/releases/tag/instrumentation-mongodb-v0.59.0.

### DIFF
--- a/nodejs/package-lock.json
+++ b/nodejs/package-lock.json
@@ -43,9 +43,9 @@
       "license": "Apache-2.0"
     },
     "node_modules/@aws-cdk/cloud-assembly-schema": {
-      "version": "48.9.0",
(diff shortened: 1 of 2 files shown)


wpessers:
Thank you for informing us, this has been updated through dependabot pr #1983

PR #1901, opened 2025-07-14
feat: add transform processor as opt-in on custom build

petrusdemelo:
## Description

Add `transform processor` as opt-in for custom build using `lambdacomponents.processor.transform` tag.

PR #2004, opened 2025-10-24
feat: auto-inject AWS account ID into telemetry from Lambda extension registration

RaphaelManke:
## Summary

Automatically extract the AWS account ID from the Lambda Extensions API registration response and inject it as the `cloud.account.id` attribute into all telemetry (traces, logs, metrics) via a confmap converter.

This implementation uses the OpenTelemetry Collector's converter pattern to mutate the loaded configuration, ensuring the account ID is automatically available without requiring configuration or environment variables.

## Changes

- **extensionapi/client.go**: Added `AccountId` field to `RegisterResponse` and request the `accountId` feature via `Lambda-Extension-Accept-Feature` header
- **accountidprocessor/converter.go**: New confmap converter that automatically injects a resource processor for `cloud.account.id` attribute into all pipelines
- **accountidprocessor/converter_test.go**: Comprehensive tests for converter behavior (empty account ID, no pipelines, multiple pipelines, leading zeros)
- **extensionapi/client_test.go**: Tests for JSON unmarshaling with leading zero preservation
- **lambdacomponents/default.go**: Updated to accept accountID and return converter factories
- **collector.go**: Updated to accept and register custom converters
- **manager.go**: Updated to pass account ID through the initialization flow

## Benefits

- Account ID automatically available in all telemetry without configuration
- No environment variables needed, uses AWS Lambda API response
- Follows OpenTelemetry Collector patterns
- Static injection at startup with no runtime overhead
- Comprehensive test coverage for edge cases (leading zeros, multiple pipelines)

### DIFF
--- a/collector/internal/confmap/converter/accountidprocessor/converter.go
+++ b/collector/internal/confmap/converter/accountidprocessor/converter.go
@@ -1,0 +1,103 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The accountidprocessor implements the Converter for mutating Collector
+// configurations to automatically inject the cloud.account.id attribute
+// via a resource processor into all pipelines.
+package accountidprocessor
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/collector/confmap"
+)
+
+const (
+	serviceKey       = "service"
+	pipelinesKey     = "pipelines"
+	processorsKey    = "processors"
+	resourceProc     = "resource/aws-account-id"
+	accountIDAttrKey = "cloud.account.id"
+)
+
+type converter struct {
+	accountID string
+}
+
+// New returns a confmap.Converter that injects cloud.account.id into all pipelines
+func New(accountID string) confmap.Converter {
+	return &converter{accountID: accountID}
+}
+
+func (c converter) Convert(_ context.Context, conf *confmap.Conf) error {
+	if c.accountID == "" {
+		return nil // Skip if no account ID
+	}
+
+	// Navigate to service.pipelines
+	serviceVal := conf.Get(serviceKey)
+	service, ok := serviceVal.(map[string]interface{})
+	if !ok {
+		return nil
+	}
+
+	pipelinesVal, ok := service[pipelinesKey]
+	if !ok {
+		return nil
+	}
+
+	pipelines, ok := pipelinesVal.(map[string]interface{})
+	if !ok {
+		return nil
+	}
+
+	updates := make(map[string]interface{})
+
+	// For each pipeline, add resource processor to beginning
+	for telemetryType, pipelineVal := range pipelines {
+		pipeline, ok := pipelineVal.(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		processorsVal, _ := pipeline[processorsKey]
+		processors, ok := processorsVal.([]interface{})
+		if !ok {
+			processors = []interface{}{}
+		}
+
+		// Prepend resource/aws-account-id processor
+		processors = append([]interface{}{resourceProc}, processors...)
+		updates[fmt.Sprintf("%s::%s::%s::%s", serviceKey, pipelinesKey, telemetryType, processorsKey)] = processors
+	}
+
+	// Configure the resource processor with cloud.account.id attribute
+	updates[fmt.Sprintf("processors::%s::attributes", resourceProc)] = []map[string]interface{}{
+		{
+			"key":    accountIDAttrKey,
+			"value":  c.accountID,
+			"action": "insert",
+		},
+	}
+
+	// Apply all updates
+	if len(updates) > 0 {
+		if err := conf.Merge(confmap.NewFromStringMap(updates)); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
--- a/collector/internal/confmap/converter/accountidprocessor/converter.go
+++ b/collector/internal/confmap/converter/accountidprocessor/converter.go
@@ -19,17 +19,16 @@ package accountidprocessor
 
 import (
 	"context"
-	"fmt"
 
 	"go.opentelemetry.io/collector/confmap"
+	semconv "go.opentelemetry.io/otel/semconv/v1.27.0"
 )
 
 const (
-	serviceKey       = "service"
-	pipelinesKey     = "pipelines"
-	processorsKey    = "processors"
-	resourceProc     = "resource/aws-account-id"
-	accountIDAttrKey = "cloud.account.id"
+	serviceKey    = "service"
+	pipelinesKey  = "pipelines"
+	processorsKey = "processors"
+	resourceProc  = "resource/aws-account-id"
 )
 
 type converter struct {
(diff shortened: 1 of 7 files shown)


collector/internal/confmap/converter/accountidprocessor/converter.go
```diff
@@ -0,0 +1,103 @@
...
+
+		processorsVal, _ := pipeline[processorsKey]
+		processors, ok := processorsVal.([]interface{})
+		if !ok {
+			processors = []interface{}{}
+		}
+
+		// Prepend resource/aws-account-id processor
+		processors = append([]interface{}{resourceProc}, processors...)
+		updates[fmt.Sprintf("%s::%s::%s::%s", serviceKey, pipelinesKey, telemetryType, processorsKey)] = processors
```
wpessers: I'm by no means a golang / otel collector expert. But I compared the way this update is being done to how the converter for the decouple processor works. It's idempotent in a way, because there it is checked whether the processor should be added instead of blindly adding it: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/internal/confmap/converter/decoupleafterbatchconverter/converter.go#L82-L86 However I'm guessing that is less relevant here as we do not expect users to define an accountIdProcessor themselves? And instead we assume that this processor will always be added once to each pipeline?

collector/internal/extensionapi/client.go
```diff
@@ -65,9 +66,11 @@ const (
 )
 
 const (
-	extensionNameHeader      = "Lambda-Extension-Name"
-	extensionIdentiferHeader = "Lambda-Extension-Identifier"
-	extensionErrorType       = "Lambda-Extension-Function-Error-Type"
+	extensionNameHeader         = "Lambda-Extension-Name"
+	extensionIdentiferHeader    = "Lambda-Extension-Identifier"
+	extensionErrorType          = "Lambda-Extension-Function-Error-Type"
+	extensionAcceptFeatureHeader = "Lambda-Extension-Accept-Feature"
```
wpessers, line 72: nit, but can you please format this

collector/internal/confmap/converter/accountidprocessor/converter.go
```diff
@@ -0,0 +1,103 @@
...
+type converter struct {
+	accountID string
+}
+
+// New returns a confmap.Converter that injects cloud.account.id into all pipelines
+func New(accountID string) confmap.Converter {
+	return &converter{accountID: accountID}
+}
+
+func (c converter) Convert(_ context.Context, conf *confmap.Conf) error {
```
maxday: I think we can tweak this function a bit:
- use `(c* converter) in the definition
- simplify a bit the for loop 
- remove the usage of sprintf

This will give something like:
```go
func (c *converter) Convert(_ context.Context, conf *confmap.Conf) error {
	if c.accountID == "" {
		return nil
	}

	svc, ok := conf.Get(serviceKey).(map[string]interface{})
	if !ok {
		return nil
	}

	pipes, ok := svc[pipelinesKey].(map[string]interface{})
	if !ok {
		return nil
	}

	for _, v := range pipes {
		p, ok := v.(map[string]interface{})
		if !ok {
			continue
		}

		raw, _ := p[processorsKey]
		procs, _ := raw.([]interface{})
		p[processorsKey] = append([]interface{}{resourceProc}, procs...)
	}

	resourceProcessor := map[string]interface{}{
		resourceProc: map[string]interface{}{
			"attributes": []map[string]interface{}{
				{
					"key":    accountIDAttrKey,
					"value":  c.accountID,
					"action": "insert",
				},
			},
		},
	}

	return conf.Merge(confmap.NewFromStringMap(map[string]interface{}{
		"processors": resourceProcessor,
	}))
}
```
WDYT?

collector/internal/confmap/converter/accountidprocessor/converter.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"go.opentelemetry.io/collector/confmap"
+)
+
+const (
+	serviceKey       = "service"
+	pipelinesKey     = "pipelines"
+	processorsKey    = "processors"
+	resourceProc     = "resource/aws-account-id"
+	accountIDAttrKey = "cloud.account.id"
```
maxday: do you know if those strings already exists in a semver convention naming package? I'm not sure for other keys, but a quick search for this one: https://github.com/open-telemetry/opentelemetry-go/blob/a0a0acdceb0608265890ab6d983e9d1f7b73e735/semconv/v1.4.0/resource.go#L25 shows that we are already exporting this one, let's reuse it?

collector/internal/confmap/converter/accountidprocessor/converter.go
RaphaelManke: Thanks for the suggestion! I tried the in-place mutation approach, but it doesn't work because `confmap.Conf.Get()` and `ToStringMap()` both return deep copies, not references. I checked the `decoupleafterbatchconverter` in this repo and it uses the same `::` path notation with `conf.Merge()` pattern.

I've adopted your other suggestions:
- Pointer receiver `(c *converter)`
- Cleaner variable names and simplified type assertions
- Using semconv constant (`go.opentelemetry.io/otel/semconv` since the collector one is deprecated)
- Removed `fmt.Sprintf` (using string concatenation instead)
- Added idempotency check (from @wpessers' suggestion)

collector/internal/confmap/converter/accountidprocessor/converter.go
RaphaelManke: I added a idempotency check 😃

collector/internal/confmap/converter/accountidprocessor/converter.go
RaphaelManke: I updated in order to use the semconv package 😃

wpessers (changes requested):
Hi @RaphaelManke the current pr looks good to me, however it seems to break the process for doing custom builds of the collector (building with `-tags lambdacomponents.custom`).

I believe you'll have to also have to update the [custom implementation](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/lambdacomponents/custom.go#L32) to match what you did in the default one.

Keep in mind though that since you need the resource processor factory for your `accountidprocessor` to work and this is for the custom build, it's possible for the resource processor to not be there in someone's custom build. So you'll probably need some extra conditional check to only add the `accountidprocessor` if the resource processor is included.

If you can resolve that, LGTM!

RaphaelManke:
I see your point about the resource processor dependency—requiring it in all pipelines is not ideal and creates a coupling.

I'm exploring an alternative: after extension registration, we create a symlink at `/tmp/.otel-account-id` where the target is the account ID itself. Then, we extend the existing awsLambdaDetector in each language to read this symlink during resource detection. This would:
- Set cloud.account.id at the source (no processor needed)
- Work automatically for users with custom instrumentation
- Follow standard OTel resource detector patterns
The symlink approach is lightweight and avoids race conditions since SDK init happens well after extension registration.

It might be a longer way to go but it also would enable people who use their own instrumentations to leverage this solution. 
On top of that we don't have a dependency to the collector processors e.g. https://github.com/open-telemetry/opentelemetry-lambda/pull/1901 will add the transform processor. With that we might remove the resource processor in the future. 

@wpessers wdyt go with the dependency way to resource processor or the big run with adjusting the env detectors? 

---

## Current Implementation (Confmap Converter)

```mermaid
flowchart TB
    subgraph Extension["Extension (Go)"]
        E1[Register with Lambda API]
        E2[Receive accountId]
        E3[Pass to lambdacomponents.Components]
    end

    subgraph Converter["Confmap Converter"]
        C1["Injects resource/aws-account-id<br/>processor into ALL pipelines"]
    end

    subgraph SDK["SDK Telemetry"]
        S1["❌ No cloud.account.id<br/>in emitted telemetry"]
    end

    subgraph TelemetryAPI["Telemetry API Receiver"]
        T1["❌ No cloud.account.id<br/>in resource"]
    end

    subgraph Processor["resource/aws-account-id Processor"]
        P1["✅ Adds cloud.account.id<br/>to all data"]
    end

    Export[Export Data]

    E1 --> E2 --> E3 --> Converter
    SDK -->|OTLP| Processor
    TelemetryAPI --> Processor
    Processor --> Export

    style SDK fill:#ffcccc
    style TelemetryAPI fill:#ffcccc
    style Processor fill:#ffffcc
```

---

## Proposed Implementation (Symlink + Extended Resource Detector)

```mermaid
flowchart TB
    subgraph Extension["Extension (Go)"]
        E1[Register with Lambda API]
        E2[Receive accountId]
        E3["Create symlink:<br/>/tmp/.otel-account-id → accountId"]
        E4[Pass accountId to<br/>Telemetry API Factory]
    end

    subgraph SDK["SDK Telemetry"]
        subgraph Detectors["Resource Detectors"]
            D1["awsLambdaDetector<br/>✨ extended to read symlink"]
            D3[envDetector]
        end
        S1["✅ cloud.account.id<br/>in resource from start"]
    end

    subgraph TelemetryAPI["Telemetry API Receiver"]
        T1["✅ cloud.account.id<br/>added directly to resource"]
    end

    Export[Export Data<br/>No processor needed]

    E1 --> E2 --> E3
    E2 --> E4
    E3 -.->|readlink| D1
    E4 --> TelemetryAPI
    Detectors --> S1
    SDK -->|OTLP| Export
    TelemetryAPI --> Export

    style SDK fill:#ccffcc
    style TelemetryAPI fill:#ccffcc
    style D1 fill:#ccffff
```

---

## Sequence Diagram

```mermaid
sequenceDiagram
    participant Lambda
    participant Extension
    participant Symlink as /tmp/.otel-account-id
    participant SDK as SDK (Node/Python/Java/Ruby)
    participant TelemetryAPI as Telemetry API Receiver
    participant Collector

    Note over Lambda,Collector: Initialization Phase
    
    Lambda->>Extension: Start extension process
    Extension->>Lambda: Register (Accept-Feature: accountId)
    Lambda-->>Extension: Response (accountId: "123456789012")
    
    Extension->>Symlink: Create symlink → "123456789012"
    Extension->>TelemetryAPI: Pass accountId via factory
    
    Lambda->>SDK: Start runtime & wrapper
    
    Note over SDK: Resource Detection (awsLambdaDetector)
    SDK->>Symlink: readlink()
    Symlink-->>SDK: "123456789012"
    SDK->>SDK: Add cloud.account.id to resource
    
    Note over Lambda,Collector: Runtime Phase
    
    SDK->>Collector: OTLP (includes cloud.account.id ✅)
    TelemetryAPI->>Collector: Telemetry (includes cloud.account.id ✅)
    Collector->>Collector: Export (no processor needed)
```

---

## Comparison

| Aspect                     | Current (Confmap)                            | Proposed (Symlink + Detectors)         |
| -------------------------- | -------------------------------------------- | -------------------------------------- |
| **Where ID is added**      | Collector pipeline (processor)               | At the source (SDK + Receiver)         |
| **SDK telemetry**          | ❌ No `cloud.account.id` until processed      | ✅ Has `cloud.account.id` from emission |
| **Telemetry API data**     | ❌ No `cloud.account.id` in receiver resource | ✅ Has `cloud.account.id` directly      |
| **Processor dependency**   | ⚠️ Required in ALL pipelines                 | ✅ Not needed                           |
| **IPC mechanism**          | None (Go-internal)                           | Symlink              |
| **Custom instrumentation** | ❌ Must configure processor manually          | ✅ Automatic via extended detector      |
| **OTel patterns**          | ⚠️ Non-standard (config mutation)            | ✅ Standard resource detector extension |

wpessers:
@RaphaelManke your proposed solution using the symlink makes sense to me and in my opinion is the way to go here. I see a lot of thought went into it and agree with the advantages of that approach. If you are willing to implement that solution I'll happily review the work!

RaphaelManke:
In this PR and all the linked ones is a first draft. I have to go through all the different repos and need to make sure they full the respective ci pipelines. 

https://github.com/open-telemetry/opentelemetry-lambda/pull/2127

RaphaelManke:
We might can avoid this "injecting" since https://github.com/open-telemetry/opentelemetry-lambda/pull/2157 sets the account id for the sdks. 

In order to make it available to compiled sdk like go and sdk I would rather extend the [Resource Detection Processor
](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/internal/aws/lambda) 
for Lambda to read the symlink. 

When https://github.com/open-telemetry/opentelemetry-lambda/issues/2047 adds the resourcedetectionprocessor the account id would be made available via this one. 

This would make it clearer and less hacky, compoared to this PR.

Issue #1803, opened 2025-05-09
NodeJS: Add option to disable AWS Lambda auto instrumentation.


pseiferth:
**Is your feature request related to a problem? Please describe.**
I want to instrument a lambda function that processes a batch of SQS messages. Each SQS message contains a W3C tracing context in some custom fields and I extract them manually for each message and updating the Otel Context. Hence, I am not really interested in the span that is automatically created on each invocation of the Lambda. I would like to disable those, but there is no option to do so. Using the AWSLambdaInstrumentation seems to be hardcoded and not configurable via `OTEL_NODE_ENABLED_INSTRUMENTATIONS` / `OTEL_NODE_DISABLED_INSTRUMENTATIONS`. I still want to use the full NodeJS lambda layer since it handles all the remaining auto-instrumentation, initialization and flushing properly. 

**Describe the solution you'd like**
Setting `OTEL_NODE_DISABLED_INSTRUMENTATIONS="aws-lambda"` should disable the AWS lambda auto instrumentation and not create a span for each invocation of the Lambda. 

**Describe alternatives you've considered**
Not using the NodeJS layer, only the Otel Collector layer, and then initializing Opentelemetry manually. But this get quite tricky if you want to use auto instrumentation, especially with the aws-sdk and bundling everything using CDK.

**Additional context**
I think this might be the problem: https://github.com/open-telemetry/opentelemetry-lambda/blob/0b2e5647698582c481c5cd1dc1cc9cc3618e947c/nodejs/packages/layer/src/wrapper.ts#L309

PR #2012, opened 2025-11-03
fix(nodejs): make aws-lambda and aws-sdk instrumentations respect OTEL_NODE_DISABLED_INSTRUMENTATIONS

DivMode:
## Problem

DynamoDB Streams and other AWS event sources lack metadata carriers for trace propagation. Unlike SQS (message attributes) or API Gateway (HTTP headers), DynamoDB Streams only contain the changed record data.

### The Trace Propagation Issue

When using OpenTelemetry with DynamoDB Streams:

1. **Producer** writes to DynamoDB with trace context stored in record:
   ```json
   {
     "data": "data",
     "_otel": {
       "traceparent": "00-7d7836d35ab7da657d35c2b5f889c3be-de0931d210e214b3-01"
     }
   }
   ```

2. **Consumer Lambda** needs to extract and restore this context

3. **Problem:** `AwsLambdaInstrumentation` creates automatic wrapper span with NEW trace ID **before** handler code runs:
   ```
   Trace A: 7d7836d35ab7da657d35c2b5f889c3be (original fetch)
   Trace B: 49792b11678269a91638c0717450480b (NEW trace - disconnected!)
   ```

4. **Result:** Distributed trace is broken - cannot correlate stream processing with original operation

## Solution

This PR extends the functionality added in the base implementation to make `aws-lambda` and `aws-sdk` instrumentations respect `OTEL_NODE_DISABLED_INSTRUMENTATIONS`.

Previously, PR #1653 added the ability to disable instrumentations via `OTEL_NODE_DISABLED_INSTRUMENTATIONS`, but the AWS Lambda and AWS SDK instrumentations were always loaded unconditionally in the `createInstrumentations()` function. This meant users couldn't disable these instrumentations even when setting the environment variable.

**Usage:**
```bash
export OTEL_NODE_DISABLED_INSTRUMENTATIONS=aws-lambda
```

**Effect:** Disables automatic Lambda wrapper span, allowing handler to manually restore trace context from event data before creating spans.

## Use Case: DynamoDB Streams Trace Propagation

**Producer (stores trace context):**
```typescript
await dynamodb.put({
  Item: {
    data: "data",
    _otel: {
      traceparent: currentTraceContext.traceparent
    }
  }
});
```

**Consumer (restores trace context with disabled AwsLambdaInstrumentation):**
```typescript
// Lambda layer: OTEL_NODE_DISABLED_INSTRUMENTATIONS=aws-lambda
// No automatic wrapper span created!

export const handler = async (event: DynamoDBStreamEvent) => {
  const record = event.Records[0];
  const otelContext = unmarshall(record.dynamodb.NewImage)._otel;

  // Manually extract trace context
  const carrier = { traceparent: otelContext.traceparent };
  const extractedContext = propagation.extract(ROOT_CONTEXT, carrier);

  // Execute in extracted context (continues ORIGINAL trace!)
  await context.with(extractedContext, async () => {
    await OTELSpan.withSpan("process stream record", async () => {
      // All operations here inherit original trace ID
      await processRecord(record);
    });
  });
};
```

**Result:** Full trace continuity from producer → DynamoDB → stream → consumer

## Related Issues

- **Implements:** [#1803 - NodeJS: Add option to disable AWS Lambda auto instrumentation](https://github.com/open-telemetry/opentelemetry-lambda/issues/1803)

## Event Sources Affected

This enables manual trace propagation for AWS event sources without metadata carriers:

| Event Source | Metadata Carrier | Auto-Extracted | Needs Manual Propagation |
|--------------|------------------|----------------|--------------------------|
| API Gateway | ✅ HTTP headers | ✅ YES | ❌ NO |
| SQS | ✅ Message attributes | ✅ YES | ❌ NO |
| SNS | ✅ Message attributes | ✅ YES | ❌ NO |
| EventBridge | ✅ Event metadata | ✅ YES | ❌ NO |
| **DynamoDB Streams** | ❌ **None** | ❌ NO | ✅ **YES (must store in record data)** |
| **Kinesis Streams** | ❌ **None** | ❌ NO | ✅ **YES (must store in payload)** |

## Industry Pattern

This approach aligns with industry best practices for DynamoDB Streams trace propagation:

- **Datadog** (2024): "Distributed Tracing with Amazon DynamoDB" - Stores trace metadata in DynamoDB attributes, extracts in consumer
- **SigNoz** (2024): "OpenTelemetry Context Propagation" - Manual inject/extract for message queues without auto-propagation
- **Better Stack** (2023): "Implementing Distributed Tracing with OpenTelemetry" - Producer stores traceparent, consumer extracts and applies

## Testing

- [x] Verified `aws-lambda` and `aws-sdk` added to `defaultInstrumentationList`
- [x] Verified conditional loading based on `getActiveInstumentations()`
- [x] Tested in development environment with `OTEL_NODE_DISABLED_INSTRUMENTATIONS=aws-lambda`
- [x] Confirmed trace propagation works for DynamoDB Streams in production

## Benefits

1. ✅ Enables distributed tracing across DynamoDB Streams (previously impossible)
2. ✅ Provides control over when Lambda instrumentation runs (user choice)
3. ✅ Maintains all other auto-instrumentations (AWS SDK, HTTP, net/TLS)
4. ✅ No performance overhead (conditional loading, not runtime checks)
5. ✅ Follows OpenTelemetry context propagation best practices
6. ✅ Implements feature requested in issue #1803

### DIFF
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -48,6 +48,8 @@ import { AWSXRayPropagator } from '@opentelemetry/propagator-aws-xray';
 import { AWSXRayLambdaPropagator } from '@opentelemetry/propagator-aws-xray-lambda';
 
 const defaultInstrumentationList = [
+  'aws-lambda',
+  'aws-sdk',
   'dns',
   'express',
   'graphql',
@@ -309,21 +311,39 @@ async function defaultConfigureInstrumentations() {
 }
 
 async function createInstrumentations() {
-  return [
-    new AwsInstrumentation(
-      typeof configureAwsInstrumentation === 'function'
-        ? configureAwsInstrumentation({ suppressInternalInstrumentation: true })
-        : { suppressInternalInstrumentation: true },
-    ),
-    new AwsLambdaInstrumentation(
-      typeof configureLambdaInstrumentation === 'function'
-        ? configureLambdaInstrumentation({})
-        : {},
-    ),
+  const activeInstrumentations = getActiveInstumentations();
+  const instrumentations = [];
+
+  // Conditionally load AWS SDK instrumentation
+  if (activeInstrumentations.has('aws-sdk')) {
+    instrumentations.push(
+      new AwsInstrumentation(
+        typeof configureAwsInstrumentation === 'function'
+          ? configureAwsInstrumentation({ suppressInternalInstrumentation: true })
+          : { suppressInternalInstrumentation: true },
+      )
+    );
+  }
+
+  // Conditionally load AWS Lambda instrumentation
+  if (activeInstrumentations.has('aws-lambda')) {
+    instrumentations.push(
+      new AwsLambdaInstrumentation(
+        typeof configureLambdaInstrumentation === 'function'
+          ? configureLambdaInstrumentation({})
+          : {},
+      )
+    );
+  }
+
+  // Load other instrumentations
+  instrumentations.push(
     ...(typeof configureInstrumentations === 'function'
       ? configureInstrumentations()
-      : await defaultConfigureInstrumentations()),
-  ];
+      : await defaultConfigureInstrumentations())
+  );
+
+  return instrumentations;
 }
 
 function getPropagator(): TextMapPropagator {
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -319,9 +319,11 @@ async function createInstrumentations() {
     instrumentations.push(
       new AwsInstrumentation(
         typeof configureAwsInstrumentation === 'function'
-          ? configureAwsInstrumentation({ suppressInternalInstrumentation: true })
+          ? configureAwsInstrumentation({
+              suppressInternalInstrumentation: true,
+            })
           : { suppressInternalInstrumentation: true },
-      )
+      ),
     );
   }
 
@@ -332,7 +334,7 @@ async function createInstrumentations() {
         typeof configureLambdaInstrumentation === 'function'
           ? configureLambdaInstrumentation({})
           : {},
-      )
+      ),
     );
   }
 
@@ -340,7 +342,7 @@ async function createInstrumentations() {
   instrumentations.push(
     ...(typeof configureInstrumentations === 'function'
       ? configureInstrumentations()
-      : await defaultConfigureInstrumentations())
+      : await defaultConfigureInstrumentations()),
   );
 
   return instrumentations;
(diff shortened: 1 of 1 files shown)


serkan-ozal:
Hi @DivMode,

Could you please fix the lint issues.

wpessers:
Hi @DivMode, in general this looks good to me. Thank you for your effort. I notice you've also made some changes to the MeterProvider config, but didn't mention the reasoning behind that change.

nodejs/packages/layer/src/wrapper.ts
```diff
@@ -465,13 +487,26 @@ async function initializeMeterProvider(
...
+
+  // Respect OTEL_METRIC_EXPORT_INTERVAL for Lambda serverless environments
+  // Default 60s is too long for Lambda (functions finish in ~5s and freeze)
+  const exportIntervalMillis = process.env.OTEL_METRIC_EXPORT_INTERVAL
+    ? parseInt(process.env.OTEL_METRIC_EXPORT_INTERVAL, 10)
+    : 60000;
+
+  const exportTimeoutMillis = process.env.OTEL_METRIC_EXPORT_TIMEOUT
+    ? parseInt(process.env.OTEL_METRIC_EXPORT_TIMEOUT, 10)
+    : 30000;
```
wpessers, line 501: If the desire here is to fallback to the defaults provided in the [metrics-sdk](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/sdk-metrics/src/export/PeriodicExportingMetricReader.ts#L95-L96) I'd prefer leaving these options undefined when the environment variables are not set. Otherwise we're just doing the same defaulting logic. I'm open to other default suggestions though.

wpessers (changes requested):
Also this is changing some fundamental behaviour documented here: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/nodejs/README.md?plain=1#L10-L11
So that readme should be updated as well to reflect the change.

wpessers:
@serkan-ozal we also need to consider, this is a breaking change. Because, if current nodejs layer users are already using the `OTEL_NODE_ENABLED_INSTRUMENTATIONS`, they are most likely not adding `aws-sdk` and `aws-lambda` in there. Because the current default behaviour, described in https://github.com/open-telemetry/opentelemetry-lambda/blob/main/nodejs/README.md?plain=1#L10-L11 states that these will **always** be loaded. This change will force these users to add `aws-sdk` and/or `aws-lambda` if they need that instrumentation.

serkan-ozal:
To me, the real problem here is the customization of the trace context propagation. So, instead of allowing users to disable AWS Lambda handler instrumentation, introducing (and/or documenting) a way to be able to customize trace context extraction might be better way.

serkan-ozal:
@DivMode I think @wpessers makes a good point. This is a breaking change and I think it is better having AWS Lambda and SDK instrumentations active even though they are not explicitly mentioned in the `OTEL_NODE_ENABLED_INSTRUMENTATIONS` config. They should be only disabled explicitly by `OTEL_NODE_DISABLED_INSTRUMENTATIONS` config.

wpessers:
Good point on the customization of the propagation @serkan-ozal , I think it should be possible by customizing the lambda instrumentation `eventContextExtractor`: https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-aws-lambda#aws-lambda-instrumentation-options 
In the nodejs layer we offer the globals, allowing you to configure some stuff yourself: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/nodejs/packages/layer/src/wrapper.ts#L73-L104

I haven't played around with that but I believe if you use some sort of preload script that provides this function in the global scope and then use `NODE_OPTIONS` to run this before anything else, you can probably achieve what is needed.

Issue #1740, opened 2025-03-12
Trying to Bind to Random Port and Failing


swcloudgenie:
**Describe the bug**
During some cold start invocations, we see this error coming from the collector layer:

```json
{
    "level": "error",
    "ts": 1741771824.3853145,
    "logger": "telemetryAPI.Listener",
    "msg": "Unexpected stop on HTTP Server",
    "error": "listen tcp 169.254.100.6:53612: bind: address already in use"
}
```

Our configuration is as follows:

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: localhost:4317
      http:
        endpoint: localhost:4318
exporters:
  otlp/groundcover:
    endpoint: ****REDACTED****
    headers:
      apikey: ****REDACTED****
processors:
  batch:
    send_batch_size: 20
  decouple:
    max_queue_size: 100
  resource/groundcover:
    attributes:
      - action: insert
        key: env_name
        value: prod
service:
  pipelines:
    traces:
      receivers:
        - otlp
      processors:
        - resource/groundcover
        - batch
        - decouple
      exporters:
        - otlp/groundcover
    metrics:
      receivers:
        - otlp
      processors:
        - resource/groundcover
        - batch
        - decouple
      exporters:
        - otlp/groundcover
    logs:
      receivers:
        - otlp
      processors:
        - resource/groundcover
        - batch
        - decouple
      exporters:
        - otlp/groundcover
```

**Steps to reproduce**
This seems to happen intermittently, maybe once or twice a day for high volume functions.

**What did you expect to see?**
We would not expect to bind to 169.254.100.6:53612 since that IP/Port does not exist in our configuration.

**What did you see instead?**
Collector tried to bind to 169.254.100.6:53612 and failed

**What version of collector/language SDK version did you use?**
Version: 
arn:aws:lambda:us-east-2:184161586896:layer:opentelemetry-collector-arm64-0_12_0:1

**What language layer did you use?**
Config: NodeJS

**Additional context**
Also use the NodeJS layer: arn:aws:lambda:us-east-2:184161586896:layer:opentelemetry-nodejs-0_11_0:1

swcloudgenie:
I did manage to find where this port is being used: https://github.com/open-telemetry/opentelemetry-lambda/blob/3fef084a780311c22a80460401b7283caa0ea97e/collector/internal/telemetryapi/listener.go#L49

[...]

PR #2023, opened 2025-11-15
improve telemetry api initialization and tests

herin049:
A few months ago my team began noticing occasional `bind: address already in use` errors showing up on cold starts when using the Lambda collector layer (see #1740). We were able to come up with a fix internally, but since then it looks like the issue has been mostly solved by #1751. However, from a code maintainability/personal taste perspective the fix implemented is not ideal since it addresses the issue naively by running a brute force search over the ephemeral port range. A more optional solution is to just bind to port `0` which will allow the OS to automatically assign an available ephemeral port (see https://man7.org/linux/man-pages/man7/ip.7.html) This PR switches to using this approach and adds a basic set of tests for the telemetry API listener. Any additional comments are welcome as always.

### DIFF
--- a/collector/internal/telemetryapi/listener.go
+++ b/collector/internal/telemetryapi/listener.go
@@ -17,14 +17,11 @@ package telemetryapi
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
-	"math/rand"
 	"net"
 	"net/http"
 	"os"
-	"syscall"
 	"time"
 
 	"github.com/golang-collections/go-datastructures/queue"
@@ -33,16 +30,7 @@ import (
 
 const (
 	initialQueueSize = 5
-	maxRetries       = 5
-	// Define ephemeral port range (typical range is 49152-65535)
-	minPort = 49152
-	maxPort = 65535
 )
-
-// getRandomPort returns a random port number within the ephemeral range
-func getRandomPort() string {
-	return fmt.Sprintf("%d", rand.Intn(maxPort-minPort)+minPort)
-}
 
 // Listener is used to listen to the Telemetry API
 type Listener struct {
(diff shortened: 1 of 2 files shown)


wpessers:
Thank you for the work @herin049, I'm definitely more in favour of this approach, getting rid of the random port selection. Have you been running your version of the collector in a production environment?

collector/internal/telemetryapi/listener.go
```diff
@@ -60,46 +48,39 @@ func NewListener(logger *zap.Logger) *Listener {
...
 	s.logger.Info("Listening for requests", zap.String("address", address))
-	s.httpServer = &http.Server{Addr: address}
-	http.HandleFunc("/", s.httpHandler)
+	mux := http.NewServeMux()
+	s.httpServer = &http.Server{
+		Addr:    address,
+		Handler: mux,
+	}
+	mux.HandleFunc("/", s.httpHandler)
+
```
wpessers, line 83: Just to make sure, as I'm quite a novice at go, this change is not really required for this PR right? But I assume you are making this change since it is more idiomatic to not add a handler on the default servemux? Or does this offer any other benefits w.r.t performance / safety?

herin049: I initially added the new serve mux in order for `go test -race -timeout 120s ./internal/telemetryapi` to work properly. Otherwise, you will end up getting panics since the `/` route will be defined multiple times. Regardless, from my understanding it is generally considered best practice to avoid using the default serve mux for security reasons (since any Go module can mutate the default serve mux) and to avoid the usage of global state (which helps with testing among other things).

herin049:
> Thank you for the work @herin049, I'm definitely more in favour of this approach, getting rid of the random port selection. Have you been running your version of the collector in a production environment?

Thank you for the review @wpessers!

Unfortunately, we have not been running these change in production, but we have ran them in our lower environments to validate that the changes fix the original bug. 

After a little bit of digging, it seems like other Lambda extension authors opt to use this approach as well, in particular the NewRelic extension: https://github.com/newrelic/newrelic-lambda-extension/blob/2a5c012f5082d2f8f056252e87d2d3eecad5479a/lambda/logserver/logserver.go#L270 
so personally, I view these changes as safe to make. 

Unfortunately, AWS does not seem to give any guidance on if it is preferable to bind to port `0` when registering telemetry listeners. In my mind this is the only viable approach especially considering the fact that multiple listeners may be created by multiple extensions.

collector/internal/telemetryapi/listener_test.go
```diff
@@ -0,0 +1,340 @@
...
+			require.NoError(t, resp.Body.Close())
+
+			deadline := time.Now().Add(1 * time.Second)
+			for time.Now().Before(deadline) {
+				queueLen := listener.queue.Len()
+				if queueLen == int64(test.expectedCount) {
+					return
+				}
+				time.Sleep(10 * time.Millisecond)
+			}
```
wpessers: Not a big fan of this "polling" inside the test. I've done some researching and I believe you should be able to use something like https://pkg.go.dev/github.com/stretchr/testify/require#Eventually instead. We already have this dependency on `stretchr/testify` anyway

collector/internal/telemetryapi/listener_test.go
```diff
@@ -0,0 +1,340 @@
...
+	invalidJSON := []byte(`{"invalid": json}`)
+	resp, err := http.Post(address, "application/json", bytes.NewReader(invalidJSON))
+	require.NoError(t, err, "Failed to post invalid JSON: %v", err)
+	require.NoError(t, resp.Body.Close(), "Failed to close response body")
+
+	time.Sleep(50 * time.Millisecond)
+	require.Equal(t, listener.queue.Len(), int64(0), "Queue should be empty after invalid JSON")
+}
+
+func TestListener_Wait_Success(t *testing.T) {
```
wpessers, line 270: Can you explain what exactly this testcase is supposed to test? If I understand correctly the `Wait()` function is some sort of synchronization mechanism, pausing the collector until `platofrm.runtimeDone` event is received. But wouldn't it make sense with how the lambda "lifecycle" works that the `Wait()` function is called before those events are actually sent to the listener address?

collector/internal/telemetryapi/listener_test.go
```diff
@@ -0,0 +1,340 @@
...
+		{
+			name:          "empty events array",
+			events:        []Event{},
+			expectedCount: 0,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			listener.queue.Dispose()
```
wpessers: My thinking is possibly flawed since you mentioned earlier that you ran the tests with the `-race` flag. But anyway, to me this looks like a possibly flaky setup that may cause a race condition. At the start of each test here the `listener.queue` is being modified, but the `httpHandler` reads `s.queue.Len()` for the debug log at https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/internal/telemetryapi/listener.go#L138 **after** it has finished putting the events on the queue.

In the test here you explicitly wait until the queue contains the expected amount of events before asserting and then the next  loop starts. Once again modifying the queue, while the `httpHandler` may be trying to access it to log that debug log, no?

wpessers (changes requested):

wpessers:
@herin049 The listener code looks good to me, just some comments / questions I'd like answered before moving forward with this one. Again disclaimer that I'm far from an expert at go, also the reason why it took me some time to get back to you.

serkan-ozal (approved):

serkan-ozal:
I am OK with the changes here

collector/internal/telemetryapi/listener_test.go
herin049: My initial reasoning was that at at the start of each run, there shouldn't be any events being processed by the listener (i.e. the listener is in an idle state), so mutating the queue should be fine. Regardless, I agree that this setup is a little ugly, I'll rewrite this test case.

collector/internal/telemetryapi/listener_test.go
herin049: Good point, looking back at this test there is indeed a race condition that could make this test flaky if the call to `http.Post()` is executed before the listener is able to start, potentially leading to an error being returned. I also agree that to properly test this scenario, we should probably start the listener, call `listener.Wait()` in another goroutine, verify the the goroutine is still blocked before any events are sent, then proceed to send a few events while also verifying the goroutine is still blocked and then finally submit the platform runtime done event and verify the goroutine is unblocked.

herin049:
@wpessers I've address all your original comments on the unit tests and made some additional improvements.

wpessers (approved):
Adressed concerns wherever relevant, looks good to me!

wpessers:
Thank you for contributing @herin049 🚀

PR #2011, opened 2025-10-29
feat(receiver): aws-like report log for invocations

adamantal:
## Description

AWS Lambda users can see the following report in AWS Cloudwatch after the invocation has finished:
```
REPORT RequestId: bf5f58db-ef7d-4cb2-b3c6-266eba1bbedb Duration: 395.99 ms Billed Duration: 662 ms Memory Size: 512 MB Max Memory Used: 301 MB Init Duration: 265.72 ms
```
This pull request adds the ability to display these report and send them along with the regular Lambda application logs. This feature is hidden behind a flag (`log_report`), which is turned off by default.

Contrary to what someone could expect [based on the documentation](https://docs.aws.amazon.com/lambda/latest/dg/telemetry-schema-reference.html#ReportMetrics), these metrics are all floats, attempt to parse them as integers would fail.

## Testing

Added unit test coverage, and tested in a production environment. 
Attaching a Grafana log query for evidence (we're using Otel -> Alloy -> Loki -> Grafana pipeline)

Collector config extract:
```yaml
receivers:
  telemetryapi:
    types: [platform, function]
    log_report: true

extensions:
  basicauth/otlp:
    ...

exporters:
  otlphttp:
    auth:
      authenticator: basicauth/otlp
    endpoint: ...

processors:
  batch: {}
  decouple: {}

service:
  pipelines:
    logs:
      receivers: [telemetryapi]
      processors: [batch, decouple]
      exporters: [otlphttp]
  extensions: [basicauth/otlp]
```

Screenshot 2025-10-29 at 15 35 14

PR #2024, opened 2025-11-18
feat(collector/receiver): expose service.instance.id

adamantal:
### Description

The `service.instance.id` identifier is unique to a component (a Lambda execution), and it may have several advantages exposing this information:
- due to its manageable cardinality, it is a better candidate for indexing (as opposed to Lambda execution ids, that can be of much bigger magnitude)
- debugging caching issues across consecutive Lambda invocation on the same instance, as the opentelemetry-lambda's collector is persisted across executions

### Testing

Tested live on a Lambda function, on top of https://github.com/open-telemetry/opentelemetry-lambda/pull/2011:
Screenshot 2025-11-18 at 15 53 43


wpessers:
I'm not entirely sure this change is really adding something new. There is already a `faas.instance` resource attribute which I believe should be handled by the aws resource detector (in language specific layers like node.js). This refers to the cloudwatch log stream id, so it has the same cardinality as your proposed `service_instance_id`.

wpessers:
Upon closer inspection, it looks like this may only be supported in node.js and java

wpessers:
So in my opinion, this feature is already solved in a different way. I believe otel already supports aws lambda resource detection for most languages, which should include the `faas.instance`. This uniquely identifies a lambda "container" through the log stream id. Some examples below:
- JS: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/resource-detector-aws/src/detectors/AwsLambdaDetector.ts#L72
- Python: https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/resource/_lambda.py
- Java: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/aws-resources/README.md
- Go: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/detectors/aws/lambda

@tylerbenson @serkan-ozal any opinion on this?

adamantal:
In our use case the (Python) Lambda code itself is **not** instrumented with opentelemetry, but the opentelemetry-collector-lambda is running alongside the execution (as a Lambda extension). The logs that we receive lack the attribute you referenced above. 

This PR adds this ability to enrich the logs implemented at collector-side (so we don't have to instrument the Python code just to collect the logs that would otherwise be available through the Telemetry API anyways).

adamantal:
I'll take a deeper look to the attributes using the debug exporter to see all metadata, I might have missed something

serkan-ozal:
What I am unable to see here is that who puts AWS CloudWatch log stream id into the `receiver.Settings` as service instance it? cc @adamantal

adamantal:
Sorry for the delay, I caught up with other issues.

To give you more information I managed to test this internally on a staging environment using a `debug` exporter. An extract of a sampled log record:
```
ResourceLog #0
Resource SchemaURL: 
Resource attributes:
     -> faas.invoked_provider: Str(aws)
     -> service.name: Str([REDACTED])
     -> faas.name: Str([REDACTED])
     -> service.instance.id: Str(4879ee99-4c85-4aee-a0f1-9c89989439df)
     -> faas.max_memory: Str(2048)
     -> faas.version: Str($LATEST)
     -> faas.invoked_region: Str(us-east-2)
ScopeLogs #0
ScopeLogs SchemaURL: 
InstrumentationScope github.com/open-telemetry/opentelemetry-lambda/collector/receiver/telemetryapi 
LogRecord #0
ObservedTimestamp: 2025-12-09 15:04:07.609211443 +0000 UTC
Timestamp: 2025-12-09 15:04:07.141 +0000 UTC
SeverityText: 
SeverityNumber: Unspecified(0)
Body: Str(REPORT RequestId: d1acf096-5d04-4a32-99b7-58a059aa69dc Duration: 163.18 ms Billed Duration: 164 ms Memory Size: 2048 MB Max Memory Used: 224 MB)
Attributes:
     -> faas.invocation_id: Str(d1acf096-5d04-4a32-99b7-58a059aa69dc)
     -> type: Str(platform.report)
Trace ID: 
Span ID: 
Flags: 0
```

As you can see above, there's a `faas.invocation_id` that connects logs belonging to the same invocation, but there's no id to connect invocations that belongs to the same execution environment. Hence the `service.instance.id` introduced by this PR.

Answering your question, the `service.instance.id` itself is generally not the same as the cloudwatch log stream id. The cloudwatch log stream is exposed to the Lambda execution through an environment variable which is not available for the collector, I don't see an easy way to match those two ids. My point is that we don't even need to do that, just link the consecutive Lambda executions within the same execution environment with any (arbitrary) unique id, where `service.instance.id` is sufficient for this purpose.

serkan-ozal:
@adamantal actually I was asking who sets the `service.instance.id` attribute in the `receiver.Settings`. I have checked the code in Lambda collector but couldn't see any place where the `service.instance.id` attribute is set into the `receiver.Settings`, so you're able to get it by `set.Resource.Attributes().Get(semconv.AttributeServiceInstanceID)`

Is it set in the collector core at in the upstream repo? Is this the place: https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/internal/resource/config.go#L32

adamantal:
oh sorry, I misunderstood.

yes, indeed, it is autogenerated in the code you provided. I was unsure if I break anything by manually setting it, so I'd rather not override the value, just use the autogenerated one.

serkan-ozal (approved):

serkan-ozal:
Hi @adamantal, there is test failure after merge from main. Could you please check?

adamantal:
right, the import changed in #2066, hopefully fixed it 🤞

serkan-ozal:
Thanks for quick fix @adamantal

wpessers (approved):

Issue #2034, opened 2025-11-26
Node.js autoinstrumentation not working in Node.js 24 Lambda runtime


garysassano:
After upgrading to the new Node.js 24 Lambda runtime (announced [here](https://aws.amazon.com/blogs/compute/node-js-24-runtime-now-available-in-aws-lambda/)), OpenTelemetry autoinstrumentation has completely stopped working. The Lambda functions execute successfully, but no telemetry data is being collected or exported.

## Environment

- **Lambda Runtime**: `nodejs24.x`
- **ADOT Lambda Layer**: `arn:aws:lambda:eu-central-1:901920570463:layer:aws-otel-nodejs-arm64-ver-1-30-2:1`

## Expected Behavior

OpenTelemetry autoinstrumentation should continue to work with the Node.js 24 runtime, collecting and exporting traces, metrics, and logs as it did with previous Node.js runtimes (18.x, 20.x, 22.x).

## Actual Behavior

After upgrading to `nodejs24.x`:
- Lambda functions execute without errors
- No telemetry data is collected
- No traces appear in configured backends
- Autoinstrumentation appears to be completely inactive

## Steps to Reproduce

1. Create a Lambda function with Node.js 24 runtime (`nodejs24.x`)
2. Attach the OpenTelemetry Lambda layer
3. Configure the required environment variables for autoinstrumentation
4. Invoke the Lambda function
5. Check the telemetry backend for traces/metrics - none appear

It's possible that changes in the runtime environment are causing compatibility issues with the current autoinstrumentation approach.

## Possible Causes

- Changes in Lambda initialization lifecycle
- Module resolution changes in Node.js 24
- AWS Lambda wrapper changes
- Environment variable handling differences

## Workaround

Currently no known workaround. Staying on Node.js 22.x runtime maintains autoinstrumentation functionality.

wpessers:
Hi @garysassano, you say you are using ADOT layers. Which is a separate project: https://github.com/aws-observability/aws-otel-lambda

garysassano:
@wpessers I switched from ADOT to OTel and was able to reproduce the bug, which confirms the issue originates upstream.

Specifically, I switched from:
- `arn:aws:lambda:eu-central-1:901920570463:layer:aws-otel-nodejs-arm64-ver-1-30-2:1`

To:
- `arn:aws:lambda:eu-central-1:184161586896:layer:opentelemetry-nodejs-0_17_0:1`
[...]

PR #2046, opened 2025-11-29
fix(nodejs): register import in the middle for nodejs24.x runtime

wpessers:
solves #2034 

This change has been tested and fixes aws-lambda instrumentation again for commonjs lambda handlers using node 24 runtime.

### DIFF
--- a/nodejs/packages/layer/src/loader.mjs
+++ b/nodejs/packages/layer/src/loader.mjs
@@ -76,7 +76,16 @@ export function registerLoader() {
   }
 }
 
-if (_isHandlerAnESModule()) {
+function _getNodeRuntimeVersion() {
+  const executionEnv = process.env.AWS_EXECUTION_ENV || '';
+  const match = executionEnv.match(/nodejs(\d+)\.x/);
+  if (!match) {
+    return undefined;
+  }
+  return Number(match[1]);
+}
+
+if (_isHandlerAnESModule() || _getNodeRuntimeVersion() >= 24) {
   /*
   We could activate ESM loader hook of the "import-in-the-middle" library,
   - by "--loader=import-in-the-middle/hook.mjs" Node CLI option, but "--loader" option has been deprecated


wpessers:
@serkan-ozal @pragmaticivan any thoughts about this? Also tried quickly to add a handler test like the ones we already have. But is difficult to add something "lightweight" since `import-in-the-middle` is loaded already in other test and you can not really "unload" it. So not so easy to isolate from the others.

If wanted, I can setup a simple unit test for the loader script. First thought to spy on the `registerLoader()` method but since it's esm you can't spy on it with sinon. So that would require either an extra depedency to do some sort of mocking, or a refactor of the script...

pragmaticivan:
Honestly, the change LGTM, it's simple enough, maybe worth including a comment pointing to the change in the AWS blog post. But other than that, LGTM.

wpessers:
@pragmaticivan That's a good point, it should probably be more clear in the code why we do this check. However I'm not sure there's any official AWS source on [this change in the runtime](https://github.com/open-telemetry/opentelemetry-lambda/issues/2034#issuecomment-3591815062). Rereading the node 24 runtime announcement from AWS I don't see this information in there (or anything hinting at it). @maxday Could you shed any light here, or point to any official resource (if exists) that we can link back to?

pragmaticivan:
> @pragmaticivan That's a good point, it should probably be more clear in the code why we do this check. However I'm not sure there's any official AWS source on [this change in the runtime](https://github.com/open-telemetry/opentelemetry-lambda/issues/2034#issuecomment-3591815062). Rereading the node 24 runtime announcement from AWS I don't see this information in there (or anything hinting at it). @maxday Could you shed any light here, or point to any official resource (if exists) that we can link back to?

https://aws.amazon.com/blogs/compute/node-js-24-runtime-now-available-in-aws-lambda/

>Removing support for callback-based function handlers

I think that's the breaking change but not really sure

wpessers:
I would think the esm loading is kind of orthogonal to the removal of callback support. The blog post does mention something removal of legacy features in general in the new runtime interface client, so maybe this also involves standardization of module loading logic but I can only guess.

serkan-ozal (approved):

PR #2055, opened 2025-12-02
chore: Add badges to readme

thompson-tomo:
This adds both the cii best practice badge as well  the fossa badges which improves the clo monitor score as well as the ossf scorecard.


trask:
@thompson-tomo please create an issue in https://github.com/open-telemetry/sig-security to discuss first before sending these PRs so we can agree on a plan and we can then link these PRs to that plan in order to provide more context to maintainers. Thanks!

PR #2041, opened 2025-11-27
fix: update supported runtimes

maxday:
Fixes: https://github.com/open-telemetry/opentelemetry-lambda/issues/2039

This PR removes deprecated runtimes and adds newly released runtimes

replacing https://github.com/open-telemetry/opentelemetry-lambda/pull/2040

PR #2056, opened 2025-12-02
Add support for Python 3.14

lukasz-mitka:
Adds Python 3.14

Follow-up to https://github.com/open-telemetry/opentelemetry-lambda/pull/2041 with some additional places where python should be updated.

### DIFF
--- a/.github/workflows/ci-python.yml
+++ b/.github/workflows/ci-python.yml
@@ -31,7 +31,7 @@ jobs:
       matrix:
         # If you add a python version here, please make sure that the collector/Makefile publish and publish-layer targets
         # get updated as well
-        python: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
 
     steps:
       - name: Checkout this repo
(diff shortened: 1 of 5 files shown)


wpessers:
LGTM in general. Have you by any chance been able to test this in an actual lambda function with python 3.14, building the layer from source yourself? I don't see any open issues regarding python 3.14 in https://github.com/open-telemetry/opentelemetry-python, nor in https://github.com/open-telemetry/opentelemetry-python-contrib so most likely it should be fine.

lukasz-mitka:
No, I haven't tested it. Unlikely I will find time to figure out how to do it and then do it.
I will test it once it's available though :)

wpessers (approved):
LGTM, will try to find some time to do a quick smoke-test myself before merging. Thank you for contributing! 🚀

serkan-ozal:
Yeah, let's do a quick smoke that to be sure that it works with Python 3.14 in real AWS Lambda environment. OpenTelemetry Python SDK supports Python version 3.9+, so I am not expecting any compatibility issue, but it is still important to verify it (at least with a very basic hello-world like Lambda function).

wpessers:
Verified with simple aws sdk call to sts, setup available at https://github.com/wpessers/python-otel-smoke-test if anyone would be interested.

Issue #2054, opened 2025-12-01
Telemetryapi receiver types - excluding extension also drops platform


jousby:
**Describe the bug**
As per the telemetryapi [README](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/receiver/telemetryapireceiver/README.md) there are three types of telemetry you can subscribe to ["platform", "function", "extension"]. If you exclude "extension" from this set it also turns off "platform". 

**Steps to reproduce**
* Create and instrument a Lambda function
* Enable telemetryapi receiver for logs and traces and constrain the types for receiver to 'platform' and 'function'.
```
  telemetryapi:
    types:
      - platform
      - function
```
* Send logs to your OTEL backend of choice and check for the presence the platform log messages (platform.initStart, platform.initRuntimeDone)

**What did you expect to see?**
No logs for the extension.
Logs from the function
Logs from the platform for initStart and initRuntimeDone

**What did you see instead?**
Logs from the function only

**What version of collector/language SDK version did you use?**

    0.19.0 (collector layer)
    0.18.0 (nodejs layer)

**What language layer did you use?**
javascript

**Additional context**

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

wpessers:
This was fixed in #2058

PR #2058, opened 2025-12-08
feat: Add support for exporting telemetry API 'platform.*' log events

herin049:
Fixes issue #2054 where `platform.*` telemetry API event types (except part of `platform.report`) do not get exported for logging. The implemented solution captures all logs by default, and simplifies the current implementation by reusing the core log record construction logic while reducing code duplication.

I have setup a toy lambda function to export logs to Grafana Cloud with the changes in this PR, here are the updated logs:
Screenshot From 2025-12-08 18-31-52

### DIFF
--- a/collector/internal/telemetryapi/listener_test.go
+++ b/collector/internal/telemetryapi/listener_test.go
@@ -45,11 +45,6 @@ func setupListener(t *testing.T) (*Listener, string) {
 
 	address, err := listener.Start()
 	require.NoError(t, err)
-
-	t.Cleanup(func() {
-		listener.Shutdown()
-	})
-
 	return listener, address
 }
 
@@ -183,6 +178,7 @@ func TestListenOnAddress(t *testing.T) {
 
 func TestListener_StartAndShutdown(t *testing.T) {
 	listener, address := setupListener(t)
+	defer listener.Shutdown()
 	require.NotEqual(t, address, "", "Start() should not return an empty address")
 	require.True(t, strings.HasPrefix(address, "http://"), "Address should start with http://")
 	require.NotNil(t, listener.httpServer, "httpServer should not be nil")
@@ -241,6 +237,7 @@ func TestListener_httpHandler(t *testing.T) {
 	for _, test := range testCases {
 		t.Run(test.name, func(t *testing.T) {
 			listener, address := setupListener(t)
+			defer listener.Shutdown()
 			submitEvents(t, address, test.events)
 			require.EventuallyWithT(t, func(c *assert.CollectT) {
 				require.Equal(c, test.expectedCount, listener.queue.Len())
@@ -302,6 +299,7 @@ func TestListener_Wait_Success(t *testing.T) {
 	for _, test := range testCases {
 		t.Run(test.name, func(t *testing.T) {
 			listener, address := setupListener(t)
+			defer listener.Shutdown()
 
 			waitDone := make(chan error, 1)
 			go func() {
--- a/collector/internal/telemetryapi/listener_test.go
+++ b/collector/internal/telemetryapi/listener_test.go
@@ -178,7 +178,6 @@ func TestListenOnAddress(t *testing.T) {
 
 func TestListener_StartAndShutdown(t *testing.T) {
 	listener, address := setupListener(t)
-	defer listener.Shutdown()
 	require.NotEqual(t, address, "", "Start() should not return an empty address")
 	require.True(t, strings.HasPrefix(address, "http://"), "Address should start with http://")
 	require.NotNil(t, listener.httpServer, "httpServer should not be nil")
@@ -190,6 +189,7 @@ func TestListener_StartAndShutdown(t *testing.T) {
 		require.NoError(t, resp.Body.Close())
 	}
 
+	listener.Shutdown()
 	require.Nil(t, listener.httpServer, "httpServer should be nil after Shutdown()")
 }
 
--- a/collector/internal/telemetryapi/listener_test.go
+++ b/collector/internal/telemetryapi/listener_test.go
@@ -189,7 +189,6 @@ func TestListener_StartAndShutdown(t *testing.T) {
 	} else {
 		require.NoError(t, resp.Body.Close())
 	}
-	listener.Shutdown()
 
 	require.Nil(t, listener.httpServer, "httpServer should be nil after Shutdown()")
 }
--- a/collector/receiver/telemetryapireceiver/config.go
+++ b/collector/receiver/telemetryapireceiver/config.go
@@ -23,7 +23,7 @@ type Config struct {
 	extensionID string
 	Port        int      `mapstructure:"port"`
 	Types       []string `mapstructure:"types"`
-	LogReport   *bool    `mapstructure:"log_report"`
+	LogReport   bool     `mapstructure:"log_report"`
 }
 
 // Validate validates the configuration by checking for missing or invalid fields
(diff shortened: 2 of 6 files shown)


collector/internal/telemetryapi/listener_test.go
```diff
@@ -45,11 +45,6 @@ func setupListener(t *testing.T) (*Listener, string) {
 
 	address, err := listener.Start()
 	require.NoError(t, err)
-
```
herin049, line 48: The use of `t.Cleanup()` appeared to trigger a race condition using the race detector with `zap.Logger` switching to using `defer` appears to fix the issue.

collector/receiver/telemetryapireceiver/config.go
```diff
@@ -23,7 +23,7 @@ type Config struct {
 	extensionID string
 	Port        int      `mapstructure:"port"`
 	Types       []string `mapstructure:"types"`
-	LogReport   bool     `mapstructure:"log_report"`
+	LogReport   *bool    `mapstructure:"log_report"`
```
herin049: Switch to pointer to allow default value to be `true`

wpessers:
@herin049 I see your solution captures all the logs by default. Does it also fix the issue, so can you exclude the extension logs and only get function and platform logs like was the reason for the original issue?

herin049:
> @herin049 I see your solution captures all the logs by default. Does it also fix the issue, so can you exclude the extension logs and only get function and platform logs like was the reason for the original issue?

Yes, I can verify again by manually deploying the layer into my AWS account. The original issue is worded a bit weirdly, the original issue is that `platform` logs are not generated. The creator of the issue was trying to exclude `extension` and include `platform` and `function` but was only getting `function` logs (he was receiving no `extension` logs as expected, but also wasn't getting the `platform` logs). 

The determining factor if you receive these logs or not if whether you add `platform`/`extension`/`function` to the event list which is passed to the Lambda TelemetryAPI during the initial subscription call. If you don't include one of these, then you'll get none of those events.

collector/internal/telemetryapi/listener_test.go
```diff
@@ -183,6 +178,7 @@ func TestListenOnAddress(t *testing.T) {
 
 func TestListener_StartAndShutdown(t *testing.T) {
 	listener, address := setupListener(t)
+	defer listener.Shutdown()
```
wpessers: Since we're deferring the `Shutdown()` call here, can we remove the other `Shutdown()` call at the end (line 192)?

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -430,14 +518,19 @@ func newTelemetryAPIReceiver(
 		}
 	}
 
+	logReport := true
+	if cfg.LogReport != nil {
+		logReport = *cfg.LogReport
+	}
```
wpessers: So in the PR where this was added: https://github.com/open-telemetry/opentelemetry-lambda/pull/2011
a deliberate choice was made to disable the cloudwatch-like report log by default. I'd prefer not changing that to default true. If you'd like that to change I suggest opening a separate issue to motivate that decision.

collector/receiver/telemetryapireceiver/config.go
wpessers: Since now an additional `nil` check is required, I'd prefer leaving this a `bool` if possible. If we want to change the defaulting behaviour to default to true, I believe we could just do that inside the receiver factory method: https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/receiver/telemetryapireceiver/factory.go#L46-L50

wpessers (changes requested):
Some minor remarks regarding implementation to check out. Other than that, overall idea and approach looks good! 🚀

collector/internal/telemetryapi/listener_test.go
herin049: Updated.

collector/receiver/telemetryapireceiver/receiver.go
herin049: Reverted back to the way it was before. I think I'll create another issue to address this.

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -188,100 +200,188 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
...
+		events, _ := record["events"].([]interface{})
+		if name != "" {
+			return fmt.Sprintf(platformExtensionLogFmt, name, state, events)
+		}
+	case string(telemetryapi.PlatformLogsDropped):
+		droppedRecords, _ := record["droppedRecords"].(float64)
+		droppedBytes, _ := record["droppedBytes"].(float64)
+		reason, _ := record["reason"].(string)
+		if reason != "" {
+			return fmt.Sprintf(platformLogsDroppedLogFmt, int(droppedRecords), int(droppedBytes), reason)
```
wpessers: Could it make sense here to not cast to int but instead change the actual format string to use floats `%.0f` instead of integers `%d`? So that we  don't need to cast to int here.

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -188,100 +200,188 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
...
+	case string(telemetryapi.PlatformExtension):
+		name, _ := record["name"].(string)
+		state, _ := record["state"].(string)
+		events, _ := record["events"].([]interface{})
+		if name != "" {
+			return fmt.Sprintf(platformExtensionLogFmt, name, state, events)
+		}
+	case string(telemetryapi.PlatformLogsDropped):
+		droppedRecords, _ := record["droppedRecords"].(float64)
+		droppedBytes, _ := record["droppedBytes"].(float64)
```
wpessers: AWS probably guarantees these fields being present. But I'm wondering if it could make sense to check the ok values for these numeric fields since otherwise it defaults to 0, which would be wrong. Again, maybe too defensive?

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -188,100 +200,188 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
...
+		initType, _ := record["initializationType"].(string)
+		phase, _ := record["phase"].(string)
+		status, _ := record["status"].(string)
+		var durationMs float64
+		if metrics, ok := record["metrics"].(map[string]interface{}); ok {
+			durationMs, _ = metrics["durationMs"].(float64)
+		}
+		if initType != "" || phase != "" || status != "" || durationMs != 0 {
+			return fmt.Sprintf(platformInitReportLogFmt, initType, phase, status, durationMs)
+		}
```
wpessers, line 323: I'm wondering if it's too defensive maybe, since `durationMs` should never actually be 0. But at the moment we're relying on the default 0 value to determine if the `durationMs` metric was present or not. The following might be clearer, but again maybe too defensive?
```suggestion
		var durationMs float64
		var hasDuration bool
		if metrics, ok := record["metrics"].(map[string]interface{}); ok {
			durationMs, hasDuration = metrics["durationMs"].(float64)
		}
		if initType != "" || phase != "" || status != "" || hasDuration {
			return fmt.Sprintf(platformInitReportLogFmt, initType, phase, status, durationMs)
		}
```

wpessers (commented):
Just some final thoughts / questions. Lmk what you think.

herin049:
@wpessers Updated all `float64` conversions to explicitly check the `ok` result instead of checking for a non-zero return value. Also updated the format string to use `%.0f` instead of casting to an int.

wpessers:
Thanks again for the work! Nice one! 🚀

wpessers (approved):

PR #2066, opened 2025-12-10
feat: add support for FaaS telemetryapi metrics

herin049:
Adds support for FaaS telemetry API metrics.

### DIFF
--- a/collector/receiver/telemetryapireceiver/metric_builder.go
+++ b/collector/receiver/telemetryapireceiver/metric_builder.go
@@ -20,7 +20,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/pdatautil"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
-	semconv2 "go.opentelemetry.io/otel/semconv/v1.24.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.24.0"
 )
 
 const MiB = float64(1 << 20)
@@ -285,9 +285,9 @@ func (c *CounterMetricBuilder) AppendDataPoints(scopeMetrics pmetric.ScopeMetric
 
 func NewFaaSInvokeDurationMetricBuilder(startTime pcommon.Timestamp, temporality pmetric.AggregationTemporality) *HistogramMetricBuilder {
 	return NewHistogramMetricBuilder(
-		semconv2.FaaSInvokeDurationName,
-		semconv2.FaaSInvokeDurationDescription,
-		semconv2.FaaSInvokeDurationUnit,
+		semconv.FaaSInvokeDurationName,
+		semconv.FaaSInvokeDurationDescription,
+		semconv.FaaSInvokeDurationUnit,
 		DurationHistogramBounds,
 		startTime,
 		temporality,
@@ -296,9 +296,9 @@ func NewFaaSInvokeDurationMetricBuilder(startTime pcommon.Timestamp, temporality
 
 func NewFaaSInitDurationMetricBuilder(startTime pcommon.Timestamp, temporality pmetric.AggregationTemporality) *HistogramMetricBuilder {
 	return NewHistogramMetricBuilder(
-		semconv2.FaaSInitDurationName,
-		semconv2.FaaSInitDurationDescription,
-		semconv2.FaaSInitDurationUnit,
+		semconv.FaaSInitDurationName,
+		semconv.FaaSInitDurationDescription,
+		semconv.FaaSInitDurationUnit,
 		DurationHistogramBounds,
 		startTime,
 		temporality,
@@ -307,9 +307,9 @@ func NewFaaSInitDurationMetricBuilder(startTime pcommon.Timestamp, temporality p
 
 func NewFaaSMemUsageMetricBuilder(startTime pcommon.Timestamp, temporality pmetric.AggregationTemporality) *HistogramMetricBuilder {
 	return NewHistogramMetricBuilder(
-		semconv2.FaaSMemUsageName,
-		semconv2.FaaSMemUsageDescription,
-		semconv2.FaaSMemUsageUnit,
+		semconv.FaaSMemUsageName,
+		semconv.FaaSMemUsageDescription,
+		semconv.FaaSMemUsageUnit,
 		MemUsageHistogramBounds,
 		startTime,
 		temporality,
@@ -318,9 +318,9 @@ func NewFaaSMemUsageMetricBuilder(startTime pcommon.Timestamp, temporality pmetr
 
 func NewFaaSColdstartsMetricBuilder(startTime pcommon.Timestamp, temporality pmetric.AggregationTemporality) *CounterMetricBuilder {
 	return NewCounterMetricBuilder(
-		semconv2.FaaSColdstartsName,
-		semconv2.FaaSColdstartsDescription,
-		semconv2.FaaSColdstartsUnit,
+		semconv.FaaSColdstartsName,
+		semconv.FaaSColdstartsDescription,
+		semconv.FaaSColdstartsUnit,
 		true,
 		startTime,
 		temporality,
@@ -329,9 +329,9 @@ func NewFaaSColdstartsMetricBuilder(startTime pcommon.Timestamp, temporality pme
 
 func NewFaaSErrorsMetricBuilder(startTime pcommon.Timestamp, temporality pmetric.AggregationTemporality) *CounterMetricBuilder {
 	return NewCounterMetricBuilder(
-		semconv2.FaaSErrorsName,
-		semconv2.FaaSErrorsDescription,
-		semconv2.FaaSErrorsUnit,
+		semconv.FaaSErrorsName,
+		semconv.FaaSErrorsDescription,
+		semconv.FaaSErrorsUnit,
 		true,
 		startTime,
 		temporality,
@@ -340,9 +340,9 @@ func NewFaaSErrorsMetricBuilder(startTime pcommon.Timestamp, temporality pmetric
 
 func NewFaaSInvocationsMetricBuilder(startTime pcommon.Timestamp, temporality pmetric.AggregationTemporality) *CounterMetricBuilder {
 	return NewCounterMetricBuilder(
-		semconv2.FaaSInvocationsName,
-		semconv2.FaaSInvocationsDescription,
-		semconv2.FaaSInvocationsUnit,
+		semconv.FaaSInvocationsName,
+		semconv.FaaSInvocationsDescription,
+		semconv.FaaSInvocationsUnit,
 		true,
 		startTime,
 		temporality,
@@ -351,9 +351,9 @@ func NewFaaSInvocationsMetricBuilder(startTime pcommon.Timestamp, temporality pm
 
 func NewFaaSTimeoutsMetricBuilder(startTime pcommon.Timestamp, temporality pmetric.AggregationTemporality) *CounterMetricBuilder {
 	return NewCounterMetricBuilder(
-		semconv2.FaaSTimeoutsName,
-		semconv2.FaaSTimeoutsDescription,
-		semconv2.FaaSTimeoutsUnit,
+		semconv.FaaSTimeoutsName,
+		semconv.FaaSTimeoutsDescription,
+		semconv.FaaSTimeoutsUnit,
 		true,
 		startTime,
 		temporality,
--- a/collector/receiver/telemetryapireceiver/metric_builder.go
+++ b/collector/receiver/telemetryapireceiver/metric_builder.go
@@ -1,0 +1,225 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package telemetryapireceiver
+
+import (
+	"sort"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	semconv2 "go.opentelemetry.io/otel/semconv/v1.24.0"
+)
+
+var DefaultHistogramBounds = []float64{0.0, 5.0, 10.0, 25.0, 50.0, 75.0, 100.0, 250.0, 500.0, 750.0, 1000.0, 2500.0, 5000.0, 7500.0, 10000.0}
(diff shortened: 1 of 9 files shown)


collector/receiver/telemetryapireceiver/metric_builder.go
```diff
@@ -0,0 +1,383 @@
...
+
+package telemetryapireceiver
+
+import (
+	"sort"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/pdatautil"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	semconv2 "go.opentelemetry.io/otel/semconv/v1.24.0"
```
wpessers: Why is this specific version needed? Can we not just use the same `1.25.0` version that is used elsewhere?

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -33,9 +33,11 @@ import (
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	"go.opentelemetry.io/collector/receiver"
 	semconv "go.opentelemetry.io/collector/semconv/v1.25.0"
+	semconv2 "go.opentelemetry.io/otel/semconv/v1.24.0"
```
wpessers: Related to other semconv version remark but I see now that you are seemingly doing this to deal with the deprecation of the old `collector/semconv` package in favor of `otel/semconv`. Could we not use "go.opentelemetry.io/otel/semconv/v1.25.0` for all semconv attributes? I'm assuming that that same version in the new package matches the one from the old package

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -470,7 +587,7 @@ func (r *telemetryAPIReceiver) createPlatformInitSpan(start, end string) (ptrace
 	span := ss.Spans().AppendEmpty()
 	span.SetTraceID(newTraceID())
 	span.SetSpanID(newSpanID())
-	span.SetName("platform.initRuntimeDone")
+	span.SetName(fmt.Sprintf("init %s", r.faasName))
```
wpessers, line 586: Just a question, why is this needed for supporting telemetryapi metrics? Or is this just a general improvement/fix?

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -71,6 +78,8 @@ type telemetryAPIReceiver struct {
 	types                   []telemetryapi.EventType
 	resource                pcommon.Resource
 	faasFunctionVersion     string
+	faasName                string
+	faaSMetricBuilders      *FaaSMetricBuilders
```
wpessers, line 81: debatble nit - naming
```suggestion
	faasMetricBuilders      *FaasMetricBuilders
```
lmk what you think.

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -170,15 +200,13 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
...
-			if r.nextTraces != nil {
-				err := r.nextTraces.ConsumeTraces(context.Background(), td)
-				if err == nil {
-					r.lastPlatformEndTime = ""
-					r.lastPlatformStartTime = ""
-				} else {
-					r.logger.Error("error receiving traces", zap.Error(err))
+	// Metrics
+	if r.nextMetrics != nil {
+		if metrics, err := r.createMetrics(slice); err == nil {
```
wpessers, line 201: Should we be worried about concurrent map writes here? Since createMetrics modifies the `faaSMetricBuilders` maps...

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -145,10 +154,31 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
 		case string(telemetryapi.PlatformInitStart):
 			r.logger.Info(fmt.Sprintf("Init start: %s", r.lastPlatformStartTime), zap.Any("event", el))
 			r.lastPlatformStartTime = el.Time
+
+			if record, ok := el.Record.(map[string]any); ok {
+				functionName, _ := record["functionName"].(string)
+				if functionName != "" {
+					r.faasName = functionName
```
wpessers, line 157: Same-ish remark here as: https://github.com/open-telemetry/opentelemetry-lambda/pull/2066/changes#r2637209797
Just wondering if there's any risk for concurrent writes here? Since this single telemetryAPIReceiver instance is shared across request goroutines.

collector/receiver/telemetryapireceiver/receiver.go
herin049: I forgot to mention this in the PR description. I added this to try to align the span name closer to the semantic convention guidelines. The spec doesn't explicitly mention cold start spans but it does mention that the invocation span name is set to the FaaS name, so I figured the init span should be `init <FaaSName>" Let me know what you think

collector/receiver/telemetryapireceiver/receiver.go
herin049: This is how the semconv library has it which is why I've used FaaS since its short for "Function as a Service"

collector/receiver/telemetryapireceiver/receiver.go
herin049: This is a good catch, and I'm very suprised this wasn't caught earlier. All modifications to shared variables in this receiver handler could result in race conditions, not just the logic that is added in this PR. That being said, I've refactored everything so that this is no longer an issue. All processing is in a single goroutine now, so there cannot be concurrent modification of  these shared variables.

collector/receiver/telemetryapireceiver/receiver.go
herin049: See comment above.

collector/receiver/telemetryapireceiver/receiver.go
herin049: I can update this, just need to make sure there aren't any changes with these libraries that would affect the attribute names.

collector/receiver/telemetryapireceiver/metric_builder.go
herin049: updated to `1.25.0`

collector/receiver/telemetryapireceiver/receiver.go
herin049: Updated.

wpessers:
Regarding the fix for the concurrent writes, I was actually thinking more in this direction: #2091 Lmk what you think. If it looks good, I suggest we just merge it and then add your changes on top after reverting 652d63e

The reason is I'd like to keep the expensive work of reading and parsing the JSON from the incoming requests concurrent. Not sure if it will really make a big difference, as I can't really estimate the rps that aws lambda telemetryapi sends, but it seems like an easy optimization to keep.

herin049:
> Regarding the fix for the concurrent writes, I was actually thinking more in this direction: #2091 Lmk what you think. If it looks good, I suggest we just merge it and then add your changes on top after reverting [652d63e](https://github.com/open-telemetry/opentelemetry-lambda/commit/652d63e978b065989f4067f22629700d29d36b5b)
> 
> The reason is I'd like to keep the expensive work of reading and parsing the JSON from the incoming requests concurrent. Not sure if it will really make a big difference, as I can't really estimate the rps that aws lambda telemetryapi sends, but it seems like an easy optimization to keep.

The changes in this PR do still keep the reading/parsing of JSON concurrent given that the `telemetryHandler` parses the record before inserting it into the channel. 

I did initially consider just using a mutex instead of a separate channel, but I went with this approach to ensure that if we receive a burst of requests, the handler will still return almost immediately (assuming there is still capacity in the channel). If we use the mutex, it might result in blocking other requests, potentially resulting in dropped telemetry events 

> If the subscriber cannot process incoming telemetry fast enough, or if your function code generates very high log volume, Lambda might drop records to keep memory utilization bounded. When this occurs, Lambda sends a platform.logsDropped event.

However, the telemetry API does perform some buffering internally, so this probably isn't as big of an issue I make it out to be. I'll leave it up to you, I'm fine with just using the mutex, I do agree it is much simpler than setting up a channel and I'm guessing the performance is going to be roughly identical.

wpessers:
#2091 has since been discussed, approved and now merged.
For future transparency: We opted for that mutex solution because it's simpler and we can also apparently configure the buffering of the telemetry api ourselves if needed. @herin049 pointed to this AWS doc that clearly shows the possible configuration: https://docs.aws.amazon.com/lambda/latest/dg/telemetry-api-reference.html#telemetry-subscribe-api

@herin049 I'm ready to merge this one whenever you get the time to rebase :)

herin049:
@wpessers rebased the changes, should be good to go!

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -206,6 +233,91 @@ func (r *telemetryAPIReceiver) getRecordRequestId(record map[string]interface{})
...
+		case string(telemetryapi.PlatformInitStart):
+			r.faaSMetricBuilders.coldstartsMetric.Add(1)
+			r.faaSMetricBuilders.coldstartsMetric.AppendDataPoints(scopeMetric, pcommon.NewTimestampFromTime(ts))
+		case string(telemetryapi.PlatformInitReport):
+			metrics, ok := record["metrics"].(map[string]any)
+			if !ok {
+				continue
+			}
+
+			status, _ := metrics["status"].(string)
```
wpessers: Doesn't look like the record's `metrics` field has a status field: https://docs.aws.amazon.com/lambda/latest/dg/telemetry-schema-reference.html#platform-initReport

```suggestion
			metrics, ok := record["metrics"].(map[string]any)
			if !ok {
				continue
			}

			status, _ := record["status"].(string)
```

herin049: Updated, should have been on the main `record` object.

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -206,6 +233,91 @@ func (r *telemetryAPIReceiver) getRecordRequestId(record map[string]interface{})
...
+			if status == telemetrySuccessStatus {
+				r.faaSMetricBuilders.invocationsMetric.Add(1)
+				r.faaSMetricBuilders.invocationsMetric.AppendDataPoints(scopeMetric, pcommon.NewTimestampFromTime(ts))
+			} else if status == telemetryFailureStatus || status == telemetryErrorStatus {
+				r.faaSMetricBuilders.errorsMetric.Add(1)
+				r.faaSMetricBuilders.errorsMetric.AppendDataPoints(scopeMetric, pcommon.NewTimestampFromTime(ts))
+			} else if status == telemetryTimeoutStatus {
+				r.faaSMetricBuilders.timeoutsMetric.Add(1)
+				r.faaSMetricBuilders.timeoutsMetric.AppendDataPoints(scopeMetric, pcommon.NewTimestampFromTime(ts))
+			}
```
wpessers, line 304: Not sure, but in case of failure or timeout, shouldn't the invocationsMetric still be incremented?

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -206,6 +233,91 @@ func (r *telemetryAPIReceiver) getRecordRequestId(record map[string]interface{})
...
+		}
+		ts, err := time.Parse(time.RFC3339, el.Time)
+		if err != nil {
+			continue
+		}
+
+		switch el.Type {
+		case string(telemetryapi.PlatformInitStart):
+			r.faaSMetricBuilders.coldstartsMetric.Add(1)
+			r.faaSMetricBuilders.coldstartsMetric.AppendDataPoints(scopeMetric, pcommon.NewTimestampFromTime(ts))
```
wpessers, line 258: This pattern of always calling `.Add(1)` and then subsequently `AppendDataPoints(...)` seems to me like it would lead to always exporting no? Since [here](https://github.com/open-telemetry/opentelemetry-lambda/pull/2066/changes#diff-4ff12eb39f63588f46bf6892fd6ef44d708e94e81e81daf9b106d4f15155f7f0R214-R223) `lastUpdated` is alwas set to the current epoch, wouldn't [the check in `AppendDataPoints`](https://github.com/open-telemetry/opentelemetry-lambda/pull/2066/changes#diff-4ff12eb39f63588f46bf6892fd6ef44d708e94e81e81daf9b106d4f15155f7f0R123) always evaluate `true`?

wpessers: Also just for the sake of clarity, do we want the epoch to increment after each single event or rather after a single batch of events (per http request from the telemetryapi)? I'm not sure it adds value to increment the epoch and export the metric separately per event, instead of considering the batch of events to belong to the same epoch?

wpessers (commented):
So sorry for taking so long on this PR. But it's quite a hefty one with a lot of additions and I feel like on every new read through it I find new stuff that catches my eye.

collector/receiver/telemetryapireceiver/receiver.go
herin049: This is one that we might have to come back and tweak, but this is what the spec says:

> Description: Number of successful invocations.

[https://opentelemetry.io/docs/specs/semconv/faas/faas-metrics/#metric-faasinvocations](https://opentelemetry.io/docs/specs/semconv/faas/faas-metrics/#metric-faasinvocations)

I'm assuming that we wouldn't want to increment the number of invocations if any of the failure scenarios above are present. That being said, one edge case that this doesn't handle is that if the runtime itself throws an exception this failure isn't actually communicated with the telemetry API. That is, it reports it as a successful invocation because it was successful from the perspective of Lambda. At some point we may need to determine a way for the runtime to communicate a failure to the extension and act accordingly. 

However, I figure that given that this is an initial implementation, we can stick with this approach, and refine it in a later PR. Alternatively, we could deviate from the spec a bit and just increment the invocation metric on every request. Let me know what you think.

wpessers: Good catch, I completely missed that in the spec. Current implementation is fine in that case!

collector/receiver/telemetryapireceiver/receiver.go
herin049: Good point, I did consider this, but it would have expanded the scope of the changes by a bit. I mostly added the epoch check for future changes so we can accumulate in memory and only export after a certain amount of time has elapsed (e.g. 60s), similar to how the SDK works. 

We could do as you said and move the call to `AppendDataPoints` to after we generate all of the metrics, but this would have the same effect as the approach above since we would end up exporting after each invocation anyways.

herin049: Let me know if you think I should update this PR to include an export interval, otherwise we can merge this PR and I can immediately start on the work to configure export intervals.

wpessers: Good point, I think a follow-up PR for export intervals makes sense!

wpessers (approved):

wpessers:
@herin049 can you rebase one last time 🙏

herin049:
@wpessers Done!

wpessers:
LGTM then! Appreciate the patience and the collaboration! 😄

herin049:
@wpessers Thanks for all the reviews! I'll follow up shortly with the changes to improve the exporting logic to not export on every invocation.

PR #2069, opened 2025-12-11
remove unnecessary call to AwsLambdaInstrumentor().instrument() in wrapper script

### LINKED ISSUE #2068: Python: Attempt to instrument while already instrument warning
**Describe the bug**
When using the OpenTelemetry Python auto-instrumentation layer, the following warning appears in the Lambda logs on every cold start:

```
[WARNING]	2025-12-11T04:45:34.152Z Attempting to instrument while already instrumented
```

**Steps to reproduce**
Deploy a lambda with the Python auto-instrumentation layer and perform an invocation.

No warning message present.

**What did you see instead?**

The warning message above is present.

Version: latest

Config: Python

herin049:
Fixes: #2068

### DIFF
--- a/python/src/otel/otel_sdk/otel_wrapper.py
+++ b/python/src/otel/otel_sdk/otel_wrapper.py
@@ -48,9 +48,6 @@ def modify_module_name(module_name):
 class HandlerError(Exception):
     pass
 
-
-AwsLambdaInstrumentor().instrument()
-
 path = os.environ.get("ORIG_HANDLER")
 
 if path is None:
--- a/python/src/otel/otel_sdk/otel_wrapper.py
+++ b/python/src/otel/otel_sdk/otel_wrapper.py
@@ -37,7 +37,6 @@ https://docs.python.org/3/library/imp.html#imp.load_module
 import os
 from importlib import import_module
 
-from opentelemetry.instrumentation.aws_lambda import AwsLambdaInstrumentor
 
 
 def modify_module_name(module_name):
--- a/python/src/otel/tests/test_otel.py
+++ b/python/src/otel/tests/test_otel.py
@@ -153,6 +153,7 @@ def mock_execute_lambda(event=None):
     Args:
         event: The Lambda event which may or may not be used by instrumentation.
     """
+    AwsLambdaInstrumentor().instrument()
 
     # The point of the repo is to test using the script, so we can count on it
     # being here for every test and do not check for its existence.


serkan-ozal:
While removing explicit `AwsLambdaInstrumentor().instrument()` seems OK to me, I am not fully sure on
- might there be a side effect? did you try the changes here with a custom layer?
- why it added before? was `AwsLambdaInstrumentor` not part of default list of auto instrumented packages before?

herin049:
> While removing explicit `AwsLambdaInstrumentor().instrument()` seems OK to me, I am not fully sure on
> 
>     * might there be a side effect? did you try the changes here with a custom layer?
> 
>     * why it added before? was `AwsLambdaInstrumentor` not part of default list of auto instrumented packages before?

@serkan-ozal Looks like the reasoning was given here in a comment which I accidentally missed:

```
Usage
-----
Patch the reserved `_HANDLER` Lambda environment variable to point to this
file's `otel_wrapper.lambda_handler` property. Do this having saved the original
`_HANDLER` in the `ORIG_HANDLER` environment variable. Doing this makes it so
that **on import of this file, the handler is instrumented**.

Instrumenting any earlier will cause the instrumentation to be lost because the
AWS Service uses `imp.load_module` to import the handler which RELOADS the
module. This is why AwsLambdaInstrumentor cannot be instrumented with the
`opentelemetry-instrument` script.
```
Looks like this was added by @NathanielRN 

However, I don't think the reasoning here is quite correct. Even IF lambda does reload the module being pointed to by `_HANDLER`, this will not matter because the original lambda handler pointed to by `ORIG_HANDLER` lives in a different module and thus, will NOT be reloaded if the lambda runtime reloads the module pointed to by `_HANLDER` (which is the `otel_wrapper` module) 

(see https://docs.python.org/3/library/importlib.html#importlib.reload)

Even the original approach doesn't accomplish the goal of instrumenting the lambda handler inside the wrapper script because the AWS Lambda instrumentation library is not excluded from being used by auto instrumentation. As a result, the lambda instrumentation library is called before the wrapper script is run and any subsequent attempts to instrument will be a no-op due to the short-circuiting logic inside the Python instrumentation library:

```
    def instrument(self, **kwargs: Any):
        """Instrument the library

        This method will be called without any optional arguments by the
        ``opentelemetry-instrument`` command.

        This means that calling this method directly without passing any
        optional values should do the very same thing that the
        ``opentelemetry-instrument`` command does.
        """

        if self._is_instrumented_by_opentelemetry:
            _LOG.warning("Attempting to instrument while already instrumented")
            return None
```

I've tested these changes by building the auto instrumentation layer exactly how it is in this PR with Python 3.14, if you'd like, I can verify that this works on the older lambda runtimes. This also makes me think that we should at some point add an integration test suite that runs in AWS against all supported lambda runtimes - at least before releasing new layer versions. Let me know what you think, we can discuss details further in one of the FaaS SIG meetings. This is something I would be very happy to lead development on.

serkan-ozal:
@herin049 I would test with older (but currently available) Python Lambda runtimes too. AWS Lambda Python runtime clients (which includes `bootstrap.py`) might have different implementations (we have faced some cases in Nodejs) and `AwsLambdaInstrumentor().instrument()` call might have been added for some weird trick/case in one of the older Lambda Python runtimes.

herin049:
@serkan-ozal I was able to test and manually verify that the changes work with Python `3.9`, `3.11`, `3.12`, `3.13` and `3.14`.

wpessers:
> This also makes me think that we should at some point add an integration test suite that runs in AWS against all supported lambda runtimes - at least before releasing new layer versions.

Talked about this in the SIG meeting with @tylerbenson today as well. I have an issue open here to track this effort as well: https://github.com/open-telemetry/opentelemetry-lambda/issues/2079

tylerbenson (approved):

tylerbenson:
Sounds good to me.  Thanks for verifying.

serkan-ozal (approved):

wpessers (approved):

PR #2081, opened 2025-12-18
Reproduce problem with Node24

mduesterhoeft:
Reproduction scenario for https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3314

Trying to use OpenTelemetery instrumentation for AWS Lambda functions on Node24
Using NodejsFunction - which uses esbuild to build the bundle before deployment

To deploy run
```
npm install
npx cdk deploy 
```


wpessers:
Closing this since it seems to be simply using the examples from the repo to reproduce some issue.

PR #2084, opened 2025-12-21
feat(collector): optimize telemetryapireceiver trace and span id generation

wpessers:
Use `crypto/rand`'s `Read` method directly. Instead of reinitializing and seeding the rng on each span / trace id generation. Saving:

1. Heap allocations
    - `rand.NewSource(rngSeed)`
    - `rand.New(...)`
2. Some init logic - CPU time
    - `rand.NewSource(rngSeed)` - PRNG state initialization (also apparently [not safe for concurrent use](https://pkg.go.dev/math/rand#NewSource))
3. Probably negligible but function call overhead of calling `rand.NewSource(rngSeed)` and `rand.New(...)`

What remains now is basically a very efficient syscall, with no heap allocations.

### DIFF
--- a/collector/receiver/telemetryapireceiver/receiver.go
+++ b/collector/receiver/telemetryapireceiver/receiver.go
@@ -17,11 +17,9 @@ package telemetryapireceiver // import "github.com/open-telemetry/opentelemetry-
 import (
 	"context"
 	crand "crypto/rand"
-	"encoding/binary"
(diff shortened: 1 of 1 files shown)


wpessers:
Noticed this in the existing code when reviewing changes from #2058

herin049 (approved):
LGTM

PR #2091, opened 2025-12-23
feat(collector): use mutex for locking struct fields but still allowi…

wpessers:
…ng the reading and parsing of the request bodies to happen concurrently


collector/receiver/telemetryapireceiver/receiver.go
wpessers, line 126: Also this seems incorrect? I only see that queue being used in a debug log down below...

herin049: I noticed this as well, we can remove it.

herin049 (approved):

PR #2096, opened 2025-12-26
refactor: improve telemetry api receiver listener initialization logic

herin049:
Improves the Telemetry API receiver listener initialization logic to more closely match the logic used elsewhere in the layer. Updated the default behavior to listen on `:0` by default to eliminate the need of manually specifying ports and eliminate the possibility of encountering `Error: address already in use` errors that are very rare and difficult to debug. The listener now handles shutdown properly by closing the HTTP listener, and several incorrect log statements have been updated with a few additional ones added.


collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -75,30 +76,52 @@ type telemetryAPIReceiver struct {
...
 }
 
+func (r *telemetryAPIReceiver) bindListener(ctx context.Context) (net.Listener, string, error) {
+	listenerAddr := listenOnAddress()
+	var lc net.ListenConfig
+	l, err := lc.Listen(ctx, "tcp", fmt.Sprintf("%s:%d", listenerAddr, r.port))
+	if err != nil {
+		return nil, "", err
+	}
+	addr := fmt.Sprintf("%s:%d", l.Addr().Network(), l.Addr().(*net.TCPAddr).Port)
```
wpessers: Looks like `l.Addr().Network()` returns the network type string "tcp". So wouldn't this produce an address like `tcp:53421` instead of `sandbox.localdomain:53421`?

That would mean the subscribe URI passed to the Telemetry API ends up as http://tcp:53421/, which Lambda wouldn't be able to resolve to deliver events to I think?

For reference, the internal listener at [internal/telemetryapi/listener.go](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/internal/telemetryapi/listener.go#L51-L59) uses the `listenerAddr` variable directly for this and I think a lot of the logic we already have there is directly applicable here as well.

herin049: Good catch, this must have been a typo, or I made this change very late at night. I can update this, and add a test, since I'm surprised this managed to slip through.

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -75,30 +76,52 @@ type telemetryAPIReceiver struct {
...
+	return l, addr, nil
+}
+
 func (r *telemetryAPIReceiver) Shutdown(ctx context.Context) error {
-	return nil
+	err := r.httpServer.Shutdown(ctx)
+	return err
 }
```
wpessers, line 139: Again looking at some "prior art" in [internal/telemetryapi/listener.go](https://github.com/open-telemetry/opentelemetry-lambda/blob/main/collector/internal/telemetryapi/listener.go#L124-L135) it looks like just to be safe we might want to also nil check `r.httpServer` here before calling the `Shutdown` method on it?

wpessers:
@herin049 if you get a chance to check out the feedback again, we can merge this sometime soon.

herin049:
> @herin049 if you get a chance to check out the feedback again, we can merge this sometime soon.

@wpessers Sorry, had this on hold for a bit, will fix all of the issues by tomorrow!

wpessers:
> @wpessers Sorry, had this on hold for a bit, will fix all of the issues by tomorrow!

No rush! Just wanted to make sure it’s still on your radar.

wpessers (approved):

PR #2114, opened 2026-01-23
feat: add configurable metric export interval to lambda telemetry API receiver

herin049:
Followup to changes made in #2066. The changes in this PR add a configurable metric export interval to prevent metrics from being generated for every Lambda invocation, reducing the number of metric datapoints generated. Relevant tests have been added to test the new exporting behavior and configuration changes. Furthermore, relevant documentation has been added documenting how to configure metric temporality and export intervals.

### DIFF
--- a/collector/receiver/telemetryapireceiver/receiver.go
+++ b/collector/receiver/telemetryapireceiver/receiver.go
@@ -84,6 +84,9 @@ type telemetryAPIReceiver struct {
 	currentFaasInvocationID string
 	lambdaInitType          lambdalifecycle.InitType
 	logReport               bool
+	exportInterval          time.Duration
+	stopCh                  chan struct{}
+	wg                      sync.WaitGroup
 }
 
 func (r *telemetryAPIReceiver) bindListener() (net.Listener, string, error) {
@@ -119,6 +122,11 @@ func (r *telemetryAPIReceiver) Start(ctx context.Context, host component.Host) e
 		}
 	}()
 
+	if r.exportInterval > 0 {
+		r.wg.Add(1)
+		go r.startMetricsExporter()
+	}
+
 	telemetryClient := telemetryapi.NewClient(r.logger)
 	if _, err := telemetryClient.Subscribe(ctx, r.types, r.extensionID, fmt.Sprintf("http://%s/", address)); err != nil {
 		r.logger.Error("Failed to subscribe to telemetry", zap.Error(err))
@@ -130,10 +138,70 @@ func (r *telemetryAPIReceiver) Start(ctx context.Context, host component.Host) e
 }
 
 func (r *telemetryAPIReceiver) Shutdown(ctx context.Context) error {
+	close(r.stopCh)
+	r.wg.Wait()
+
+	var errs []error
+
+	if r.exportInterval > 0 {
+		if err := r.flushMetrics(ctx); err != nil {
+			r.logger.Error("error while flushing metrics", zap.Error(err))
+			errs = append(errs, err)
+		}
+	}
+
 	if r.httpServer != nil {
 		if err := r.httpServer.Shutdown(ctx); err != nil {
-			return err
+			r.logger.Error("error shutting down http server", zap.Error(err))
+			errs = append(errs, err)
 		}
+	}
+
+	return errors.Join(errs...)
+}
+
+func (r *telemetryAPIReceiver) startMetricsExporter() {
+	defer r.wg.Done()
+	ticker := time.NewTicker(r.exportInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			if err := r.flushMetrics(context.Background()); err != nil {
+				r.logger.Error("error while flushing metrics", zap.Error(err))
+			}
+		case <-r.stopCh:
+			return
+		}
+	}
+}
+
+func (r *telemetryAPIReceiver) flushMetrics(ctx context.Context) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return r.flushMetricsLocked(ctx)
+}
+
+func (r *telemetryAPIReceiver) flushMetricsLocked(ctx context.Context) error {
+	metric := pmetric.NewMetrics()
+	resourceMetric := metric.ResourceMetrics().AppendEmpty()
+	r.resource.CopyTo(resourceMetric.Resource())
+	scopeMetric := resourceMetric.ScopeMetrics().AppendEmpty()
+	scopeMetric.Scope().SetName(scopeName)
+	scopeMetric.SetSchemaUrl(semconv.SchemaURL)
+
+	ts := pcommon.NewTimestampFromTime(time.Now())
+	r.faaSMetricBuilders.coldstartsMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.errorsMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.timeoutsMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.initDurationMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.memUsageMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.invocationsMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.invokeDurationMetric.AppendDataPoints(scopeMetric, ts)
+
+	if metric.MetricCount() > 0 && r.nextMetrics != nil {
+		return r.nextMetrics.ConsumeMetrics(ctx, metric)
 	}
 	return nil
 }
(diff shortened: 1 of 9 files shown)


wpessers:
Good stuff! Getting back to this later today but overall seems good! 🚀

wpessers (commented):
I've been thinking about this change for a while now. And I know this PR is actually in response to my earlier comment on the previous PR for adding telemetryAPI metrics support: https://github.com/open-telemetry/opentelemetry-lambda/pull/2066#discussion_r2710282718

But I think I've come around on the idea... Maybe it's fine to keep the behaviour as is. Because the metrics are still the same / correct whether you flush them once or multiple times. And probably the volume of metrics is acceptable so that we don't really need to be doing this aggregation before flushing?

If sending multiple metrics "exports" from the collector to your observability backend of choice is a concern, we have the `batch` processor for that, that can be configured to be used inside the telemetry pipelines of the collector.

TLDR: I'm thinking it's actually fine to always flush the metrics, the change (although nicely, cleanly implemented) introduces some extra complexity that feels a little bit unnecessary. What do you think @herin049 ? It's possible that I'm missing some nuances / reasons why it might still be a good idea to add this anyway.

herin049:
> I've been thinking about this change for a while now. And I know this PR is actually in response to my earlier comment on the previous PR for adding telemetryAPI metrics support: [#2066 (comment)](https://github.com/open-telemetry/opentelemetry-lambda/pull/2066#discussion_r2710282718)
> 
> But I think I've come around on the idea... Maybe it's fine to keep the behaviour as is. Because the metrics are still the same / correct whether you flush them once or multiple times. And probably the volume of metrics is acceptable so that we don't really need to be doing this aggregation before flushing?
> 
> If sending multiple metrics "exports" from the collector to your observability backend of choice is a concern, we have the `batch` processor for that, that can be configured to be used inside the telemetry pipelines of the collector.
> 
> TLDR: I'm thinking it's actually fine to always flush the metrics, the change (although nicely, cleanly implemented) introduces some extra complexity that feels a little bit unnecessary. What do you think @herin049 ? It's possible that I'm missing some nuances / reasons why it might still be a good idea to add this anyway.

@wpessers Sorry for the late reply. I do believe that we still want to enable this feature since the `batch` processor does not actually do any "aggregation" of metrics in the sense that while it does batch calls to downstream exporters, it still results in the same number of metric data-points being created. This can be very problematic for many observability vendors because the billing model is based on the number of metric data-points ingested. By accumulating these metrics in memory and exporting periodically, we reduce the number of data-points at the cost of slightly less granular data (e.g. an export interval of 30s gives you a maximum resolution of 30s).

On a somewhat related note, the issue I described above can make user-defined metrics inside Lambda very expensive based on my experience at my employer, because every Lambda invocation flushes all metrics. Unfortunately, it doesn't seem like there are many good processors in the collector or collector-contrib repos that easily allow one to "downsample" the frequency of data-points generated, since it is assumed that aggregation should be primarily done in the SDK. Unfortunately, with Lambda, we don't have many options since we don't know when a Lambda instance will be torn down. I'm actually working on introducing some changes that will allow us to opt-in to not exporting on every invocation, and instead flush periodically while also handling the `SIGTERM` signal to flush remaining metrics before the Lambda instance is terminated.

wpessers:
> the `batch` processor does not actually do any "aggregation" of metrics in the sense that while it does batch calls to downstream exporters, it still results in the same number of metric data-points being created. This can be very problematic for many observability vendors because the billing model is based on the number of metric data-points ingested. By accumulating these metrics in memory and exporting periodically, we reduce the number of data-points at the cost of slightly less granular data (e.g. an export interval of 30s gives you a maximum resolution of 30s).

Yeah now that I think about it, this same reasoning is how I eventually came to that conclusion and why I put that comment in the previous pr review. Going to give it a final look tomorrow and probably merge it so it can be included in the release this weekend.

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -106,9 +114,60 @@ func (r *telemetryAPIReceiver) Start(ctx context.Context, host component.Host) e
...
+	ts := pcommon.NewTimestampFromTime(time.Now())
+	r.faaSMetricBuilders.coldstartsMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.errorsMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.timeoutsMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.initDurationMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.memUsageMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.invocationsMetric.AppendDataPoints(scopeMetric, ts)
+	r.faaSMetricBuilders.invokeDurationMetric.AppendDataPoints(scopeMetric, ts)
+
+	if metric.MetricCount() > 0 {
```
wpessers: Edge case here but think of the following scenario:
- User configures delta temporality
- User removes their metrics pipeline but forgets to remove the delta temporality config
Their collector config could look something like this:
```yaml
receivers:
  telemetryapi:
    types:
      - platform
    metrics_temporality: "delta"
...
service:
  pipelines:
    traces:
      receivers: [telemetryapi]
      exporters: [otlp]
```
This leds to `r.nextMetrics` being nil (see [factory.go](https://github.com/open-telemetry/opentelemetry-lambda/pull/2114/changes#diff-e24bc76f95645dc2aa7e5057e8b295f1584f9fed420b120fa4a30dac2fc84d51R55) where the `createMetricsReceiver` method is not called in this case).

Whenever the [ticker fires](https://github.com/open-telemetry/opentelemetry-lambda/pull/2114/changes#diff-274cc9dddab49c0035b0c0a49cc0d4cd10b7390a6ac2de41fcf772cf6a419c59R133) and `flushMetricsLocked` method is now called, the `AppendDataPoints` methods will **not** early return. Because for delta temporality specifically the export variable will be `true`: 
https://github.com/open-telemetry/opentelemetry-lambda/blob/4141c02870c3b666b2bf9b5a6e678ffe5d0c5391/collector/receiver/telemetryapireceiver/metric_builder.go#L120

I believe that is problematic, because now following calls inside that same `AppendDataPoints` method are causing an empty `Metric` struct to be appended to `scopeMetrics`:
https://github.com/open-telemetry/opentelemetry-lambda/blob/4141c02870c3b666b2bf9b5a6e678ffe5d0c5391/collector/receiver/telemetryapireceiver/metric_builder.go#L134

So the check for `metric.MetricCount() > 0` here will always evaluate to `true` when using delta temporality. But since we no longer have that metrics pipeline, the `r.nextMetrics.ConsumeMetrics(ctx, metric` call below will panic because `r.nextMetrics` is `nil`.

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -106,9 +114,60 @@ func (r *telemetryAPIReceiver) Start(ctx context.Context, host component.Host) e
 }
 
 func (r *telemetryAPIReceiver) Shutdown(ctx context.Context) error {
+	close(r.stopCh)
```
wpessers, line 141: Should we be shutting down the httpServer here first?

```suggestion
	r.httpServer.Shutdown(ctx)

	close(r.stopCh)
```

collector/receiver/telemetryapireceiver/receiver.go
wpessers: Also, even when properly using an actual metrics pipeline We run the risk of just producing metrics with 0 datapoints when temporality is delta, which seems useless / wasteful.

wpessers (changes requested):
Sorry for taking so long, but some final (I swear) remarks. Please check @herin049 and thank you for the patience! 🙏

collector/receiver/telemetryapireceiver/receiver.go
herin049: Good catch, I will update the `AppendDataPoints` methods to return early if there are no datapoints to append. For delta temporality, after each call to `AppendDataPoints` we clear the datapoints map, so this should not result in any extraneous datapoints being generated. Furthermore, I'll also add a nil check to ensure we never call `ConsumeMetrics` on a nil metrics consumer.

collector/receiver/telemetryapireceiver/receiver.go
herin049: My thought is that exporting metrics should be our highest priority so I've actually opted to try to flush metrics first and then shutdown the Http server. It isn't a huge issue if the Http server shutdown fails, but metric flushing failure will result in data loss. We could spawn separate goroutines to do the work in parallel but in my opinion this is a bit overkill. Let me know what you think.

wpessers: Sounds a bit overkill indeed, I was just thinking about this (unlikely?) scenario:

Lambda Execution Env goes into shutdown phase
-> background goroutine exits
-> metrics are flushed

And the httpServer is still running while this is ongoing, could this mean we still receive incoming requests and then those new metrics would never be flushed since the "final" flush has already happened?

herin049: Good point, after thinking about this a bit more, it's probably better to flush after the HTTP server is shutdown just in case some extra metrics were being exported. We should still be able to stop the background goroutine first, we just need to make sure that we do the flush at the very last step. Will update the PR

wpessers: That sounds great, as I said probably an edge case but just making sure

wpessers (approved):
LGTM! 🚀

wpessers:
@herin049 some merge conflicts since I merged your other pr first. As soon as they're resolved I'll merge this one!

herin049:
Done! @wpessers

Issue #2079, opened 2025-12-17
Setup integration tests for each lambda layer


wpessers:
**Is your feature request related to a problem? Please describe.**
There have been some problems in the past with unexpected behaviour of the instrumentation layers. Very recently for example an upgrade of a transitive dependency in the nodejs layer lead to issues with instrumentation libs no longer working inside the bundled code we deploy through the lambda layer, see: https://github.com/nodejs/require-in-the-middle/issues/120 and https://github.com/open-telemetry/opentelemetry-js-contrib/issues/3219

Not long after that the node 24 runtime was released and again we had some issues: https://github.com/open-telemetry/opentelemetry-lambda/issues/2034

At other times we've mentioned preferring to "smoke-test" some change to make sure stuff like runtime upgrades etc. will work nicely with the layers: https://github.com/open-telemetry/opentelemetry-lambda/pull/2056#pullrequestreview-3533941134

**Describe the solution you'd like**
An integration test suite to quickly "smoke-test" each of the language-specific layers, as well as the collector layer. We can then run the relevant workflow(s) when a pr is submitted. Not certain how easy it is to set that up as a workflow that has to be approved, but for example only run the nodejs integration test when only nodejs code has been impacted.

Alternatively the integration tests could only be run during the release process as well, instead of on every pr.

Since we have an AWS account for publishing the lambda layers, we should be able to use that to run these integration tests as well. It should be relatively simple:
1. setup infra (lambda function with layer(s), cloudwatch log group,...)
2. invoke lambda function
3. check debug logs to assert the required telemetry has been generated / exported
4. destroy infra

**Describe alternatives you've considered**
Currently I have some small repos for testing nodejs and python lambda layers. The samples for each layer are also a nice starting point. But it's annoying to have to deploy and manually check when this can all be automated.

**TODO**

- [ ] Add tests for all runtime versions of each language
- [ ] Evaluate sending spans to CW OTLP endpoint and using transaction search for the assertions
- [ ] Look into checks for other telemetry signals (metrics and logs)

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

herin049:
[...]

PR #2122, opened 2026-02-04
Properly create telemetry factory when initializing the lambda collector

johnbley:
Per https://github.com/open-telemetry/opentelemetry-collector/issues/14003 `otelcol.Factories.Telemetry` is now a required field.  I noticed this because our (Splunk/Cisco's) CI test was failing because the lambda collector failed to start, with the error message
```
    "error": "failed to get config: cannot unmarshal the configuration: otelcol.Factories.Telemetry must not be nil. For example, you can use otelconftelemetry.NewFactory to build a telemetry factory"
```
This adds the necessary telemetry factory to both places it should be initialized, and additionally drops the now-defunct `telemetry:metrics:address` section of the configuration yml, because that was also causing a failure-to-start.  In my local test the extension now properly starts.

### DIFF
--- a/collector/config.yaml
+++ b/collector/config.yaml
@@ -18,6 +18,3 @@ service:
     metrics:
       receivers: [otlp]
       exporters: [debug]
-  telemetry:
-    metrics:
-      address: localhost:8888
--- a/collector/go.mod
+++ b/collector/go.mod
@@ -238,9 +238,11 @@ require (
 	go.opentelemetry.io/collector/semconv v0.128.0 // indirect
 	go.opentelemetry.io/collector/service v0.143.0 // indirect
 	go.opentelemetry.io/collector/service/hostcapabilities v0.143.0 // indirect
+	go.opentelemetry.io/contrib/bridges/otelzap v0.13.0 // indirect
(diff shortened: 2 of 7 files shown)


atoulme (approved):

atoulme:
Maybe consider adding a test running the collector?

johnbley:
> Maybe consider adding a test running the collector?

I'd love to!  In a previous discussion it turned up that we (otel) don't currently have any organizational ability to deploy to a test lambda container in CI (billing, auth, etc. etc.).  I've asked around internally (Splunk/Cisco) about that but it's always been a low priority to tackle.  So I kinda of have to use Splunk CI to test things. 😞 

We could fake out the registration server (I have a codex-generated stub for that) to test locally and it would have caught this but wouldn't be as good as a proper deployment to a real lambda environment.

atoulme:
re adding a test, looks like a pendant to issue https://github.com/open-telemetry/opentelemetry-lambda/issues/2079

johnbley:
In case it needs to be called out more explicitly, this bug is preventing the lambda collector from starting, which prevents lambda functions which use the collector from starting, which breaks applications, which is a really bad outcome for an instrumentation system.

wpessers (approved):

wpessers:
Thank you for the fix @johnbley

PR #2127, opened 2026-02-09
feat(lambda): write cloud.account.id symlink from extension

RaphaelManke:
## Summary

- After the Lambda Extensions API `Register()` call, the extension now creates a symlink at `/tmp/.otel-aws-account-id` with the AWS account ID as the target
- Requests the `accountId` feature via the `Lambda-Extension-Accept-Feature: accountId` header
- Handles execution environment reuse by removing stale symlinks before creating new ones
- Fails silently (warn log only) if symlink creation fails

This enables all OTel SDK Lambda resource detectors to read `cloud.account.id` via `readlink()` without needing IAM calls.

## Changes

- `collector/internal/extensionapi/client.go` — Parse `accountId` from Register response, send Accept-Feature header
- `collector/internal/lifecycle/manager.go` — Create symlink after registration
- Tests for both header/parsing and symlink creation/cleanup

## Related PRs

SDK-side implementations that read this symlink (will be linked once created).

## Test plan

- [x] Unit test: symlink created with correct target
- [x] Unit test: leading zeros preserved
- [x] Unit test: stale symlink replaced
- [x] Unit test: empty accountID skipped silently
- [x] Unit test: Accept-Feature header sent correctly
- [x] Unit test: accountId parsed from response JSON


serkan-ozal (approved):

collector/internal/lifecycle/manager.go
```diff
@@ -33,6 +33,8 @@ import (
 	"github.com/open-telemetry/opentelemetry-lambda/collector/lambdacomponents"
 )
 
+const accountIDSymlinkPath = "/tmp/.otel-account-id"
```
herin049: Small nit:
```suggestion
const accountIDSymlinkPath = "/tmp/.otel-aws-account-id"
```

collector/internal/lifecycle/manager.go
```diff
@@ -178,3 +182,14 @@ func (lm *manager) notifyEnvironmentShutdown() {
...
 }
+
+func writeAccountIDSymlink(logger *zap.Logger, accountID string) {
+	if accountID == "" {
+		return
+	}
+	// Remove any stale symlink from a previous execution environment reuse.
+	os.Remove(accountIDSymlinkPath)
+	if err := os.Symlink(accountID, accountIDSymlinkPath); err != nil {
+		logger.Debug("Failed to create account ID symlink", zap.Error(err))
```
herin049: Maybe we should make this a Warning log instead of debug?

collector/internal/lifecycle/manager.go
herin049: Looks like you've already made the PRs for all the contrib repos, so to save the effort of changing the name everywhere, `.otel-account-id` is probably fine.

wpessers: Agree since all the other PRs have already been created, we can leave it as is.

wpessers (approved):
Thank you for the contribution. Will merge as soon as the open remarks have been addressed.

collector/internal/lifecycle/manager.go
RaphaelManke: Thanks for the suggestion\! I went ahead and renamed it to `.otel-aws-account-id` across all repos — it's a better name since this is AWS-specific, and the rename was straightforward to do everywhere at once.

collector/internal/lifecycle/manager.go
RaphaelManke: Good call, changed to `Warn`.

wpessers (approved):
Thanks @RaphaelManke 🚀 This will be included in the next release, should be either tomorrow or friday!

PR #2157, opened 2026-03-07
Add cloud.account.id resource attribute to language wrapper scripts

RaphaelManke:
## Summary

- Reads the AWS account ID from the `/tmp/.otel-aws-account-id` symlink (created by the collector extension during registration, see #2127) in each language's wrapper script
- Appends `cloud.account.id` as a resource attribute to `OTEL_RESOURCE_ATTRIBUTES` if not already set
- Covers all languages: Node.js, Java (javaagent + wrapper), Python, and Ruby

## How it works

```
┌─────────────────────────────────────────────────────────────────────┐
│                        Lambda Init Phase                            │
│                                                                     │
│  ┌──────────────────────────────────────────────────────────────┐   │
│  │  1. Collector Extension starts                               │   │
│  │     └─► Calls extensionClient.Register()                     │   │
│  │         └─► Receives AccountID in response                   │   │
│  │             └─► writeAccountIDSymlink()              (#2127) │   │
│  │                 Creates: /tmp/.otel-aws-account-id            │   │
│  │                          -> "123456789012" (symlink target)   │   │
│  └──────────────────────────────────────────────────────────────┘   │
│                              │                                      │
│                              ▼                                      │
│  ┌──────────────────────────────────────────────────────────────┐   │
│  │  2. Lambda Runtime starts                                    │   │
│  │     └─► AWS_LAMBDA_EXEC_WRAPPER invokes wrapper script       │   │
│  │         ┌────────────────────────────────────────────────┐   │   │
│  │         │  otel-handler / otel-instrument       (this PR)│   │   │
│  │         │                                                │   │   │
│  │         │  ACCOUNT_ID=$(readlink /tmp/.otel-aws-account-id)  │   │
│  │         │  export OTEL_RESOURCE_ATTRIBUTES=              │   │   │
│  │         │    "cloud.account.id=$ACCOUNT_ID,..."          │   │   │
│  │         └────────────────────────────────────────────────┘   │   │
│  │         └─► Language SDK initializes with resource attributes │   │
│  └──────────────────────────────────────────────────────────────┘   │
│                              │                                      │
│                              ▼                                      │
│  ┌──────────────────────────────────────────────────────────────┐   │
│  │  3. Function handler executes                                │   │
│  │     └─► All spans, logs, metrics from the SDKs include cloud.account.id    │   │
│  └──────────────────────────────────────────────────────────────┘   │
└─────────────────────────────────────────────────────────────────────┘
```

**Key insight:** The extension always registers *before* the runtime starts, so the symlink is guaranteed to exist by the time the wrapper script runs.

## Motivation

The collector extension writes the AWS account ID as a symlink during registration (#2127). Since the language-specific wrapper scripts (`otel-handler` / `otel-instrument`) execute after the extension has registered, they can reliably read this symlink and propagate the account ID to the language SDKs via `OTEL_RESOURCE_ATTRIBUTES`.

This ensures `cloud.account.id` is present on all telemetry signals produced by the language SDKs without requiring users to manually configure it.

## Test plan

- [x] Deploy Lambda with collector extension and verify `cloud.account.id` appears on spans/logs/metrics for each language runtime
- [x] Verify that if `cloud.account.id` is already set in `OTEL_RESOURCE_ATTRIBUTES`, it is not overwritten
- [x] Verify graceful degradation when symlink does not exist (error logged to stderr, execution continues)


wpessers (approved):
Looks good 🚀  I'll be awaiting the results of your testing to make sure since you added the testing plan with open tasks to this PR.

RaphaelManke:
Tested it and works as expected. 
I also removed the error log as it would produce unnecessary messages when for example the lamguuage layer is used with and old or other collector and extended the README documenting the env vars from the wrapper script

wpessers:
the check-links action failure is unrelated to your pr, seems like the `include-fragments` config should now be `include_fragments` instead here:
https://github.com/open-telemetry/opentelemetry-lambda/blob/4141c02870c3b666b2bf9b5a6e678ffe5d0c5391/.github/lychee.toml#L1

I can fix it if you want, or if you'd rather do it yourself that's fine too. Up to you 😄

wpessers:
So if you prefer keeping the link to that pr in there, that's fine btw @RaphaelManke. It seems like it's just the changed config var name in lychee that is causing this to fail.

RaphaelManke:
No, its fine without the link 😃

python/src/otel/otel_sdk/otel-instrument
```diff
@@ -146,6 +146,14 @@ else
     export OTEL_RESOURCE_ATTRIBUTES="$LAMBDA_RESOURCE_ATTRIBUTES,$OTEL_RESOURCE_ATTRIBUTES";
 fi
 
+CLOUD_ACCOUNT_ID=$(readlink /tmp/.otel-aws-account-id 2>/dev/null || true)
+if [ -n "$CLOUD_ACCOUNT_ID" ]; then
```
herin049, line 150: Is there a reason we are deviating from the strategy above for initializing the resource attributes? Generally, I think that the single `if` statement is a bit more readable.

wpessers: nice catch, agree

maxday:
Thanks for working on this! Dropping a file at the root of `/tmp` could catch users off guard, imagine code that iterates over everything in /tmp. Would it make sense to add an env var to configure the file path, so it can be written somewhere more specific like `/tmp/otel/.my_custom_location` ?

RaphaelManke:
Do you have a suggestion for an env var name?

maxday:
I'm terrible at naming, what about : `OTEL_LAMBDA_AWS_ACCOUNT_ID_FILE_LOCATION` 
too long?

PR #2172, opened 2026-03-15
feat(python): upgrade aws lambda instrumentation and adapt tests

wpessers:
Some changes in the latest release of the aws lambda instrumentation package for python require us to adapt the tests here as well.

See: https://github.com/open-telemetry/opentelemetry-python-contrib/commit/1eca3e6b8eba1683d77c00f975215ee76ce198d1

The invocation span now has the lambda function name as `span.name`.


herin049 (approved):
Thanks! And apologies I should have communicated these changes better 😆

wpessers:
No worries 😄

PR #2182, opened 2026-03-17
build(deps): bump the opentelemetry-deps-collector  group

wpessers:
Long overdue go version upgrade on top of dependabot changes, fixes CI issues at #2167


collector/processor/coldstartprocessor/go.mod
```diff
@@ -29,20 +29,20 @@ require (
...
+	go.opentelemetry.io/collector/consumer/xconsumer v0.147.0 // indirect
+	go.opentelemetry.io/collector/featuregate v1.53.0 // indirect
+	go.opentelemetry.io/collector/internal/componentalias v0.147.0 // indirect
+	go.opentelemetry.io/collector/pdata/pprofile v0.147.0 // indirect
+	go.opentelemetry.io/collector/pdata/testdata v0.147.0 // indirect
+	go.opentelemetry.io/collector/pipeline v1.53.0 // indirect
+	go.opentelemetry.io/collector/processor/xprocessor v0.147.0 // indirect
+	go.opentelemetry.io/otel v1.42.0 // indirect
+	go.opentelemetry.io/otel/metric v1.42.0 // indirect
 	go.opentelemetry.io/otel/sdk v1.40.0 // indirect
```
herin049: Should this also be `v1.42.0`?

wpessers: Good point, the current collector packages `go.opentelemetry.io/collector/processor/processorhelper` and `go.opentelemetry.io/collector/component` that these are transitive dependencies of seem to actually depend on an earlier version `1.40.0`. But some quick testing with the latest `1.42.0` shows the upgrade is backwards compatible. So even though it's a bit weird imo that Dependabot doesn't take this into account, I guess we'll just upgrade the sdk package for now. Good catch!

herin049 (approved):

PR #2186, opened 2026-03-20
fix(disablequeuedretryconverter): add renamed exporters

### LINKED ISSUE #2185: disablequeuedretryconverter doesn't recognize renamed otlp_http exporter
**Describe the bug**
The `disablequeuedretryconverter` doesn't recognize the `otlp_http` exporter, which was renamed from `otlphttp` in collector v0.144.0. The [converter's exporter list](https://github.com/open-telemetry/opentelemetry-lambda/blob/layer-collector/0.20.0/collector/internal/confmap/converter/disablequeuedretryconverter/converter.go) only contains `otlphttp`, so it never disables the sending queue for `otlp_http`. This leaves the async queue active in Lambda, causing spans to be dropped at shutdown.

**Steps to reproduce**
1. Use collector Lambda layer v0.20.0
2. Configure the `otlp_http` exporter:
```yaml
exporters:
  otlp_http:
    endpoint: http://some-endpoint:4318
```
3. Deploy to a Lambda function
4. After a period of inactivity, Lambda shuts down the environment
5. The async sending queue tries to drain during shutdown but fails

The converter should recognize `otlp_http` and disable its sending queue, matching the behavior for the old `otlphttp` name.

**What did you see instead?**
The sending queue remains active. At shutdown, spans are dropped with:
```
"msg":"Exporting failed. Dropping data."
"caller":"internal/queue_sender.go:50"
"error":"interrupted due to shutdown: failed to make an HTTP request: Post \"...\": context deadline exceeded"
"dropped_items":1
```

- Lambda collector layer: `v0.20.0`
- OpenTelemetry Collector: `v0.145.0`

Go (custom collector config, no language layer — using the collector layer only)

**Additional context**
Workaround is to manually disable the sending queue:
```yaml
exporters:
  otlp_http:
    sending_queue:
      enabled: false
```

The same issue likely affects other exporters that were renamed in v0.144.0+.

Hoffs:
fixes #2185

### DIFF
--- a/collector/internal/confmap/converter/disablequeuedretryconverter/converter_test.go
+++ b/collector/internal/confmap/converter/disablequeuedretryconverter/converter_test.go
@@ -48,10 +48,17 @@ func TestConvert(t *testing.T) {
 			err:      nil,
 		},
 		{
-			name:     "many queuing exporters",
-			conf:     confmap.NewFromStringMap(map[string]any{"exporters": map[string]any{"otlphttp": map[string]any{}, "otlp": map[string]any{}}}),
-			expected: confmap.NewFromStringMap(map[string]any{"exporters": map[string]any{"otlphttp": map[string]any{"sending_queue": map[string]any{"enabled": false}}, "otlp": map[string]any{"sending_queue": map[string]any{"enabled": false}}}}),
-			err:      nil,
+			name: "many queuing exporters",
+			conf: confmap.NewFromStringMap(map[string]any{"exporters": map[string]any{"otlphttp": map[string]any{}, "otlp": map[string]any{}, "otlp_http": map[string]any{}, "otlp_grpc": map[string]any{}}}),
+			expected: confmap.NewFromStringMap(map[string]any{
+				"exporters": map[string]any{
+					"otlphttp":  map[string]any{"sending_queue": map[string]any{"enabled": false}},
+					"otlp":      map[string]any{"sending_queue": map[string]any{"enabled": false}},
+					"otlp_http": map[string]any{"sending_queue": map[string]any{"enabled": false}},
+					"otlp_grpc": map[string]any{"sending_queue": map[string]any{"enabled": false}},
+				},
+			}),
+			err: nil,
 		},
 	} {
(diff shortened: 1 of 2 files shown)


collector/internal/confmap/converter/disablequeuedretryconverter/converter_test.go
```diff
@@ -53,6 +53,12 @@ func TestConvert(t *testing.T) {
 			expected: confmap.NewFromStringMap(map[string]any{"exporters": map[string]any{"otlphttp": map[string]any{"sending_queue": map[string]any{"enabled": false}}, "otlp": map[string]any{"sending_queue": map[string]any{"enabled": false}}}}),
 			err:      nil,
 		},
+		{
+			name:     "renamed exporters",
+			conf:     confmap.NewFromStringMap(map[string]any{"exporters": map[string]any{"otlp_http": map[string]any{}, "otlp_grpc": map[string]any{}}}),
+			expected: confmap.NewFromStringMap(map[string]any{"exporters": map[string]any{"otlp_http": map[string]any{"sending_queue": map[string]any{"enabled": false}}, "otlp_grpc": map[string]any{"sending_queue": map[string]any{"enabled": false}}}}),
+			err:      nil,
+		},
```
wpessers: I'm not a fan of adding this separate test for this. We should not encode "why" we are testing, instead we should encode "what behaviour" we are testing. Therefore I'd suggest simply adding these exporters to the existing `"many queuing exporters"` test above.

wpessers (changes requested):

collector/internal/confmap/converter/disablequeuedretryconverter/converter_test.go
Hoffs: yeah, totally, updated

wpessers: Thank you!🙏

wpessers (approved):

wpessers:
Thank you for the contribution @Hoffs! 🚀

PR #2201, opened 2026-03-25
doc(approvers): add Raphael Manke <@RaphaelManke> as approver

wpessers:


README.md
```diff
@@ -128,15 +128,16 @@ See the [Contributing Guide](CONTRIBUTING.md) for details.
...
 
 ### Approvers
 
-- [Ivan Santos](https://github.com/pragmaticivan)
-- [Lukas Hering](https://github.com/herin049)
-- [Maxime David](https://github.com/maxday)
+- [Ivan Santos](https://github.com/pragmaticivan), EverQuote
+- [Lukas Hering](https://github.com/herin049), Oracle
+- [Maxime David](https://github.com/maxday), AWS
+- [Raphael Manke](https://github.com/raphaelmanke), Dash0
```
tylerbenson: In the past we've opted to leave off employer from approvers since they tend to get stale, but not opposed to it if you think it's important.

tylerbenson (approved):
Welcome @RaphaelManke!

README.md
wpessers: Ah, I saw this was being done in the collector repo and though it might make sense to update ours as well. But good point about it becoming stale. Don't think it's **that** important, so will revert this commit!

PR #2239, opened 2026-04-09
chore(nodejs): use native npm workspaces instead of lerna

wpessers:
We don't use any lerna features, it's simply used as a script runner that uses the underlying npm workspaces. Therefore I just got rid of it entirely in favour of native npm workspace usage.


pragmaticivan (approved):
LGTM, but was not sure if we ever want workspaces without tests, but --if-present there makes sense to me. One could argue the same applies to all commands, but not a blocker.

wpessers:
Good point, I had to add the flag because the sample app workspace doesn't have any tests. 😅

PR #2242, opened 2026-04-11
chore: remove misspell

thompson-tomo:
This removes the deprecated misspell which is purely used for performing spell check of text.


wpessers (approved):

wpessers:
Thank you @thompson-tomo!

Issue #2173, opened 2026-03-16
telemetryapi receiver: function log records not fully parsed (severity, custom fields, and body extraction)


RaphaelManke:
**Describe the bug**

The `telemetryapireceiver` does not fully parse function log records. Severity, timestamp, request ID, and application-defined custom fields are silently dropped depending on the log format and how the application emits logs. In the worst case (raw JSON written to stdout with `process.stdout.write`), the entire log body disappears — the record is emitted with no `body` and no attributes.

**Steps to reproduce**

Deploy a Node.js Lambda function with the following handler:

```typescript
export const handler = async () => {
    console.log("Hello World")                                                                                       // case 1
    console.log(JSON.stringify({msg: "JSON stringified message", userId: "user-123", action: "login", durationMs: 42}))  // case 2
    process.stdout.write("Plain text written directly to stdout" + "\n")                                            // case 3
    process.stdout.write(JSON.stringify({msg: "JSON stringified text written directly to stdout", userId: "user-123", action: "login", durationMs: 42}) + "\n") // case 4
    await new Promise(r => setTimeout(r, 1000));
}
```

Invoke the function and observe the log records exported by the telemetry API receiver.

**What did you expect to see?**

All four cases should produce consistent, well-structured OTel log records — independent of how the application logged the data — with `body` containing the message text, `severityNumber` populated from the level, `faas.invocation_id` from the request ID, and any custom fields (e.g. `userId`, `action`, `durationMs`) preserved as log attributes.

**What did you see instead?**

### Plain text log format (default) — all cases broken

| # | body | severityNumber | faas.invocation_id |
|---|------|----------------|-------------------|
| 1 | `"2026-03-15T...\tREQ_ID\tINFO\tHello World\n"` (raw prefix embedded) | 0 ❌ | absent ❌ |
| 2 | `"2026-03-15T...\tREQ_ID\tINFO\t{\"msg\":\"...\",\"userId\":\"user-123\",...}\n"` (JSON unparsed) | 0 ❌ | absent ❌ |
| 3 | `"Plain text written directly to stdout\n"` | 0 ❌ | absent ❌ |
| 4 | `"{\"msg\":\"...\",\"userId\":\"user-123\",...}\n"` (JSON unparsed, custom fields unreachable) | 0 ❌ | absent ❌ |

### JSON log format (`AWS_LAMBDA_LOG_FORMAT=JSON`) — case 4 is completely empty

[...]

PR #2251, opened 2026-04-14
feat(collector/receiver): support extra fields in structured JSON function logs

tomsobpl:
## Summary

When a Lambda function emits structured JSON logs (i.e., a log line with a `message` field), the `telemetryapireceiver` now extracts any additional fields from the JSON object and maps them to log record attributes.

Previously, only the `message` field was captured from structured JSON log records. Any extra fields present in the JSON object (e.g., `userId`, `orderId`, custom business data) were silently discarded.

```json
{ "message": "order processed", "orderId": "123", "userId": "abc", "level": "info" }
```

**Before:**
→ Log record body: "order processed", no extra attributes.

**After:**
→ Log record body: "order processed", attributes: orderId = "123", userId = "abc".

## Changes
- `receiver.go`: After setting the log record body from the message field, iterate over all remaining keys in the parsed JSON record and add them as log record attributes. The well-known fields (level, message, requestId, timestamp) are skipped since they are already handled separately.
- `receiver_test.go`: Extended test coverage for JSON logs containing extra fields, covering nested values, arrays, and numeric types.

**Notes:**
- The attribute key names are taken as-is from the JSON log (no prefix added), keeping the mapping straightforward.
- Values that cannot be converted via pcommon.Value.FromRaw emit a warning and are skipped rather than failing the entire record.


collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -490,6 +490,19 @@ func (r *telemetryAPIReceiver) createLogs(slice []event) (plog.Logs, error) {
...
+				for key, value := range record {
+					switch key {
+					case "level", "message", "requestId", "timestamp":
+						continue
+					default:
+						attr, _ := logRecord.Attributes().GetOrPutEmpty(key)
+						if err := attr.FromRaw(value); err != nil {
+							r.logger.Warn(fmt.Sprintf("error creating attribute: %s", key), zap.Error(err))
+							continue
+						}
```
wpessers, line 504: If an error is returned out of the `FromRaw` call you log but shouldn't you also remove the attribute from the map?

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -490,6 +490,19 @@ func (r *telemetryAPIReceiver) createLogs(slice []event) (plog.Logs, error) {
...
 				logRecord.Body().SetStr(line)
+
+				for key, value := range record {
+					switch key {
+					case "level", "message", "requestId", "timestamp":
+						continue
+					default:
+						attr, _ := logRecord.Attributes().GetOrPutEmpty(key)
+						if err := attr.FromRaw(value); err != nil {
+							r.logger.Warn(fmt.Sprintf("error creating attribute: %s", key), zap.Error(err))
```
wpessers: `fmt.Sprintf` is okay but I'd prefer leveraging zap as much as possible here. Also not sure why we'd prefer logging this as a warning vs an error level but happy to hear your thoughts on that. I'd suggest something like:

```suggestion
							r.logger.Warn("Failed to convert field to attribute", zap.String("key", key), zap.Error(err))
```

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -490,6 +490,19 @@ func (r *telemetryAPIReceiver) createLogs(slice []event) (plog.Logs, error) {
 				}
 			} else if line, ok := record["message"].(string); ok {
 				logRecord.Body().SetStr(line)
+
+				for key, value := range record {
+					switch key {
+					case "level", "message", "requestId", "timestamp":
```
wpessers: You really should include `"type"` here as well. Look at: https://github.com/open-telemetry/opentelemetry-lambda/blob/e0f470f6c9d70dd52c6db5bb3805cf65d5b467f2/collector/receiver/telemetryapireceiver/receiver.go#L430
Every log record gets an attribute with key "type" and a value like "function" or "platform.start", = the event type from the Telemetry API. 

Now what if we have a log line originating from a user's lambda function code:
```json
{
  "message": "buy order",
  "type": "checkout",
  "orderId": "123"
}
```
So now in your newly added code below (in the default branch) you will simply overwrite the `type` attribute. And its value will now be `"checkout"` instead of `"function"`...

Another, maybe better, option would be to prefix the additional fields that are in a user's logs so we can keep both.

collector/receiver/telemetryapireceiver/receiver_test.go
```diff
@@ -479,6 +480,109 @@ func TestCreateLogs(t *testing.T) {
...
+				{
+					Time: "2026-02-26T20:15:32.000Z",
+					Type: "function",
+					Record: map[string]any{
+						"timestamp":              "2026-02-26T20:15:32.000Z",
+						"level":                  "INFO",
+						"requestId":              "79b4f56e-95b1-4643-9700-2807f4e6",
+						"message":                "Hello world, I am a function with extra data!",
+						"extraString":            "stringValue",
+						"extraNumber":            2217,
```
wpessers: This can never happen in production. a Record is always populated via `json.Unmarshal`, which always decodes JSON numbers as float64. Consider using `float64(2217)` as input so the test mirrors reality.

See: https://pkg.go.dev/encoding/json#Unmarshal

wpessers (changes requested):

collector/receiver/telemetryapireceiver/receiver.go
tomsobpl: I personally do not feel that after emiting an error level process should continue and that's why I chose warning here.

Probably it is more preferred to prioritize handling request correctly instead stopping it due to malformed log metadata. Warning developer that something was not correctly logged should be enough IMHO.

Anyway agree on the zap usage.

collector/receiver/telemetryapireceiver/receiver.go
tomsobpl: That was an oversight, good catch.

In the first place I was thinking about adding prefixes to the fields but after some thought I abandoned this option. Developers should be familiar with limitations and reserved fields and should handle that themselves on business logic level.

But maybe it's worth to update documentation and point out all the reserved internal fields with info that they will be prioritized over users metadata. WDYT?

collector/receiver/telemetryapireceiver/receiver.go
wpessers: fair point, let's keep it at warning level

collector/receiver/telemetryapireceiver/receiver.go
wpessers: Yeah I do agree with that, but indeed given you update the docs.

tomsobpl:
Hey @wpessers, I've updated the code and addressed all the feedback.

RaphaelManke:
Is this one solving https://github.com/open-telemetry/opentelemetry-lambda/issues/2173
Or at least partially?

tomsobpl:
> Is this one solving #2173 Or at least partially?

@RaphaelManke partially yes. It introduces support for extra logs metadata but without any transformation. Everything is in developers hands as this just copy 1:1 json logs metadata fields into otel attributes

collector/receiver/telemetryapireceiver/receiver.go
wpessers: nit, but I'd still prefer making use of zap over the `fmt.Sprintf` call.

wpessers: But your call if you prefer it as is, not going to hold up this PR because of it.

wpessers (approved):

wpessers:
LGTM, waiting to merge to allow @RaphaelManke to give feedback if he wants since the PR is relevant to that issue he still has open.

collector/receiver/telemetryapireceiver/receiver.go
tomsobpl: But it is changed to zap. I just didn't commit suggestion but pushed it directly with the rest of the changes :) https://github.com/open-telemetry/opentelemetry-lambda/pull/2251/changes#diff-274cc9dddab49c0035b0c0a49cc0d4cd10b7390a6ac2de41fcf772cf6a419c59R502

RaphaelManke:
Ok thanks for the clarification 🙂 I'm good with merging 🙂

PR #2288, opened 2026-05-11
fix(collector): use Telemetry API event time as metric timestamp

### LINKED ISSUE #2263: telemetryapi: Metric timestamps are delayed - set via date.Now() on lambda thaw
**Describe the bug**
The telemetryapireceiver stamps all FaaS metric data points (`faas.invocations`, `faas.errors`, `faas.coldstarts`, `faas.timeouts`, `faas.init_duration`, `faas.invoke_duration`, `faas.mem_usage`) with `time.Now()` at flush time rather than using the event's time field from the Telemetry API payload.

In Lambda's freeze-thaw execution model, time.Now() at flush time reflects when the runtime was thawed, not when the event occurred. This causes metric timestamps to be dislocated from the actual event time, especially after a period of inactivity where the environment was frozen.

The log pipeline (createLogs) already correctly parses the event's time field and uses it as the log record timestamp. The metric pipeline does not — flushMetricsLocked creates a single `time.Now()` timestamp and applies it to all data points, and recordMetrics does not pass event timestamps to the metric builder Add/Record calls.

**Steps to reproduce**
Deploy a Lambda function with the OpenTelemetry collector layer and the telemetryapi receiver enabled in a metrics pipeline
Configure a metrics backend that displays data point timestamps (e.g., Prometheus, Grafana)
Invoke the function, then wait several minutes for the Lambda environment to freeze
Invoke the function again and compare the metric data point timestamps with the log record timestamps for the same events

Metric data point timestamps matching the event occurrence time (the time field from the Telemetry API payload), consistent with how log record timestamps are already set.

**What did you see instead?**
Metric data point timestamps reflecting the flush/thaw time rather than the event time. After a freeze-thaw cycle, metric timestamps can be significantly later than the actual event time, while log timestamps from the same events are correct.

Version: 0_21_0 of the lambda layer

Just the collector layer with configuration for `telemetryapi` reciever

dmedovich:
- Track the latest Telemetry API event time observed in `recordMetrics` on the receiver (`lastEventTime`).
- Use `lastEventTime` as the data point timestamp in `flushMetricsLocked`, falling back to `time.Now()` only when no event has been observed yet (in which case `AppendDataPoints` exports nothing anyway).
- Access is serialized by the existing `r.mu`, which already protects both `recordMetrics` and `flushMetricsLocked`.

This mirrors the behavior of the log pipeline and resolves the metric/log timestamp divergence described in the linked issue.

## Testing

- Added `TestMetricTimestampMatchesEventTime`, which records events with a fixed RFC3339 time, performs a delayed flush, and asserts every Sum/Histogram data point's `Timestamp()` matches the event time rather than wall-clock `time.Now()`.
- `go test ./...` and `go vet ./...` pass for the package.

Fixes #2263

### DIFF
--- a/collector/receiver/telemetryapireceiver/receiver.go
+++ b/collector/receiver/telemetryapireceiver/receiver.go
@@ -241,12 +241,6 @@ func (r *telemetryAPIReceiver) httpHandler(w http.ResponseWriter, req *http.Requ
 		return
 	}
 
-	for i := range slice {
-		if t, err := time.Parse(time.RFC3339, slice[i].Time); err == nil {
-			slice[i].parsedTime = t
-		}
-	}
-
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
@@ -369,11 +363,7 @@ func (r *telemetryAPIReceiver) recordMetrics(slice []event) {
 			continue
 		}
 
-		t := el.parsedTime
-		if t.IsZero() {
-			t, _ = time.Parse(time.RFC3339, el.Time)
-		}
-		if !t.IsZero() {
+		if t, err := time.Parse(time.RFC3339, el.Time); err == nil {
 			if ets := pcommon.NewTimestampFromTime(t); ets > r.lastEventTime {
 				r.lastEventTime = ets
 			}
@@ -448,16 +438,13 @@ func (r *telemetryAPIReceiver) createLogs(slice []event) (plog.Logs, error) {
 		r.logger.Debug(fmt.Sprintf("Event: %s", el.Type), zap.Any("event", el))
 		logRecord := scopeLog.LogRecords().AppendEmpty()
 		logRecord.Attributes().PutStr("type", el.Type)
-		t := el.parsedTime
-		if t.IsZero() {
-			var err error
-			if t, err = time.Parse(time.RFC3339, el.Time); err != nil {
-				r.logger.Error("error parsing time", zap.Error(err))
-				return plog.Logs{}, err
-			}
+		if t, err := time.Parse(time.RFC3339, el.Time); err == nil {
+			logRecord.SetTimestamp(pcommon.NewTimestampFromTime(t))
+			logRecord.SetObservedTimestamp(pcommon.NewTimestampFromTime(time.Now()))
+		} else {
+			r.logger.Error("error parsing time", zap.Error(err))
+			return plog.Logs{}, err
 		}
-		logRecord.SetTimestamp(pcommon.NewTimestampFromTime(t))
-		logRecord.SetObservedTimestamp(pcommon.NewTimestampFromTime(time.Now()))
 		if record, ok := el.Record.(map[string]interface{}); ok {
 			requestId := r.getRecordRequestId(record)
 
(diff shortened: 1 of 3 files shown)


dmedovich:
@wpessers

wpessers:
Thank you for working on this @dmedovich! I will take some time tomorrow to take a closer look at your changes!

collector/receiver/telemetryapireceiver/receiver.go
```diff
@@ -359,6 +363,12 @@ func (r *telemetryAPIReceiver) recordMetrics(slice []event) {
 			continue
 		}
 
+		if t, err := time.Parse(time.RFC3339, el.Time); err == nil {
+			if ets := pcommon.NewTimestampFromTime(t); ets > r.lastEventTime {
+				r.lastEventTime = ets
+			}
+		}
```
wpessers: nit: unlikely to have performance impact, but `time.Parse(time.RFC3339, el.Time)` is now being done here and again in `createLogs` at: https://github.com/open-telemetry/opentelemetry-lambda/blob/65e61f7c55c175aa99217c220536b678d40682e7/collector/receiver/telemetryapireceiver/receiver.go#L431
with different error handling (`createLogs` aborts the batch on failure). Not blocking, but if you wanted to keep this aligned the parse could also move to `httpHandler`'s existing loop and feed both `recordMetrics` and `createLogs` a parsed value. The `httpHandler`'s loop I'm referring to: https://github.com/open-telemetry/opentelemetry-lambda/blob/65e61f7c55c175aa99217c220536b678d40682e7/collector/receiver/telemetryapireceiver/receiver.go#L243

wpessers:
@dmedovich one small nit, otherwise LGTM!

dmedovich:
@wpessers  good catch, thx.  Moved the parse to httpHandler so both recordMetrics and createLogs share the pre-parsed value. Added a fallback in both methods for cases where they're called directly (like  tests). Updated in the latest commit.

wpessers (commented):
Couple of remarks.
- First and foremost I appreciate the work that has gone into this and how quick you were to update after my previous feedback.
- Second, it's quite difficult for me to now review your follow-up fix since it seems like you amended your commit and force pushed. We always squash on merge, so please to help us reviewers out just commit your fixes in a new commit. It's easier for everyone involved. And now since I have some doubts about the solution we ended up with, it's hard for me to remember what exactly the initial fix looked like. Brings me to my third point.
- I'm not entirely convinced about the current solution. What is mostly sitting wrong with me is that there is now this fallback code in `receiver.go` that is dead code in production, and is merely there to accommodate the tests. Also it seems like what I asked in my original feedback didn't fully land. Instead of moving the parsing of the timestamp to the already existing loop, you added a second loop to do so. All this considered, I should've made it more clear that the nit I brought up was not to be taken as a "requirement". I much prefer contributors engage in discussion when they disagree, which I highly suspect you did at **some** point given the changes you've made to accommodate my feedback.

All this to say, I propose you revert back to the original solution you submitted in the PR before my feedback. As I said in my comment back then I don't think there's any performance impact and I'd argue it was more readable than what we have now. Again if you disagree, don't hesitate to say so!

dmedovich:
@wpessers  yeah, i agree we misunderstood each other a bit, and the second loop turned out to be not quite what it should have been. I should have clarified the criteria and implementation details on my side before jumping into it. I reverted the changes back to the original format in 2a74570. I'll be more careful with these kinds of details next time. Thanks for the patience!

wpessers:
@dmedovich thank you! I want to reiterate that most of all though, I really appreciate the effort you put into the fix and the quick follow ups! Merging this now 🚀

dmedovich:
@wpessers  big love big brother

wpessers (approved):

PR #2322, opened 2026-05-27
feat(ci): automate integration tests

### LINKED ISSUE #2079: Setup integration tests for each lambda layer

wpessers:
### Automated integration testing
resolves #2079 

### Description
Currently the PR includes a new GitHub workflow to run integration tests. It can be triggered manually, and also is already triggered automatically on tag pushes for the python layer. For an example of such a run: https://github.com/wpessers/opentelemetry-lambda/actions/runs/26541989902 Here the publish step fails, but that's simply because I don't have the necessary infra in my personal aws account that is being used to test this.

### TODO
Some stuff left to do before merging this PR:

- [x] Refactor tests to one parameterized test (`nodejs.test.ts` and `python.test.ts` are nearly identical)
- [x] Do the one-time cloudformation stack setup in CNCF AWS account
- [x] Add some documentation for the cloudformation stack setup
- [x] Add Java integration test
- [x] Add Ruby integration test
- [x] Trigger nodejs, java, ruby integration tests on tag push also


wpessers:
Succesful run of all integration tests: https://github.com/wpessers/opentelemetry-lambda/actions/runs/26711829387

RaphaelManke (approved):
Tested it locally, LGTM

PR #2324, opened 2026-05-31
fix(nodejs): normalize exporter env values before parsing

immanuwell:
`OTEL_*_EXPORTER=none` should still work if the value shows up as `NoNe` or with extra spaces. In Lambda thats pretty easy to hit, env vars are just literal strings.

Repro:
1. On current `main`, set `OTEL_TRACES_EXPORTER=' NoNe '`.
2. Call `wrap()` in the NodeJS layer.
3. The tracer provider still registers, and it logs `Invalid exporter "none"`.

This normalizes exporter env values before parsing, so `none` is handled case insensitively for traces, metrics, and logs. Added regression tests too.

Checks:
`npm test --workspace @opentelemetry-lambda/sdk-layer`

### DIFF
--- a/nodejs/packages/layer/src/wrapper.ts
+++ b/nodejs/packages/layer/src/wrapper.ts
@@ -108,6 +108,18 @@ function getActiveInstrumentations(): Set<string> {
   return instrumentationSet;
 }
 
+function getNormalizedEnvList(name: string): string[] {
+  const value = process.env[name];
+  if (value == null || value.trim() === '') {
+    return [];
+  }
+
+  return value
+    .split(',')
+    .map(item => item.toLowerCase().trim())
+    .filter(item => item !== '');
+}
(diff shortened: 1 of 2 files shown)


wpessers:
@immanuwell the spec doesn't seem to mention anything about case insensitivity: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#exporter-selection specifying only all-lowercase values. The js sdk also doesn't do any such normalisation of the env var values. Additionally it just doesn't make much sense to me why we'd need this, the contract is simple: you provide your exporter(s) of choice in lowercase. That's how it is supposed to work and how it is documented in the spec. Same goes for example for OTEL_PROPAGATORS. Therefore I will close this issue, the code is spec-compliant and in line with js sdk behaviour as-is.

PR #2336, opened 2026-06-05
ci: Add renovate setup

thompson-tomo:
This defines as a renovate setup as per dependabot with updates simply grouped based on language & if they are otel or not.

See config document at https://docs.renovatebot.com/configuration-options/ and after merge an issue needs to be raised in the otel community repo for renovate to be activated.


wpessers:
Ok well that was fast, give me some time to get to this. Trying to get some integration testing stuff wrapped up, then aiming to release this weekend. After that I can look into this PR. Thanks for the effort!

thompson-tomo:
No worries at all. 🙂 for reference An example of a more complex renovate config I have written would be  https://github.com/open-telemetry/opentelemetry-ruby/blob/main/.github/renovate.json5

.github/renovate.json5
```diff
@@ -0,0 +1,87 @@
+{
+  $schema: "https://docs.renovatebot.com/renovate-schema.json",
+  extends: [
+    "config:best-practices",
```
wpessers, line 4: I see this extends [`config:recommended`](https://docs.renovatebot.com/presets-config/#configrecommended), which in turn extends [`:ignoreModulesAndTests`](https://docs.renovatebot.com/presets-default/#ignoremodulesandtests). Looking at the latter's `ignorePaths` I believe we would no longer get updates for python's https://github.com/open-telemetry/opentelemetry-lambda/blob/main/python/src/otel/tests/requirements.txt

We do want this to be included though.

.github/renovate.json5
```diff
@@ -0,0 +1,87 @@
...
+    {
+      groupName: "ruby-other",
+      labels: ["ruby"],
+      matchCategories: ["ruby"],
+      schedule: ["before 8am on Monday"],
+    },
+    {
+      groupName: "actions",
+      labels: ["github_actions"],
+      matchManager: ["github-actions"],
```
wpessers: typo? https://docs.renovatebot.com/configuration-options/#packagerulesmatchmanagers

wpessers (changes requested):
Also I didn't originally include dependabot coverage when I recently added the integration tests (`integration-tests/`). I suppose its `package.json` and `package-lock.json` would be auto-discovered by renovate with the current setup?

.github/renovate.json5
thompson-tomo: You are correct typo. Have fixed it.

wpessers:
Also @thompson-tomo, I might have to go research this myself but just wondering if you know. Currently we get vulnerabilities reported by dependabot whenever we have some vulnerable dependency version, can we keep that when switching to renovate? Or can renovate take care of it instead?

.github/renovate.json5
thompson-tomo: Have added an explicit rule to include the tests directory.

Note I would suggest getting an initial file merged in & then in a subsequent pr do any adjustments/extensions which are needed based on reviewing the Dashboard. I am more than happy to make those adjustments.

thompson-tomo:
Correct it would be automatically discovered. If desired we could setup integration tests to be in a dedicated group.

.github/renovate.json5
wpessers: Sounds good!

thompson-tomo:
> Currently we get vulnerabilities reported by dependabot whenever we have some vulnerable dependency version, can we keep that when switching to renovate? Or can renovate take care of it instead?

Dependabot Vulnerability updates will work alongside renovate without issue. It's what we do in ruby repo's etc. The equivalent renovate functionality is in beta hence not currently used.

wpessers:
Thanks for the clarification!

wpessers:
And thanks for the contribution @thompson-tomo, LGTM! 🚀

wpessers (approved):

PR #2341, opened 2026-06-07
feat(collector): decouple extension from internal log level

### LINKED ISSUE #2101: Enhanced logging for opentelemetry extension startup
There is no existing log statement that marks when the AWS opentelemetry lambda extension has finished starting up and the time taken to start.

Currently if you consume the full 'info' level firehose of log statements from the extension it looks something like:

Image

There is an initial log statement:
`{"level":"info","ts":1767138850.7595742,"msg":"Launching OpenTelemetry Lambda extension","version":"v0.138.0"}`

Then later on there is a log statement that looks like it's saying the OpenTelemetry Lambda extension is done starting but we continue to see log statements arriving for various componenets for several seconds after this:
`{"level":"info","ts":1767138851.1722095,"caller":"service@v0.138.0/service.go:245","msg":"Everything is ready. Begin running and processing data.","resource":{"service.instance.id":"2a52aa2e-513b-4392-9876-2a147d3d1ed6","service.name":"otelcol-lambda","service.version":"v0.138.0"}}`

Ideally I want to drop all the info level noise (and just listen for WARN/ERROR from the extension) apart from two info log statements recording the extension starting and finished starting (with duration), something like:

*  `"msg": "OpenTelemetry Lambda AWS Extension Starting..."`
*  `"msg": "OpenTelemetry Lambda AWS Extension Started.", "initDurationMs": 3292`

This way i can accurately measure if a change in my collector configuration (new receivers/exporters/processors) and/or an upgrade in the lambda layers has made a material change to the opentelemetry lambda aws extension startup time.

**Additional context**
Whats mildly confusing here is that both AWS Lambda and the Otel Collector have the concept of extensions. You see the first log stmt which relates to the AWS Lambda extension starting and then you see collector related log statement for collector extensions starting. If you can think of a better way to make it clearer that the start and finish statements related to the aws extension starting and not otel collector extensions starting that would be great.

wpessers:
Follow-up to #2243 - this time removing the "heuristic" fallback that was included there. Fully decoupling the log levels, and some refactoring to keep the logger creation in one place. Since we want to keep output format (JSON encoding, emitted to STDOUT) aligned and avoid having to update that in 2 places.

### DIFF
--- a/collector/internal/logging/logger.go
+++ b/collector/internal/logging/logger.go
@@ -1,0 +1,59 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logging
+
+import (
+	"os"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+const extensionLogLevelEnvVar = "OPENTELEMETRY_EXTENSION_LOG_LEVEL"
+
+type CoreFunc func(zapcore.LevelEnabler) zapcore.Core
+
+var (
+	encoderConfig = zap.NewProductionEncoderConfig()
+	stdoutSyncer  = zapcore.Lock(zapcore.AddSync(os.Stdout))
+)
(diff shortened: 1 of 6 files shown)


collector/internal/logging/logger.go
```diff
@@ -0,0 +1,57 @@
...
+	stdoutSyncer  = zapcore.Lock(zapcore.AddSync(os.Stdout))
+)
+
+func NewLogger() *zap.Logger {
+	lvl, err := parseLevel(os.Getenv(extensionLogLevelEnvVar))
+
+	l := zap.New(NewCore(lvl))
+
+	if err != nil {
+		l.Warn("unable to parse log level from environment", zap.Error(err))
```
herin049: nit:
```suggestion
		l.Warn("unable to parse log level from environment, falling back to default log level", zap.Error(err), zap.Stringer("default_level", lvl))
```

herin049 (approved):

PR #2347, opened 2026-06-08
build(deps): bump otel nugets

thompson-tomo:
This bumps the otel dependencies in csproj file which removes the deprecated dependencies and tackles the issues flagged in renovate dashboard #2345.


wpessers:
@thompson-tomo I'm fine with doing this, but there's a CI failure. This is only a sample and there's practically no community demand for `.NET` so I'd rather not spend too much time on this myself.

thompson-tomo:
@wpessers no worries, I am just keen to get the Dashboard clean. Looking at the logs of the previous checks it should be a straight forward fix.

Note currently the CI doesn't fail when the build fails hence I thought the failure was ok.

wpessers:
Ok, cool. Good stuff! 🚀

wpessers (approved):

PR #2348, opened 2026-06-08
build(deps): standardise terraform dependencies

thompson-tomo:
This standardises/pin the terraform dependencies which was already done in one case.


wpessers (approved):
@thompson-tomo I'm fine with this, but I do have some doubts letting renovate manage these versions also (as I suspect that's why you're making the change). These are just sample apps. But I suppose if it were just 1 weekly renovate PR that updates all these tf module versions at once, it's fine.

Issue #2340, opened 2026-06-07
ruby: layer doesn't work on arm arch


wpessers:
**Is your feature request related to a problem? Please describe.**
When using the ruby layer on arm lambda function, the lambda function will fail during the init phase. This is due to the build script not passing a `--platform` flag to the docker commands, and our release workflow running on `x86_64`. So every gem that is installed during this build process will be compiled for the `x86_64` arch.

For example the `opentelemetry-exporter-otlp` gem that is required, depends on `google-protobuf`. `google-protobuf` has some platform-specific precompiled binaries. The same is true for multiple other gems that are required transitively through `opentelemetry-instrumentation-all`.

**Describe the solution you'd like**
Build and release the layer for both x86_64 and arm architectures. Similarly to how it is already being done for the collector layer. Re-visit the integration tests we have since #2322 as currently the ruby integration tests will always fail for the `arm` arch.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

DCchoudhury15:
@wpessers can i work on this issue?

wpessers:
@DCchoudhury15 for sure, assigning it to you!

PR #2349, opened 2026-06-08
feat(ruby): build and release layer for both amd64 and arm64 architec…

DCchoudhury15:
### Problem

The Ruby Lambda layer was only built for **x86_64**. When used with **ARM-based Lambda functions**, initialization failed because native gems such as `google-protobuf` were compiled for the wrong architecture.

### Changes

- Added architecture-aware Docker builds in `ruby/src/build.sh` and generated architecture-specific ZIP artifacts.
- Added an `architecture` input to the Ruby layer build action and enabled cross-platform builds using QEMU and Buildx.
- Forwarded the architecture parameter through the layer build workflow.
- Updated release workflows to build and publish layers for both x86_64 and ARM64.
- Updated integration tests to use architecture-specific artifacts and exclude unsupported matrix combinations.

### Result

Ruby Lambda layers are now built correctly for both **x86_64** and **ARM64**, preventing architecture-related runtime failures.

**Fixes:** #2340


DCchoudhury15:
@wpessers pr up for review whenever you get time

wpessers (changes requested):
Hi @DCchoudhury15, first of all appreciate the effort and time that went into the PR. However, I'd like to point out that this dual-arch build logic is something we already have for the collector layer. Maybe it would be wise to check out how the build and release workflow for that one work.

Also I'm not entirely convinced about the QEMU stuff, you're going to be compiling the ruby interpreter from source and that will take a **long** time. I'm wondering, why can't we just use an `arm64` gh action runner instead?

Finally and most importantly, it looks like you disabled integration testing on arm64 for all other language layers. That's not something we want to do, could you explain the thought process here?

DCchoudhury15:
Hi @wpessers,

Thank you for taking the time to review this and for the helpful feedback.

Regarding the QEMU setup, I wasn't entirely sure whether native `arm64` runners were already available and commonly used across the repository, so I defaulted to emulation as the safer option. After looking into your suggestion, I completely agree that using QEMU just to build Ruby native extensions adds unnecessary overhead and slows things down considerably. I've removed the QEMU steps and updated the workflow to follow the same native runner approach used by the collector layer.

For the integration tests, I got a bit too focused on getting the Ruby artifact naming changes working on my branch and missed the broader impact of the matrix exclusions. You're absolutely right that those changes would have disabled ARM test coverage for the other languages as well, which was not my intention. I've removed the exclusions so the existing test coverage remains unchanged.

I'll push the updated commit shortly. When you have a chance, I'd appreciate it if you could take another look and let me know whether the new runner setup is more in line with what you had in mind.

Thanks again for the your time and patience!

.github/workflows/integration-test.yml
```diff
@@ -66,20 +66,34 @@ jobs:
...
+        architecture: [amd64, arm64]
+        exclude:
+          - language: nodejs
+            architecture: arm64
+          - language: python
+            architecture: arm64
+          - language: javaagent
+            architecture: arm64
+          - language: javawrapper
+            architecture: arm64
```
thompson-tomo, line 83: This should be able to be simplified to

```suggestion
        runner: ubuntu-latest
        include:
          - language: ruby
            architecture: amd64
            runner: ubuntu-latest
          - language: ruby
            architecture: arm64
            runner: ubuntu-24.04-arm
```

.github/workflows/integration-test.yml
```diff
@@ -66,20 +66,34 @@ jobs:
   build-layer:
     needs: prepare-languages
     if: ${{ !inputs.use-existing-layer-artifact }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.architecture == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
```
thompson-tomo, line 69: alongside other suggestion

```suggestion
    runs-on: ${{ matrix.runner }}
```

.github/workflows/integration-test.yml
```diff
@@ -66,20 +66,32 @@ jobs:
   build-layer:
     needs: prepare-languages
     if: ${{ !inputs.use-existing-layer-artifact }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
     strategy:
       fail-fast: false
       matrix:
         language: ${{ fromJson(needs.prepare-languages.outputs.languages) }}
+        runner: ubuntu-latest
```
thompson-tomo:
```suggestion
        runner: ["ubuntu-latest"]
```

.github/workflows/integration-test.yml
```diff
@@ -66,20 +66,32 @@ jobs:
...
       matrix:
         language: ${{ fromJson(needs.prepare-languages.outputs.languages) }}
+        runner: ["ubuntu-latest"]
+        include:
+          - language: ruby
+            architecture: amd64
+            runner: ubuntu-latest
+          - language: ruby
+            architecture: arm64
+            runner: ubuntu-24.04-arm
```
wpessers: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#expanding-or-adding-matrix-configurations

Not sure how this will work when triggered for a single (non-ruby) language like `python`. I'm guessing this has the unintended side-effect of running the ruby integration test each time a non-ruby integration test is being ran.

.github/actions/build-ruby-layer/action.yml
```diff
@@ -1,28 +1,43 @@
...
+        unzip -q build/opentelemetry-ruby-layer-amd64.zip
         export GEM_PATH=$PWD/ruby/gems/3.3.0/
         SDK_VERSION=$(ruby -e 'require "opentelemetry-sdk"; puts OpenTelemetry::SDK::VERSION')
         echo "component-version=$SDK_VERSION" >> "$GITHUB_OUTPUT"
+
+    - name: Skip version output for non-amd64
+      if: ${{ inputs.architecture != 'amd64' }}
+      id: version-skip
+      shell: bash
+      run: echo "component-version=" >> "$GITHUB_OUTPUT"
```
wpessers: I find this quite confusing, you seem to have done a lot of work here and in the release workflow to skip outputting the version. IIRC the collector release has the same matrixed amd64/arm64 build and handles this with no such extra logic. As I said earlier, might be worth taking a look how the collector builds / releases are setup. Let me reiterate that you're basically building the same mechanism that is already in place there.

.github/workflows/release-layer-ruby.yml
```diff
@@ -11,21 +11,53 @@ permissions:
...
+      component-version: ${{ steps.read-version.outputs.version }}
+    steps:
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: ruby-component-version
+
+      - id: read-version
+        shell: bash
+        run: echo "version=$(cat component-version.txt)" >> "$GITHUB_OUTPUT"
+
```
wpessers: Same comment here, can we please take a look at the collector release workflow? I'm not convinced that we need all of this stuff.

wpessers (changes requested):

.github/workflows/release-layer-ruby.yml
DCchoudhury15: Hi @wpessers,  Thanks for your patience and for pointing me back to the collector pattern , that was definitely the right approach.

For `integration-test.yml`, I originally tried using `runner + include` to assign the native arm64 runner only to Ruby builds. I overlooked the fact that `include` adds extra matrix combinations rather than modifying existing ones, which would have caused the Ruby tests to run with every non-Ruby language trigger as well. I've switched to using `exclude` instead, which is much cleaner and avoids that issue entirely.

For the `version-skip` step and `get-version` job in the release workflow, I was trying to prevent a matrix race condition where the arm64 shard finishing last might overwrite the `component-version` output with an empty value. After taking another look at the collector workflow, I realized it already handles this cleanly with an `if: architecture == 'amd64'` guard on the version step and a direct outputs reference. I was overthinking it, so I've reverted the implementation to match the collector pattern.Again than you for your patience and guidance.

DCchoudhury15: @wpessers changes up for review

tylerbenson:
Maybe worth asking @xuan-cao-swi to take a look as well.

xuan-cao-swi:
If we want to support `arm64` for Ruby—and given that Ruby is currently the only language layer requiring multi-architecture builds—I’d suggest avoiding changes to shared workflows like `.github/actions/build-layer/action.yml` and `integration-test.yml`.

Instead, we can scope the changes to `release-layer-ruby.yml` and `build-ruby-layer/action.yml` by adding separate jobs in `release-layer-ruby.yml` (e.g., `build-amd-layer` and `build-arm-layer`).

For `integration-test.yml`, to keep changes to a minimum, we could add an `architecture` input and set `runs-on: ${{ inputs.architecture == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}`, rather than expanding the test matrix with `architecture: [amd64, arm64]`. This avoids unnecessary churn on shared workflows used by all languages.

wpessers:
Fully agree with @xuan-cao-swi here. Thanks for sharing your view! This was roughly hoe I envisioned it myself. Admittedly I'm very thinly spread at the moment adn deprioritized this particular PR.

PR #2364, opened 2026-06-13
chore: Make renovate behave more like dependabot

thompson-tomo:
This has been tested to resolve the following issues:
- incorrect grouping of otel dependencies - resolved by changing order & match pattern
- Not updating tests directory - adjustment to the ignore paths

Partially test to resolve
- incorrect pr titles - review of logs showed title changing due to adding scope to label rule but not all pr combinations tested

this also seperates major updates from other groups & adds approval to avoid accidentally merging major bump as well as enable to delay them till after minor etc. We can certainly change this rule.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -116,7 +116,7 @@
     },
     {
       groupName: 'opentelemetry-deps-dotnet',
-      updateTypes: [
+      matchUpdateTypes: [
         '!major'
       ],
       matchDepNames: [
@@ -131,7 +131,7 @@
     },
     {
       groupName: 'java-other',
-      updateTypes: [
+      matchUpdateTypes: [
         '!major'
       ],
       matchCategories: [
@@ -143,7 +143,7 @@
     },
     {
       groupName: 'opentelemetry-deps-java',
-      updateTypes: [
+      matchUpdateTypes: [
         '!major'
       ],
       matchDepNames: [
@@ -158,7 +158,7 @@
     },
     {
       groupName: 'nodejs-other',
-      updateTypes: [
+      matchUpdateTypes: [
         '!major'
       ],
       matchCategories: [
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -76,7 +76,7 @@
     },
     {
       groupName: 'collector-other',
-      updateTypes: [
+      matchUpdateTypes: [
         '!major'
       ],
       matchCategories: [
@@ -88,7 +88,7 @@
     },
     {
       groupName: 'opentelemetry-deps-collector',
-      updateTypes: [
+      matchUpdateTypes: [
         '!major'
       ],
       matchPackageNames: [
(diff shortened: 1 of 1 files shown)


tylerbenson (approved):

.github/renovate.json5
```diff
@@ -5,22 +5,79 @@
...
+        'github_actions',
+      ],
+      semanticCommitType: 'build',
+      matchManagers: [
+        'github-actions',
+      ],
+    },
+    {
+      groupName: 'collector-other',
+      updateTypes: [
```
wpessers: When I look for this config, I only find `matchUpdateTypes` here: https://docs.renovatebot.com/configuration-options/#packagerulesmatchupdatetypes

The intended change here is that we will have to manually approve renovate to open the pr for major version bumps, correct? That's not how it is today with Dependabot, but I'm down to try that. Will have to see how it goes for some weeks and then we can decide whether that config works for us or not!

.github/renovate.json5
```diff
@@ -57,24 +130,24 @@
...
       schedule: [
-        'before 8am on Monday'
+        'before 8am on Monday',
       ],
     },
     {
       groupName: 'opentelemetry-deps-java',
-      labels: [
-        'java',
+      updateTypes: [
```
wpessers:
```suggestion
      matchUpdateTypes: [
```

.github/renovate.json5
```diff
@@ -84,24 +157,24 @@
       ],
     },
     {
-      groupName: 'java-other',
-      labels: [
-        'java',
+      groupName: 'nodejs-other',
+      updateTypes: [
```
wpessers:
```suggestion
      matchUpdateTypes: [
```

.github/renovate.json5
```diff
@@ -42,12 +103,24 @@
...
+      matchCategories: [
         'dotnet'
       ],
+      schedule: [
+        'before 8am on Monday'
+      ],
+    },
+    {
+      groupName: 'opentelemetry-deps-dotnet',
+      updateTypes: [
```
wpessers:
```suggestion
      matchUpdateTypes: [
```

.github/renovate.json5
```diff
@@ -57,24 +130,24 @@
       ],
     },
     {
-      groupName: 'dotnet-other',
-      labels: [
-        'dotnet'
+      groupName: 'java-other',
+      updateTypes: [
```
wpessers:
```suggestion
      matchUpdateTypes: [
```

wpessers (commented):
@thompson-tomo found some remaining ones, after this LGTM!

.github/renovate.json5
thompson-tomo: Thanks for spotting that, been fixed.

Correct, let's see how approvals for major go given there shouldn't be many.

thompson-tomo:
Thanks @wpessers those suggestions have been committed.

PR #2387, opened 2026-06-19
chore: group docker updates

thompson-tomo:
This groups docker updates and sets the expected commit type.

note the docker dependency has also been pinned.


wpessers (approved):

thompson-tomo:
This is to handle pr such as #2386

PR #2392, opened 2026-06-21
ci: use sdk version files

thompson-tomo:
This changes the ci tasks to source versions from sdk files.

This decouples the sdk updates from the github action updates & enables a consistent version to be used locally.

Note renovate will still detect the dependencies & offer updates to them.


wpessers (commented):
@thompson-tomo I think your idea here is good. Just thinking we probably want to move to something like mise in the future to make DX better than it is.For now I think we can go through with this though.

python/src/.python-version
```diff
@@ -0,0 +1 @@
+3.11
```
wpessers: This was upgraded from 3.10 to 3.11 with the move from inline python version to this separate file. For now I'd like to keep it at 3.10, we can let renovate do the actual upgrading eventually.
```suggestion
3.10
```

wpessers (changes requested):

wpessers:
@thompson-tomo one nit, other than that LGTM.

thompson-tomo:
No worries @wpessers have committed that suggestion.

wpessers (approved):

PR #2401, opened 2026-06-22
chore: Add gomodTidy to postUpdateOptions in renovate config

thompson-tomo:
We are currently having an issue where gomodtidy is not running on pr's which is causing pr's aka https://github.com/open-telemetry/opentelemetry-lambda/pull/2394 to fail ci checks. The hope is that will address that.

Note collector also a custom ci step to add additional commit but hope is to avoid That.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -95,6 +95,9 @@
       matchCategories: [
         'golang',
       ],
+      postUpdateOptions: [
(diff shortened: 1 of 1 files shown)


wpessers (approved):

wpessers:
Thanks for the effort @thompson-tomo 🚀

PR #2409, opened 2026-06-22
feat(collector): support inline Base64 config via `OPENTELEMETRY_COLLECTOR_CONFIG_CONTENT`

fatmcgav:
Adds a new `OPENTELEMETRY_COLLECTOR_CONFIG_CONTENT` environment variable that
accepts a standard Base64-encoded YAML collector config. The decoded content is
fed into the existing `yamlprovider` (`yaml:` URI scheme), so no new provider is
needed.

Resolution order:
1. `OPENTELEMETRY_COLLECTOR_CONFIG_URI` (existing, unchanged — warns if `_CONTENT` is also set)
2. `OPENTELEMETRY_COLLECTOR_CONFIG_CONTENT` (new — Base64-decoded YAML)
3. `OPENTELEMETRY_COLLECTOR_CONFIG_FILE` (deprecated fallback, unchanged)
4. `/opt/collector-config/config.yaml` (default)

If Base64 decoding fails an error is logged and the extension falls through to
the next option. `NewCollector`'s signature is unchanged.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/collector/README.md
+++ b/collector/README.md
@@ -176,14 +176,40 @@ from an S3 object using a CloudFormation template:
 
 Loading configuration from S3 will require that the IAM role attached to your function includes read access to the relevant bucket.
 
+### Inline configuration via Base64
+
+You can supply the collector configuration inline as a Base64-encoded string using `OPENTELEMETRY_COLLECTOR_CONFIG_CONTENT`. This is useful when you want to avoid deploying a config file or referencing an external URI:
+
+```
+aws lambda update-function-configuration --function-name Function \
+  --environment Variables={OPENTELEMETRY_COLLECTOR_CONFIG_CONTENT=$(base64 < collector.yaml)}
+```
+
+Or in a CloudFormation template:
+
+```yaml
+  Function:
+    Type: AWS::Serverless::Function
+    Properties:
+      ...
+      Environment:
+        Variables:
+          OPENTELEMETRY_COLLECTOR_CONFIG_CONTENT: <output of `base64 < collector.yaml`>
+```
+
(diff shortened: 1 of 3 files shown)


tylerbenson:
I don't think we can accept this here as it really should be something standardized on the collector side first.

fatmcgav:
> I don't think we can accept this here as it really should be something standardized on the collector side first.

Isn't this a lambda specific use case though? I can't personally see a need to use a base64 encoded string on any of the other common use cases..

wpessers (changes requested):
I'm with @tylerbenson on this. And on top of that, I believe it's already a thing that **has** been standardised in the collector upstream (see: https://github.com/open-telemetry/opentelemetry-collector/issues/2910). We even already ship it in the collector extension:
https://github.com/open-telemetry/opentelemetry-lambda/blob/e285846f398b9b5d7b88f63402f1cec89c19ce95/collector/internal/collector/collector.go#L79

So you should just be able to use the env provider. If there's no further questions, I'll thus be closing this PR.

fatmcgav:
Interesting.. I didn't find any docs related to that when I was looking... Will take another look next week though.. 👍

wpessers:
@fatmcgav I'm willing to accept a PR to add some documentation for it here if that would make sense to you.

Since the extension layer basically ships a slimmed down version of the otel collector it's hard to decide what is worth documenting.

For now, I'll close the PR. If for any reason you can't get it to work with the env provider I'd advise opening an issue.

PR #2411, opened 2026-06-23
feat(nodejs): add prisma instrumentation 

rndD:
Added Prisma as an additional enableable instrumentation for the NodeJS layer.


rndD:
Hi. Any chance we can merge this?

wpessers:
It seems fine, before such a change though we usually take a look at impact on the layer artifact size when including an entirely new package. Could you please deliver that?

serkan-ozal:
@rndD Doesn't OTEL JS SDK have its own Prisma instrumentation so you add 3rd party dependency for it?

PR #2412, opened 2026-06-23
chore: Bump unstable

thompson-tomo:
This allows bumps of the specified unstable packages to the next unstable regardless of new version (major, minor or patch).

Package rule has been copied from otel-android where it is working.


wpessers (approved):

wpessers:
Nice fix, again thanks for putting in the effort!

PR #2413, opened 2026-06-27
feat(collector): log extension startup completion and duration

### LINKED ISSUE #2101: Enhanced logging for opentelemetry extension startup

mvanhorn:
## Summary

The collector extension emits a `"Launching OpenTelemetry Lambda extension"` info log at process start, but nothing marks that startup has finished or how long it took. This adds a single info-level `"OpenTelemetry Lambda extension startup complete"` log carrying a `startup_duration` field, emitted once the collector has started successfully.

A start timestamp is captured in `main.go` immediately before the launch log and threaded into `lifecycle.NewManager` (the only production caller, so the signature change is contained). After `collector.Start` returns successfully in `manager.Run`, the completion log is emitted with `zap.Duration("startup_duration", time.Since(startTime))`. The duration measures from the extension's launch log through successful collector start.

## Why this matters

Operators consuming the info-level firehose currently cannot tell when the extension is actually ready, nor measure its contribution to cold start. In #2101, @wpessers confirmed this is a valid concern, split the original report into two sub-features, and narrowed the remaining ask to exactly this startup-duration logging. This keeps it a single info-level statement so it is visible in the default firehose, with no new log-level knob (the log-decoupling sub-feature is out of scope and already addressed separately).

## Testing

Added `TestRunLogsStartupDuration` in `manager_test.go` using a zap observer core:
- Happy path: a successful `MockCollector` start emits exactly one startup-complete log with a non-negative `startup_duration` field.
- Edge case: a zero-value start time still produces a valid (non-negative) duration and does not panic.
- Error path: when `collector.Start` returns an error, the startup-complete log is NOT emitted and `Run` returns the error, preserving current failure behavior.

`go build ./...`, `go vet`, and `go test ./internal/lifecycle/...` pass in the `collector` module.

Fixes #2101

### DIFF
--- a/collector/internal/lifecycle/manager.go
+++ b/collector/internal/lifecycle/manager.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/open-telemetry/opentelemetry-lambda/collector/lambdalifecycle"
 
@@ -53,9 +54,10 @@ type manager struct {
 	wg                 sync.WaitGroup
 	lifecycleListeners []lambdalifecycle.Listener
 	initType           lambdalifecycle.InitType
+	startTime          time.Time
 }
 
-func NewManager(ctx context.Context, logger *zap.Logger, version string) (context.Context, *manager) {
+func NewManager(ctx context.Context, logger *zap.Logger, version string, startTime time.Time) (context.Context, *manager) {
 	ctx, cancel := context.WithCancel(ctx)
 
 	sigs := make(chan os.Signal, 1)
@@ -102,6 +104,7 @@ func NewManager(ctx context.Context, logger *zap.Logger, version string) (contex
 		extensionClient: extensionClient,
 		listener:        listener,
 		initType:        initType,
+		startTime:       startTime,
 	}
 
 	factories, _ := lambdacomponents.Components(res.ExtensionID)
@@ -118,6 +121,8 @@ func (lm *manager) Run(ctx context.Context) error {
 		}
 		return err
 	}
+
+	lm.logger.Info("OpenTelemetry Lambda extension startup complete", zap.Duration("startup_duration", time.Since(lm.startTime)))
 
 	lm.wg.Add(1)
 	go func() {
--- a/collector/internal/lifecycle/manager_test.go
+++ b/collector/internal/lifecycle/manager_test.go
@@ -169,11 +169,12 @@ func TestRun(t *testing.T) {
 		extensionClient: extensionapi.NewClient(logger, u.Host, extensionEventTypes),
 	}
 	lm.wg.Add(1)
+	runErr := make(chan error, 1)
 	go func() {
-		require.NoError(t, lm.Run(ctx))
+		runErr <- lm.Run(ctx)
 	}()
 	lm.wg.Done()
-
+	assert.NoError(t, <-runErr)
 }
 
 func TestProcessEvents(t *testing.T) {
--- a/collector/internal/lifecycle/manager_test.go
+++ b/collector/internal/lifecycle/manager_test.go
@@ -24,14 +24,100 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
+	"go.uber.org/zap/zaptest/observer"
 
 	"github.com/open-telemetry/opentelemetry-lambda/collector/internal/extensionapi"
 	"github.com/open-telemetry/opentelemetry-lambda/collector/internal/telemetryapi"
 )
+
+const startupCompleteMsg = "OpenTelemetry Lambda extension startup complete"
+
+func TestRunLogsStartupDuration(t *testing.T) {
+	shutdownServer := func(t *testing.T) *httptest.Server {
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(200)
+			_, err := w.Write([]byte(`{"time":"2006-01-02T15:04:05.000Z", "eventType":"SHUTDOWN", "record":{}}`))
+			require.NoError(t, err)
+			_, err = io.ReadAll(r.Body)
+			require.NoError(t, err, "failed to read request body: %v", err)
+		}))
+		t.Cleanup(server.Close)
+		return server
+	}
+	extensionEventTypes := []extensionapi.EventType{extensionapi.Invoke, extensionapi.Shutdown}
+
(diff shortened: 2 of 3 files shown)


wpessers:
Thanks @mvanhorn for working on this! I'm a bit swamped at the moment but will take a look as soon as possible!

collector/internal/lifecycle/manager_test.go
wpessers, line 113: I know this is an existing test you didn't touch but since your addition of the extension startup log line, it has become flaky. I've noticed it panicking several times when I was running the tests locally. Looks like it's because the last case in here runs `lm.Run(ctx)` inside a goroutine that is never joined, using the logger which is `zaptest.NewLogger(t)`, and thus it panics when anything logs after the test returns.

I think the goroutine needs to be joined before this test returns, and maybe switch `require.NoError` to `assert.NoError` for safety also.

collector/internal/lifecycle/manager.go
```diff
@@ -119,6 +122,8 @@ func (lm *manager) Run(ctx context.Context) error {
 		return err
 	}
 
+	lm.logger.Info("OpenTelemetry Lambda extension startup complete", zap.Duration("startup_duration", time.Since(lm.startTime)))
```
wpessers, line 125: The log line in itself is correct, but an existing test has become flaky due to its addition. See comment in the testfile.

wpessers (changes requested):

collector/internal/lifecycle/manager_test.go
```diff
@@ -24,15 +24,101 @@ import (
...
+		require.NotPanics(t, func() {
+			require.NoError(t, lm.Run(context.Background()))
+		})
+
+		entries := logs.FilterMessage(startupCompleteMsg).All()
+		require.Len(t, entries, 1)
+		duration, ok := entries[0].ContextMap()["startup_duration"].(time.Duration)
+		require.True(t, ok)
+		assert.GreaterOrEqual(t, duration, time.Duration(0))
+	})
```
wpessers: This seems not so useful. Leaving startTime as the zero value means `time.Time{}` or in other words `01/01/0001 00:00:00`. If you use that to calculate the `time.Since()` on inside `manager.go` you will exceed the max value that can be represented and thus clamp to said max value. So, this literally **can't** panic and duration will always be greater than 0. Not sure what is actually meant to be tested here?

mvanhorn:
Good catch on the flakiness - pushed be5b17d, which synchronizes that assertion so it no longer races the startup log goroutine. go test -count=5 on the lifecycle package stays green locally.

wpessers (changes requested):
@mvanhorn I forgot to submit this follow-up review. Can you check please?

collector/internal/lifecycle/manager_test.go
```diff
@@ -24,15 +24,101 @@ import (
...
+		}
+		require.NoError(t, lm.Run(context.Background()))
+
+		entries := logs.FilterMessage(startupCompleteMsg).All()
+		require.Len(t, entries, 1, "expected exactly one startup-complete log")
+		field, ok := entries[0].ContextMap()["startup_duration"]
+		require.True(t, ok, "startup-complete log must carry a startup_duration field")
+		duration, ok := field.(time.Duration)
+		require.True(t, ok, "startup_duration must be a duration field")
+		assert.GreaterOrEqual(t, duration, time.Duration(0), "startup_duration must be non-negative")
```
wpessers: I don't think this can ever fail. What was the intention behind this test?

wpessers (commented):
One final comment.

mvanhorn:
All three in b061e91.

The flake was the waitgroup case racing itself: `wg.Done()` fired right after launching the goroutine, with nothing ordering it against `lm.Run`'s own use of the same waitgroup. The extension server in that case now blocks until the request is actually in flight and only then releases, so `Done()` lands at a deterministic point. `go test -race -count=100` on it is clean here, though you saw it panic locally and I would rather you confirm on your machine than take my run for it.

You were right about the other two. `time.Since(time.Time{})` clamps, so that subtest was asserting on a saturated value, and the non-negative duration check could not fail. Both gone.

wpessers (approved):

tylerbenson (approved):

mvanhorn:
Thanks, both diagnoses were right and the second one taught me something.

On the flake: the cause is that every httptest handler in this file calls `require.*`, which testify implements with `runtime.Goexit` and only supports on the test goroutine. A handler still running as the test returns then panics through zaptest's logger. The startup log line didn't introduce that, it just widened the window enough to hit it. Switched the in-handler calls to `assert.*` as you suggested. 30 consecutive `TestRun` runs are clean.

On the assertion: you're right it couldn't fail. It only checked the field's type, which `zap.Duration` guarantees. It now bounds the value, and the upper bound is the part that matters, since `time.Since` on an unset `startTime` clamps to the max duration rather than erroring. I verified it by injecting `startTime: time.Time{}` and confirming the new bound fails, then reverting.

One thing worth flagging: my first attempt asserted the duration was positive, and that failed intermittently. With a `MockCollector` the elapsed time genuinely rounds to 0ns, so it is `>= 0`, not `> 0`. Would have been a new flake if I hadn't run it repeatedly.

wpessers (approved):

mvanhorn:
Thanks @wpessers for reviewing and merging the extension startup logging, that visibility should help debug cold starts.

PR #2428, opened 2026-06-30
build: run goMod in ci

thompson-tomo:
This runs go mod tidy as part of the ci when go.mod/go.sum has changed and the branch is named renovate.

This has been tested in my fork where i merged these changes into main & verified that it ran as expected including resolving the build issue. This can be seen in https://github.com/thompson-tomo/opentelemetry-lambda/pull/109


.github/workflows/renovate-go.yml
```diff
@@ -0,0 +1,52 @@
...
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
+        with:
+          go-version-file: collector/go.mod
+      - run: find . -name "go.mod" -execdir go mod tidy \;
```
wpessers, line 28: Not 100% sure, but isn't this dangerous? When `go mod tidy` fails, the error would be swallowed? Or would we be able to see that failure in a later workflow that runs on the PR? (I think so, just checking to make sure...)

.github/workflows/renovate-go.yml
```diff
@@ -0,0 +1,52 @@
...
+            echo "changed=true" >> $GITHUB_OUTPUT
+          else
+            echo "changed=false" >> $GITHUB_OUTPUT
+          fi
+      - name: Commit and push changes
+        if: steps.check_changes.outputs.changed == 'true'
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add collector/
```
wpessers, line 46: The go sample app is not included in this, if that was intended it's fine by me though.

wpessers (approved):

wpessers:
@thompson-tomo I approved for now, with just 2 minor questions. Related to the first one, can we confirm that after this workflow runs and pushes the `chore: go mod tidy` commit, it actually retriggers CI?

.github/workflows/renovate-go.yml
thompson-tomo: Yes it should be seen in the build workflow like we are seeing atm.

.github/workflows/renovate-go.yml
thompson-tomo: Good spot, i do however think sample-apps are ok as some updates are currently working. I am inclined to leave it as is given it unblocks ci & only add the additional path if we need it & we have a way to test it.

thompson-tomo:
@wpessers Yes, it does trigger. If you look at https://github.com/thompson-tomo/opentelemetry-lambda/pull/109 the pr is now passing. prior to this change that pr had 1 commit & was falling. That branch is a push of the failing branch from this repo.

PR #2485, opened 2026-07-27
ci: Update Dependabot schedule to monthly

thompson-tomo:
Changed update schedule from weekly to monthly for various package ecosystems in dependabot configuration as renovate is now managing them.

This is a progression towards removing the dependabot rules all together.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,16 +2,6 @@
 
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    directories:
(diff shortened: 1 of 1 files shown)


tylerbenson (approved):
Good plan. Thanks!
@wpessers feel free to revert if you feel differently.

thompson-tomo:
No worries, i have left go as i noticed there is 1 dependency file not being updated so i will look at that issue. And go is the most complex with gomod tidy etc.

PR #2487, opened 2026-07-28
ci: Allow major golang requires

thompson-tomo:
This removes approval being required for the go dependency updates.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -333,6 +333,9 @@
       matchUpdateTypes: [
         'major'
       ],
+      matchDepTypes: [
+        '!require',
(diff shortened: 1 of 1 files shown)


tylerbenson:
I'm not sure what this means... could you elaborate a bit please?  Maybe show an example?

thompson-tomo:
When a dependency is detected additional metadata is captured/set which can then be used in rules. One of these pieces of information is deptype. This rule enables major `require` dependencies  to not need approval before being updated. These affects the golang updates which are currently awaiting approval.

tylerbenson (approved):

PR #2491, opened 2026-07-28
ci: Auto label pr's

thompson-tomo:
This moves labeling out of renovate and into a dedicated workflow. This enables all pr's to be auto labeled and further simplifies renovate rules.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,46 @@
+ci:
+  - changed-files:
+      - any-glob-to-any-file:
+          - ".github/**"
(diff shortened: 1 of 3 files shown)


tylerbenson (approved):
good idea.  looks reasonable to me.  prob worth applying the same change to dependabot config while you're at it even though it's going away.

thompson-tomo:
I did think of that but there is no dependabot config to set labels it uses a default behaviour & the labels are additive so no issue if 2 processes are setting the labels.

PR #2516, opened 2026-08-04
docs: Combine Emeritus Maintainers and Approvers sections

thompson-tomo:
Updated the Emeritus section and added a new 'Thanks to all the people who have contributed' section.

Note the consolidation is as has been done in other repos & is as it is expected to be for the central Emeritus workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,18 +141,17 @@ For more information about the maintainer role, see the [community repository](h
 
 For more information about the approver role, see the [community repository](https://github.com/open-telemetry/community/blob/main/guides/contributor/membership.md#approver).
 
-### Emeritus Maintainers
+### Emeritus
 
-- [Alex Boten](https://github.com/codeboten)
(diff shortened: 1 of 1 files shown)


tylerbenson:
Do you have some links to PRs/repos where this has been done already?

thompson-tomo:
Take a look at https://github.com/search?q=org%3Aopen-telemetry+Move+inactive+members+to+emeritus&type=pullrequests it's all the prs that have been raised by the monthly workflow. Key thing is it looks for the emeritus heading to know where to add people if they should be moved to emeritus. If it can't be found it just adds another section at the very end.

tylerbenson:
Ah, so by us separating it, it would break that workflow somewhat?

tylerbenson (approved):

thompson-tomo:
Correct, if it found someone who should be moved to emeritus it would create another emeritus section as it cannot find the heading it needs.

PR #2517, opened 2026-08-04
fix(nodejs): bundle missing winston-transport dependency in Lambda layer

### LINKED ISSUE #2065: Winston Instrumentation Race Condition in AWS Lambda Layer
Winston Instrumentation Race Condition in AWS Lambda Layer

**Describe the bug**

Winston instrumentation fails silently when enabled via `OTEL_NODE_ENABLED_INSTRUMENTATIONS` in AWS Lambda. Logs are not captured or exported to OTLP endpoints. The root cause is a race condition: Winston instrumentation registers synchronously during `init()` before `LoggerProvider` is created asynchronously in `wrap()`, causing `logs.getGlobalLoggerProvider()` to return `undefined`.

**Steps to reproduce**

1. Deploy AWS Lambda function with Node.js 20.x runtime

2. Attach AWS Lambda Layer: `arn:aws:lambda:us-east-1:901920570463:layer:aws-otel-nodejs-arm64-ver-1-30-2:1`

3. Set environment variables:

```bash
AWS_LAMBDA_EXEC_WRAPPER=/opt/otel-handler
OTEL_NODE_ENABLED_INSTRUMENTATIONS=aws-sdk,http,winston
OTEL_EXPORTER_OTLP_ENDPOINT=https://otlp.nr-data.net
OTEL_EXPORTER_OTLP_HEADERS=api-key=<your-license-key>
OTEL_TRACES_EXPORTER=otlp
OTEL_LOGS_EXPORTER=otlp
```

4. Create a Winston logger in your function:
```javascript
const logger = require('winston').createLogger({
  level: 'info',
  transports: [new winston.transports.Console()]
});
```

5. Invoke the Lambda function with log statements

6. Check your OTLP backend (New Relic, DataDog, etc.)

Winston logs should be automatically captured by the OpenTelemetry instrumentation and exported to the configured OTLP endpoint with proper trace context propagation.

**What did you see instead?**

Logs never reach the OTLP backend
No error messages or warnings in Lambda logs

**Root Cause**

Timing issue in nodejs/packages/layer/src/wrapper.ts:

```bash
┌─────────────────────────────────────────────────────┐
│ init() - SYNCHRONOUS                                │
├─────────────────────────────────────────────────────┤
│ createInstrumentations()                            │
│   → includes Winston instrumentation               │
│                                                      │
│ registerInstrumentations({ instrumentations })     │
│   → Winston calls logs.getGlobalLoggerProvider()   │
│   → Returns UNDEFINED ❌                            │
│   → Instrumentation becomes no-op                  │
└─────────────────────────────────────────────────────┘
                           ↓
┌─────────────────────────────────────────────────────┐
│ wrap() - ASYNCHRONOUS (too late)                    │
├─────────────────────────────────────────────────────┤
│ await initializeProvider()                          │
│   → await initializeLoggerProvider()               │
│   → LoggerProvider created and registered ✅       │
│   → But Winston already failed above               │
└─────────────────────────────────────────────────────┘
```

**Workaround**

Disable Winston automatic instrumentation and use manual OpenTelemetryTransportV3:

1.Remove winston from OTEL_NODE_ENABLED_INSTRUMENTATIONS:
```bash
OTEL_NODE_ENABLED_INSTRUMENTATIONS=aws-sdk,http
````

2. Install the transport package:
```bash
npm install @opentelemetry/winston-transport
```

3. Use manual transport in your logger:
```javascript
const winston = require('winston');
const { OpenTelemetryTransportV3 } = require('@opentelemetry/winston-transport');

const logger = winston.createLogger({
  level: 'info',
  format: winston.format.combine(
    winston.format.json(),
  ),
  transports: [
    new winston.transports.Console(),
    new OpenTelemetryTransportV3()  // Logs sent to OTLP ✅
  ]
});
```

This works because the logger is instantiated after wrap() has already created the global LoggerProvider.

AWS Lambda Layer: aws-otel-nodejs-arm64-ver-1-30-2
OpenTelemetry Core: v1.30.0 (bundled in layer)
OpenTelemetry API: @opentelemetry/api@1.9.0
OpenTelemetry Winston Instrumentation: included in layer
Winston: ^3.13.0

Node.js 20.x (ARM64 architecture)

**Additional context**

Environment: AWS Lambda Node.js 20.x, ARM64
OTLP Backend: New Relic (tested with https://otlp.nr-data.net)
Issue Location: opentelemetry-lambda/nodejs/packages/layer/src/wrapper.ts (lines ~100-150)
Related Repository: opentelemetry-js-contrib/packages/instrumentation-winston
Impact: Any Lambda function using Winston logger with automatic instrumentation cannot export logs to observability backends
Suggested Fix: Move Winston instrumentation registration to wrap() function after initializeLoggerProvider() completes

pujitha24:
While looking at the Winston issue, I found that the Lambda layer already includes `@opentelemetry/instrumentation-winston`, but the transport it needs for exporting logs isn't actually bundled.

Webpack was already warning that it couldn't resolve `@opentelemetry/winston-transport`. That require happens when the Winston instrumentation tries to attach its OTel transport, so it fails silently and log export never gets enabled. Trace/span ID injection is a separate path and is not affected.

This adds `@opentelemetry/winston-transport` to the layer dependencies and adds a test using Winston itself to verify the OTel transport gets attached.

I also confirmed the webpack warning disappears and the layer build, tests and lint pass.

AI assistance: I used AI tools while investigating this and preparing the change; I have reviewed and tested it myself.

Fixes #2065

### DIFF
--- a/nodejs/package-lock.json
+++ b/nodejs/package-lock.json
@@ -496,6 +496,15 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@colors/colors": {
+      "version": "1.6.0",
+      "resolved": "https://registry.npmjs.org/@colors/colors/-/colors-1.6.0.tgz",
+      "integrity": "sha512-Ir+AOibqzrIsL6ajt3Rz3LskB7OiMVHqltZmspbW/TJuTVuyOMirVqAkjfY6JISiLHgyNqicAC8AyHHGzNd/dA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=0.1.90"
+      }
+    },
     "node_modules/@cspotcode/source-map-support": {
       "version": "0.8.1",
       "resolved": "https://registry.npmjs.org/@cspotcode/source-map-support/-/source-map-support-0.8.1.tgz",
@@ -507,6 +516,18 @@
       },
       "engines": {
         "node": ">=12"
+      }
+    },
+    "node_modules/@dabh/diagnostics": {
+      "version": "2.0.8",
+      "resolved": "https://registry.npmjs.org/@dabh/diagnostics/-/diagnostics-2.0.8.tgz",
+      "integrity": "sha512-R4MSXTVnuMzGD7bzHdW2ZhhdPC/igELENcq5IjEverBvq5hn1SXCWcsi6eSsdWP0/Ur+SItRRjAktmdoX/8R/Q==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@so-ric/colorspace": "^1.1.6",
+        "enabled": "2.0.x",
+        "kuler": "^2.0.0"
       }
     },
     "node_modules/@discoveryjs/json-ext": {
@@ -1863,6 +1884,20 @@
         "@opentelemetry/api": "^1.1.0"
       }
     },
+    "node_modules/@opentelemetry/winston-transport": {
+      "version": "0.31.0",
+      "resolved": "https://registry.npmjs.org/@opentelemetry/winston-transport/-/winston-transport-0.31.0.tgz",
+      "integrity": "sha512-higyMoGHbDTNlCX5ue1dJjDQuYSWwC0L4k6x57RKLG0uXdCIdyIu/OQYnjciL+7Kv2XmPMy4VQloeh36NTJTuQ==",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@opentelemetry/api-logs": "^0.221.0",
+        "@opentelemetry/semantic-conventions": "^1.41.1",
+        "winston-transport": "4.*"
+      },
+      "engines": {
+        "node": "^18.19.0 || >=20.6.0"
+      }
+    },
     "node_modules/@pkgjs/parseargs": {
       "version": "0.11.0",
       "resolved": "https://registry.npmjs.org/@pkgjs/parseargs/-/parseargs-0.11.0.tgz",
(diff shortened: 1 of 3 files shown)


pujitha24:
This has been rebased and green for a little while now - happy to make any changes if something would help move review along.

tylerbenson:
What are winston logs?  I've never heard of this before.  Please provide a bit more context.

pujitha24:
Sorry, should've defined that upfront. Winston (https://github.com/winstonjs/winston) is a popular Node.js logging library. This layer already ships auto-instrumentation for it (`@opentelemetry/instrumentation-winston`) that's supposed to attach a transport to a user's winston logger so their log calls get exported as OTel log records — but that transport never got attached because its dependency (`@opentelemetry/winston-transport`) wasn't bundled into the layer, so the require silently failed. I've expanded the PR description with this context so it doesn't assume familiarity with winston. No code changes from the last review — same fix as before.

thompson-tomo:
I am little bit confused. If i look at https://www.npmjs.com/package/@opentelemetry/winston-transport it mentions

> If using @opentelemetry/instrumenation-winston package there is no need to instantiate the transport as the instrumentation will take care of that.

If i then look at the docs for the instrumentation it doesn't use the transport package https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-winston.

Looking at the changes here i can't see what difference the additional package is making to what is released. The issue suggests making changes to the wrapper but that hadn't been done here.

Issue #2099, opened 2025-12-30
Coldstart Processor creating broken traces for cold start and next invocation


jousby:
**Describe the bug**
I was unable to get the coldstart processor to do what is says that it does on the tin. If i submit 4 requests to a helloworld function behind an api gateway I only end up with 3 traces in my observability platform. Two of these traces will use the same internal spans, have missing spans and appears to be related to the cold start. One trace goes missing. Invocations from coldstart + 3 appear normal. 

**Steps to reproduce**
* Create  a hello world Lambda function fronted by either a REST or HTTP api in AWS
* Deploy the api and function with the otel collector and nodejs layers bundled with a custom collector.yaml that enables the coldstart processor. ie.
```
processors:
  coldstart:
  resource:
    attributes:
      - key: deployment.environment.name
        value: '${env:ENVIRONMENT_NAME}'
        action: upsert
```
Additional configure the collector.yaml to send to your observability platform of choice.
I had the following env variables set for the function:
```
      // OpenTelemetry configuration
      AWS_LAMBDA_EXEC_WRAPPER: '/opt/otel-handler',
      OPENTELEMETRY_COLLECTOR_CONFIG_URI: '/var/task/collector.yaml',
      OTEL_TRACES_SAMPLER: 'always_on',
      OTEL_PROPAGATORS: 'tracecontext,baggage',
      OTEL_LAMBDA_DISABLE_AWS_CONTEXT_PROPAGATION: 'true',
      OTEL_NODE_ENABLED_INSTRUMENTATIONS: 'aws-sdk,aws-lambda,http,undici',
      OTEL_SERVICE_NAME: serviceName,
      OPENTELEMETRY_EXTENSION_LOG_LEVEL: 'INFO',
```
I tried active tracing on, passthrough and off. 
I tried the insights layer on and off.
* Manually trigger a http get request to the api/function endpoint four times
* Check traces that arrive in the observability platform

**What did you expect to see?**
I expected to see 4 traces arrive in my observability platform (grafana cloud). I expected the first trace that would have triggered a cold start to have a trace that includes a span for the init phase and be of a longer duration than the other three.

**What did you see instead?**
* Only three traces arrived.

<img width="1667" height="152" alt="Image" src="https://github.com/user-attachments/assets/b22c2ab9-8c48-490f-b09e-504e6f1d4f3e" />

* One trace went missing
* The first two traces are broken
   * trace 1
[...]

Issue #1652, opened 2025-01-09
Adding coldstart processor to pipeline drops platform.initRuntimeDone span and does not set faas.coldstart on exec span


SeanHood:
**Describe the bug**
Adding coldstart processor to pipeline drops `platform.initRuntimeDone` span and does not set `faas.coldstart` on exec span.

With `coldstart` in the list of processors I lose the `platform.initRuntimeDone` span. And I don't get `faas.coldstart` apended to my execution span. If I remove `coldstart` from `processors` I get both spans.

**Steps to reproduce**
I'm using the following config for the collector
```
receivers:
  telemetryapi:
  otlp:
    protocols:
      grpc:
        endpoint: "localhost:4317"
      http:
        endpoint: "localhost:4318"

exporters:
  debug:
    verbosity: detailed

processors:
  coldstart:

service:
  pipelines:
    traces:
      receivers: [otlp, telemetryapi]
      processors: [coldstart]
      exporters: [debug]
```

**What did you expect to see?**
I expected to see `faas.coldstart` on the execution span and for the `platform.initRuntimeDone` span not to be dropped.

**What did you see instead?**

Without `coldstart`:
```
ResourceSpans #0
Resource SchemaURL: 
Resource attributes:
     -> faas.invoked_provider: Str(aws)
     -> service.name: Str(otel_lambda_hello_world)
     -> faas.name: Str(otel_lambda_hello_world)
     -> faas.max_memory: Str(128)
     -> faas.version: Str($LATEST)
     -> faas.invoked_region: Str(eu-west-2)
ScopeSpans #0
ScopeSpans SchemaURL: 
InstrumentationScope github.com/open-telemetry/opentelemetry-lambda/collector/receiver/telemetryapi 
Span #0
    Trace ID       : f630a91538b856f05d14f5b868f7e1cc
    Parent ID      : 
    ID             : 4550e3b670a330d3
    Name           : platform.initRuntimeDone
    Kind           : Internal
    Start time     : 2025-01-09 23:21:10.115 +0000 UTC
    End time       : 2025-01-09 23:21:11.549 +0000 UTC
    Status code    : Unset
    Status message : 
Attributes:
     -> faas.coldstart: Bool(true)

ResourceSpans #0
Resource SchemaURL: 
Resource attributes:
     -> telemetry.sdk.language: Str(python)
     -> telemetry.sdk.name: Str(opentelemetry)
[...]

PR #2520, opened 2026-08-05
[processor/coldstart] pair cold start spans using faas.invocation_id

naari3:
## Description

The processor identifies execution spans by `faas.execution`, which semantic conventions renamed to `faas.invocation_id` in v1.19.0. Current instrumentations set only the new name, so no execution span is ever recognized and the cold start span is never released.

The failure is silent rather than a no-op: a held cold start span is removed from the batch, so a batch containing only that span becomes empty and the processor returns `ErrSkipProcessingData`. Adding the coldstart processor to a pipeline drops `platform.initRuntimeDone` instead of enriching it.

This accepts both attribute names, so current and older instrumentations both work.

## Relation to #1216

@borchero submitted #1216 for this in March 2024. It went without review and was closed by the author. That PR renamed the attribute outright; this one keeps `faas.execution` working. Credit for finding the issue goes there, and I'm happy to close this in favour of a revived #1216.

## Testing

`TestPairingByInvocationID` covers a cold start span released by an execution span carrying `faas.invocation_id`. It fails without the fix. The existing tests were left on `faas.execution` as the backward-compatibility regression test and pass unchanged.

## Note

This also moves off `go.opentelemetry.io/collector/semconv`, deprecated since collector v0.126.0, to `go.opentelemetry.io/otel/semconv` as already used by `telemetryapireceiver`.

Relates to #2099 and #1652.


tylerbenson (approved):
Thanks for resubmitting this.  I think it's a good contribution.  Sorry @borchero that the original PR didn't get reviewed in a timely fashion.

borchero:
No worries, thanks for reviving @naari3 :)

wpessers (approved):

PR #2532, opened 2026-08-11
Enhance Renovate workflow with app token creation

thompson-tomo:
Added steps to create GitHub app token for non-fork PRs and updated token usage in checkout step.

This is to eliminate the need to be manually approving the ci run after the go mod tidy command has run.

The bot setup has been requested via https://github.com/open-telemetry/community/issues/3631

### DIFF
--- a/.github/workflows/renovate-go.yml
+++ b/.github/workflows/renovate-go.yml
@@ -23,8 +23,8 @@ jobs:
         uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
         id: app-token
         with:
-          client-id: ${{ vars.OTELBOT_FAAS_CLIENT_ID }}
-          private-key: ${{ secrets.OTELBOT_FAAS_PRIVATE_KEY }}
+          client-id: ${{ vars.OTELBOT_LAMBDA_CLIENT_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/renovate-go.yml
```diff
@@ -18,10 +18,27 @@ jobs:
...
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Create otelbot app token
+        if: ${{ github.event.pull_request.head.repo.fork == false }}
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        id: app-token
+        with:
+          client-id: ${{ vars.OTELBOT_LAMBDA_CLIENT_ID }}
+          private-key: ${{ secrets.OTELBOT_LAMBDA_PRIVATE_KEY }}
+          permission-contents: write
+      - name: Fork checkout
```
tylerbenson, line 29: When would this be a fork?  Doesn't renovate always run against a non-forked branch?

thompson-tomo: Correct but someone could checkout a renovate branch to make a fix in their fork and then raise a pr to add their fix + the renovate change back to this repo.

We only check branch name.

tylerbenson (approved):

Issue #2536, opened 2026-08-13
Publish a opentelemetry-javawrapper-layer.tar.gz


bcmedeiros:
**Is your feature request related to a problem? Please describe.**
I'm currently having to do all this to have this layer in my docker image:
```
FROM --platform=$BUILDPLATFORM public.ecr.aws/lambda/java:17.al2023 AS otel-layer
ADD --checksum=sha256:9a6723de27846ccc7e1691a464733ba40b62622171060dbff06698e346b7d3de \
    https://github.com/open-telemetry/opentelemetry-lambda/releases/download/layer-javaagent%2F0.21.0/opentelemetry-javawrapper-layer.zip /tmp/layer.zip
RUN dnf install -y unzip && unzip -q /tmp/layer.zip -d /tmp/layer

FROM public.ecr.aws/lambda/java:17.al2023
COPY --from=otel-layer --chmod=755 /tmp/layer/ /opt/
```

**Describe the solution you'd like**
If this project publishes a `tar.gz` version of the artifact in the release, we could have a much simpler version:
```
FROM public.ecr.aws/lambda/java:17.al2023
ADD https://github.com/open-telemetry/opentelemetry-lambda/releases/download/layer-javaagent%2F0.21.0/opentelemetry-javawrapper-layer.tar.gz /opt/
```

tylerbenson:
Good suggestion.  Would you be interested in implementing the change yourself?

bcmedeiros:
> Good suggestion. Would you be interested in implementing the change yourself?

That would be my pleasure to help, @tylerbenson.

Created https://github.com/open-telemetry/opentelemetry-lambda/pull/2538

bcmedeiros:
Hey, I see the PR has been approved, thank you for the time going through that.
It seems to still be marked as "changes requested" by @wpessers, though.
I also clicked "update branch" which in hindsight I shouldn't, sorry for that, but it might need a new approval.
Let me know if you need anything from me, and what I should expect in terms of release timelines.

bcmedeiros:
@tylerbenson @wpessers can I expect a release of this any time soon? The PR has been merged, I'm not sure if I should close this issue as the author, or if that's part of some internal process. Let me know if there's something I could do to have a release out.

PR #2538, opened 2026-08-16
chore(java): provide tar.gz artifact in releases

bcmedeiros:
Solution for https://github.com/open-telemetry/opentelemetry-lambda/issues/2536

### DIFF
--- a/java/layer-wrapper/build.gradle.kts
+++ b/java/layer-wrapper/build.gradle.kts
@@ -14,10 +14,7 @@ dependencies {
 }
 
 tasks {
-    val createLayer = register<Zip>("createLayer") {
-        archiveFileName.set("opentelemetry-javawrapper-layer.zip")
-        destinationDirectory.set(file("$buildDir/distributions"))
-
+    val layerContents = copySpec {
         from(configurations["runtimeClasspath"]) {
             into("java/lib")
         }
@@ -30,8 +27,21 @@ tasks {
         from("scripts")
     }
 
+    val createZipLayer = register<Zip>("createLayer") {
+        archiveFileName.set("opentelemetry-javawrapper-layer.zip")
+        destinationDirectory.set(file("$buildDir/distributions"))
+        with(layerContents)
+    }
+
+    val createTarLayer = register<Tar>("createTarLayer") {
+        archiveFileName.set("opentelemetry-javawrapper-layer.tar.gz")
+        destinationDirectory.set(file("$buildDir/distributions"))
+        compression = Compression.GZIP
+        with(layerContents)
+    }
+
     named("assemble") {
-        dependsOn(createLayer)
+        dependsOn(createZipLayer, createTarLayer)
     }
 }
 
--- a/java/layer-wrapper/build.gradle.kts
+++ b/java/layer-wrapper/build.gradle.kts
@@ -24,7 +24,11 @@ tasks {
             into("java/lib")
         }
 
-        from("scripts")
+        from("scripts") {
+                filePermissions {
+                        unix("755")
+                }
+        }
     }
 
     val createZipLayer = register<Zip>("createLayer") {
--- a/.github/actions/build-java-layer/action.yml
+++ b/.github/actions/build-java-layer/action.yml
@@ -8,6 +8,9 @@ outputs:
   javawrapper-artifact-path:
     description: 'Absolute path to the javawrapper layer zip artifact'
     value: ${{ github.workspace }}/java/layer-wrapper/build/distributions/opentelemetry-javawrapper-layer.zip
+  javawrapper-targz-artifact-path:
+    description: 'Absolute path to the javawrapper layer tar.gz artifact'
+    value: ${{ github.workspace }}/java/layer-wrapper/build/distributions/opentelemetry-javawrapper-layer.tar.gz
   javaagent-version:
     description: 'OpenTelemetry Java agent version included in the layer'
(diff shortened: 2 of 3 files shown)


wpessers (approved):

wpessers:
@bcmedeiros in the workaround you mention in that issue you explicitly add the `--chmod=755` arg. I think that something similar would still be required when trying to use this archive in your Dockerfile directly.

It looks like gradle's default file permissions make it so our wrapper scripts are all `644` (not executable)... When we use them inside a lambda layer there's no issue because aws lambda seems to automatically do some sort of normalization of permissions when extracting layer contents.

I think we can make the script files in the archive executable though, look at this example from gradle docs: https://docs.gradle.org/current/userguide/working_with_files.html#using_copyspec_dir_permissions
I'd suggest doing something similar to configure `755` permissions for our scripts, in that `copySpec` block.

java/layer-wrapper/build.gradle.kts
```diff
@@ -30,8 +27,21 @@ tasks {
         from("scripts")
```
wpessers: wrt my comment, something like this should do the trick:

```suggestion
        from("scripts") {
                filePermissions {
                        unix("755")
                }
        }
```

or

```suggestion
        from("scripts") {
                filePermissions {
                        unix("rwxr-xr-x")
                }
        }
```

wpessers (changes requested):

bcmedeiros:
The tar.gz package will preserve the original permissions of the file as committed in git, while zip will not, that's probably why the workout was needed in my example.
I'm pretty sure I tested the tar had 755 scripts, but I'm on my phone now, I can't check how the files are in git.
Anyway, we could apply your suggestion, it won't cause any harm.

java/layer-wrapper/build.gradle.kts
bcmedeiros: Screenshot_20260818_055532_Firefox
The file is already executable in git, and tar.gz wik preserve that during packaging, so I think this will be redundant. let me know if you really want to apply this change.

bcmedeiros: cc @wpessers

wpessers: Let me check locally when creating the archive through gradle directly. How I read it in the gradle docs about permissions it looked like it would default to `644`.

bcmedeiros: all right, let me know how it goes.

wpessers: Actually just checked the existing release artifacts, latest release:
```
-rw-r--r--  2.0 unx      761 b- defN 80-Feb-01 00:00 otel-handler
-rw-r--r--  2.0 unx      771 b- defN 80-Feb-01 00:00 otel-proxy-handler
-rw-r--r--  2.0 unx      762 b- defN 80-Feb-01 00:00 otel-sqs-handler
-rw-r--r--  2.0 unx      764 b- defN 80-Feb-01 00:00 otel-stream-handler
```

So gradle docs seem to be correct, did a quick cross check with some older releases and it seems like this behaviour changed with the upgrade from gradle 8 to 9. For the 0.15.0 version artifact for example:
```
-rwxr-xr-x  2.0 unx      770 b- defN 25-Aug-20 18:48 otel-proxy-handler
-rwxr-xr-x  2.0 unx      761 b- defN 25-Aug-20 18:48 otel-sqs-handler
-rwxr-xr-x  2.0 unx      760 b- defN 25-Aug-20 18:48 otel-handler
-rwxr-xr-x  2.0 unx      764 b- defN 25-Aug-20 18:48 otel-stream-handler
```

Will check with your branch now locally as well, but I'm expecting the exact same behaviour.

wpessers: Verified after building locally:
```
~/p/o/java (add-tar-gz-java)> tar -tzvf layer-wrapper/build/distributions/opentelemetry-javawrapper-layer.tar.gz | grep otel-                                        <region:eu-west-1>
-rw-r--r--  0 0      0         761 Jan  2  1970 otel-handler
-rw-r--r--  0 0      0         771 Jan  2  1970 otel-proxy-handler
-rw-r--r--  0 0      0         762 Jan  2  1970 otel-sqs-handler
-rw-r--r--  0 0      0         764 Jan  2  1970 otel-stream-handler
```

bcmedeiros:
you were right, @wpessers, Gradle's `from("scripts")` normalizes the file permissions. I've accepted your proposed changes.

java/layer-wrapper/build.gradle.kts
bcmedeiros: we ended up testing at the same time :)
yes, you were right. I remember `zip` not preserving file permissions while `tar` would, but you tracked that down to the Gradle upgrade which fits perfectly IMV.

tylerbenson (approved):
Thanks for taking care of this!

bcmedeiros:
@tylerbenson @wpessers what's the usual path forward from here? Can I help getting this released in some way? We are currently rolling this out to quite a few images, it would be really nice to have the compact setup version available.

wpessers:
@bcmedeiros my bad I'll get it merged in a sec. Will be released with next release of the lambda layers. I will look into expediting that next release so that you can start using it ASAP!

wpessers (approved):

wpessers:
@bcmedeiros thank you for the contribution! 🚀

bcmedeiros:
It's my pleasure to help, looking forward to the release! thanks!

PR #2558, opened 2026-08-25
ci: Update postUpdateOptions for go

thompson-tomo:
This adjusts the go update options to tidy all files with the -e option and enables import path updates. This combination should reduce how much is being updated via the custom Workflow.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -33,7 +33,9 @@
         'golang',
       ],
       postUpdateOptions: [
-        'gomodTidy',
+        'gomodTidyAll',
+        'gomodTidyE',
+        'gomodUpdateImportPaths',
       ],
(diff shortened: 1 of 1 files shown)


tylerbenson (approved):

wpessers:
You suggest closing #2546 then @thompson-tomo in favour of this one? Just to be clear, do we now know for sure this will stop that looping behaviour from being triggered? Or will we still often rely on human intervention (setting the stop-updating label)?

thompson-tomo:
@wpessers what i am inclined to do is get this in and monitor how it behaves. For instance does it reduce the diff generated by the custom go mod tidy,  does it resolve lockfile or is it still an issue.

I don't know for sure it will resolve but i expect it to improve the situation.

PR #2561, opened 2026-08-28
fix(dotnet): target net8.0 so the wrapper sample emits spans

vasireddy99:
The dotnet `aws-sdk/wrapper` integration test was failing because the trace contained only Lambda's passive Active-tracing segments; the instrumented handler span and downstream S3 span never reached X-Ray.

## Changes
- Retarget the `aws-sdk` sample to **net8.0** (the `dotnet6` Lambda runtime is deprecated) in the `.csproj`, `build.sh`, and Terraform `main.tf`.

## Testing
Files affected are limited to `dotnet/sample-apps/aws-sdk/`.

### DIFF
--- a/dotnet/sample-apps/aws-sdk/deploy/wrapper/main.tf
+++ b/dotnet/sample-apps/aws-sdk/deploy/wrapper/main.tf
@@ -5,7 +5,7 @@ module "hello-lambda-function" {
   architectures = compact([var.architecture])
   function_name = var.name
   handler       = "AwsSdkSample::AwsSdkSample.Function::TracingFunctionHandler"
-  runtime       = "dotnet6"
+  runtime       = "dotnet8"
 
   create_package         = false
   local_existing_package = "${path.module}/../../wrapper/SampleApps/build/function.zip"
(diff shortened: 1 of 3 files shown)


dotnet/sample-apps/aws-sdk/deploy/wrapper/main.tf
```diff
@@ -5,14 +5,18 @@ module "hello-lambda-function" {
...
   create_package         = false
   local_existing_package = "${path.module}/../../wrapper/SampleApps/build/function.zip"
 
   memory_size = 384
   timeout     = 20
 
+  environment_variables = {
+    OTEL_SERVICE_NAME = var.name
+  }
+
```
thompson-tomo: Why is this necessary?

vasireddy99: The wrapper doesn't set service.name anywhere, so the OpenTelemetry .NET SDK applies its  spec fallback and the instrumented span is exported to X-Ray as unknown_service:dotnet. I confirmed this by deploying the sample without the variable and inspecting the trace:
```
 SEGMENT name='unknown_service:dotnet' origin=None
  subsegment name='S3' namespace='aws'
 ```

thompson-tomo: I understand, so it is a nice to have but it is totally unrelated to the bump in sdk/framework to net 8.

For me i don't see the need to set it given this is just a sample of what is needed. The other examples don't have it & there is alternative ways to setting it which are likely more efficient.

As such Let's take it out as it's not necessary.

vasireddy99: agreed, removed this change.

thompson-tomo (approved):

PR #2598, opened 2026-09-13
chore: Add rules to control go updates

### LINKED ISSUE #1838: Go: auto-bump go version

thompson-tomo:
Closes #1838

As part of the go lock file pr, i noticed it was bumping the min go version. This bumping should be done via a dedicated pr which requires approval. In cases where a dependency requires a newer version it should be ignored until such a time that the min version has been bumped.

This blocking is achieved by enforcing strict constraint behaviour. The bumping is achieved by the other rule which requires approval for minor and major go updates.


collector/internal/tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/open-telemetry/opentelemetry-lambda/collector/internal/tools
 
-go 1.26.1
```
wpessers, line 3: Probably no big deal to move to `1.26.0`. But, these have been on `1.26.1` for over 6 months at this point. Is there a reason this changed?

thompson-tomo: I lowered it to be consistent with the collector and this way we have renovate only update major/minor.

wpessers: Makes sense then.

wpessers:
Thanks for the work as always, good stuff! 🚀

wpessers (approved):

PR #2607, opened 2026-09-14
ci: Add CI workflow for Ruby project

thompson-tomo:
Now when a ruby related dependency changes we are checking that it builds.


tylerbenson (approved):
Thanks!

